PR #31, opened 2023-09-18
chore: enforce use of conventional commits

tswast:
This will prevent accidental merging of commits that release-please can't handle.


ashleyxuu (approved):
LGTM

PR #36, opened 2023-09-20
perf: simplify join order to use multiple order keys instead of string.

TrevorBergeron:
Change-Id: I8c37e9296b2e4e0ea87f6a7e836d48988d161d37


TrevorBergeron:
Question from  @tswast 
> Can we remove string_encoded now? Or is it still used elsewhere?
We still use string_encoded for concat. I can remove that next, but it will be a bit tricky, so leaving out of this pr.

tswast (approved):

PR #53, opened 2023-09-22
deps: migrate to `ibis-framework >= "7.1.0"`

tswast:
This should unlock some bug fixes as well as potential `UNNEST` support in a future change.

Towards internal issue 301459557 🦕


tswast:
We're getting a few failures, such as

```
____________________________ test_df_nlargest[all] _____________________________
[gw0] linux -- Python 3.11.1 /tmpfs/src/github/python-bigquery-dataframes/.nox/system_prerelease/bin/python

scalars_df_index =           bool_col                                          bytes_col  \
rowindex                                     ....999999+00:00  
8                         T             <NA>                              <NA>  

[9 rows x 13 columns]
scalars_pandas_df_index =           bool_col  ...                     timestamp_col
rowindex            ...                                  
0 ... ...  2038-01-19 03:14:17.999999+00:00
8            False  ...                              <NA>

[9 rows x 13 columns]
keep = 'all'

    @pytest.mark.parametrize(
        ("keep",),
        [
            ("first",),
            ("last",),
            ("all",),
        ],
    )
    def test_df_nlargest(scalars_df_index, scalars_pandas_df_index, keep):
>       bf_result = scalars_df_index.nlargest(
            3, ["bool_col", "int64_too"], keep=keep
        ).to_pandas()

[tests/system/small/test_dataframe.py:156](https://cs.corp.google.com/piper///depot/google3/tests/system/small/test_dataframe.py?l=156): 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[bigframes/dataframe.py:929](https://cs.corp.google.com/piper///depot/google3/bigframes/dataframe.py?l=929): in nlargest
    return DataFrame(block_ops.nlargest(self._block, n, column_ids, keep=keep))
[bigframes/core/block_transforms.py:548](https://cs.corp.google.com/piper///depot/google3/bigframes/core/block_transforms.py?l=548): in nlargest
    block, counter = block.apply_window_op(
[bigframes/core/blocks.py:768](https://cs.corp.google.com/piper///depot/google3/bigframes/core/blocks.py?l=768): in apply_window_op
    block = Block(
[bigframes/core/blocks.py:112](https://cs.corp.google.com/piper///depot/google3/bigframes/core/blocks.py?l=112): in __init__
    self._expr = self._normalize_expression(expr, self._index_columns)
[bigframes/core/blocks.py:1093](https://cs.corp.google.com/piper///depot/google3/bigframes/core/blocks.py?l=1093): in _normalize_expression
    col_id for col_id in expr.column_ids if col_id not in index_columns
[bigframes/core/__init__.py:82](https://cs.corp.google.com/piper///depot/google3/bigframes/core/__init__.py?l=82): in column_ids
    return self._compile_ordered().column_ids
[bigframes/core/__init__.py:95](https://cs.corp.google.com/piper///depot/google3/bigframes/core/__init__.py?l=95): in _compile_ordered
    return compiler.compile_ordered(self.node)
[bigframes/core/compile/compiler.py:33](https://cs.corp.google.com/piper///depot/google3/bigframes/core/compile/compiler.py?l=33): in compile_ordered
    return typing.cast(compiled.OrderedIR, compile_node(node, True))
[bigframes/core/compile/compiler.py:45](https://cs.corp.google.com/piper///depot/google3/bigframes/core/compile/compiler.py?l=45): in compile_node
    return _compile_node(node, ordered)
/[usr/local/lib/python3.11/functools.py:909](https://cs.corp.google.com/piper///depot/google3/usr/local/lib/python3.11/functools.py?l=909): in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
[bigframes/core/compile/compiler.py:192](https://cs.corp.google.com/piper///depot/google3/bigframes/core/compile/compiler.py?l=192): in compile_window
    result = compile_ordered(node.child).project_window_op(
[bigframes/core/compile/compiled.py:868](https://cs.corp.google.com/piper///depot/google3/bigframes/core/compile/compiled.py?l=868): in project_window_op
    result = self._set_or_replace_by_id(output_name or column_name, window_op)
[bigframes/core/compile/compiled.py:1185](https://cs.corp.google.com/piper///depot/google3/bigframes/core/compile/compiled.py?l=1185): in _set_or_replace_by_id
    return builder.build()
[bigframes/core/compile/compiled.py:1344](https://cs.corp.google.com/piper///depot/google3/bigframes/core/compile/compiled.py?l=1344): in build
    return OrderedIR(
[bigframes/core/compile/compiled.py:632](https://cs.corp.google.com/piper///depot/google3/bigframes/core/compile/compiled.py?l=632): in __init__
    super().__init__(table, columns, predicates)
[bigframes/core/compile/compiled.py:65](https://cs.corp.google.com/piper///depot/google3/bigframes/core/compile/compiled.py?l=65): in __init__
    self._column_names = {column.get_name(): column for column in self._columns}
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

.0 = <tuple_iterator object at 0x7f117affa7d0>

>   self._column_names = {column.get_name(): column for column in self._columns}
E   TypeError: unhashable type: 'Deferred'

[bigframes/core/compile/compiled.py:65](https://cs.corp.google.com/piper///depot/google3/bigframes/core/compile/compiled.py?l=65): TypeError
```

I asked some friends over on the ibis project, and they shared that Deferred objects aren't bound to a table expression. Also, `op()` is hashable but the expression might not be. The "more future-proof way is to resolve the deferred through the table’s `__getitem__`" that is to say "`table[deferred].op()`"

tswast:
Marking as Draft again until we have a greater need to upgrade.

ashleyxuu (approved):

TrevorBergeron (approved):

tswast:
qcut support currently blocked by what I believe to be a regression in ibis related to those Deferred objects: https://github.com/ibis-project/ibis/issues/7631

Will investigate if there's a way to work around it. Seems we are ending up in this code path somehow: https://github.com/ibis-project/ibis/blob/d6a2f0922ea92b0e03a0431bf9011e7103565061/ibis/expr/types/generic.py#L760

bigframes/core/compile/compiled.py
```diff
@@ -62,7 +63,16 @@ def __init__(
         self._columns = tuple(columns)
         # To allow for more efficient lookup by column name, create a
         # dictionary mapping names to column values.
-        self._column_names = {column.get_name(): column for column in self._columns}
+        self._column_names = {
```
kszucs, line 66: Apparently you are adding support for `Deferred` objects. Currently we have a couple of objects interpretable as a value expression besides the expressions themselves, e.g. column names or selector in addition to deferred objects. The issue that ibis doesn't provide a clear, concise and backward compatible API function to bind those objects to a table. We have various protected functions and methods used in a hardly followable fashion. 

Instead of providing a stable API for deferred, wouldn't it be better to have a stable function which convert any sort of ibis objects to an expression (given a table) if possible, something like `ibis.bind(obj, table)`?

Asking because we have something similar under preparation, but could be a good idea to have a public variant of that: https://github.com/ibis-project/ibis/pull/7580/files#diff-4400549b2eaa4ca3a3107f13fa78210e37277659acfe2a1eb4129cfbfaa66531R105-R134

bigframes/remote_function.py
```diff
@@ -516,17 +516,14 @@ def remote_function_node(
     """Creates an Ibis node representing a remote function call."""
 
     fields = {
-        name: rlz.value(type_) if type_ else rlz.any
+        name: rlz.ValueOf(None if type_ == "ANY TYPE" else type_)
```
kszucs, line 519: Actually we are using almost the same code internally, possibly it could worth to expose it in a more developer friendly way after we understand the actual requirements. We can open an issue on our side to track this.

third_party/bigframes_vendored/ibis/expr/operations/analytic.py
```diff
@@ -2,22 +2,24 @@
 
 from __future__ import annotations
 
-from ibis.expr.operations.analytic import Analytic
+import ibis.expr.datatypes as dt
+import ibis.expr.operations.analytic as ibis_ops_analytic
+import ibis.expr.operations.core as ibis_ops_core
```
kszucs, line 7: I suggest to just simply use `import ibis.expr.operations as ops` since all op types are exposed there.

third_party/bigframes_vendored/ibis/expr/operations/analytic.py
```diff
@@ -2,22 +2,24 @@
...
 import ibis.expr.rules as rlz
 
 
-class FirstNonNullValue(Analytic):
+class FirstNonNullValue(ibis_ops_analytic.Analytic):
     """Retrieve the first element."""
 
-    arg = rlz.column(rlz.any)
-    output_dtype = rlz.dtype_like("arg")
+    arg: ibis_ops_core.Column[dt.Any]
```
kszucs, line 14: You can just simply use `ops.Column` without the parameter, it means the same thing as `ops.Column[dt.Any]`

third_party/bigframes_vendored/ibis/expr/operations/analytic.py
```diff
@@ -2,22 +2,24 @@
...
+    dtype = rlz.dtype_like("arg")
 
 
-class LastNonNullValue(Analytic):
+class LastNonNullValue(ibis_ops_analytic.Analytic):
```
kszucs, line 18: Since we have been breaking several internal parts of the codebase (sorry for that), maintaining custom operations this way is probably inconvenient. I wanted to have a documented and stable way of creating custom nodes for a long time now, just didn't have enough usage examples which we could draft a desired API from. This might be another topic we should discuss in our issue tracker.

tswast: Good call. Filed https://github.com/ibis-project/ibis/issues/7767

I don't even know that it needs to be a stable interface for custom ops. Maybe just an escape hatch of some sort for raw SQL that's just a Value, not a full table expression like the current Backend.sql() escape hatch. That said, such an escape hatch is going to look a lot like a custom op, since ideally we'd propagate type info somehow so that expressions can continue to be built on it.

bigframes/core/compile/compiled.py
tswast: > Instead of providing a stable API for deferred, wouldn't it be better to have a stable function which convert any sort of ibis objects to an expression (given a table) if possible, something like ibis.bind(obj, table)?

Yes, we'd definitely use something like that here.

Honestly, I'm not 100% how we are ending up with deferred objects in the first place. It seems to be with how we're building some of our windowed functions, but in all of the cases I know of there is an underlying table. I think we may just be hitting this bug: https://github.com/ibis-project/ibis/issues/7631

bigframes/remote_function.py
tswast: True. I was hoping to upstream this sooner, but now that I see https://ibis-project.org/reference/scalar-udfs#ibis.expr.operations.udf.scalar.builtin maybe I can migrate to that? I'll try that in a separate PR.

tswast: Started on this here: https://github.com/googleapis/python-bigquery-dataframes/pull/277

I found a few issues already with UDF in the BQ backend. I'll get those documented soon.

PR #54, opened 2023-09-22
feat: Add more index methods

TrevorBergeron:

### DIFF
--- a/bigframes/core/indexes/index.py
+++ b/bigframes/core/indexes/index.py
@@ -24,8 +24,10 @@ import pandas
 
 import bigframes.constants as constants
 import bigframes.core as core
+import bigframes.core.block_transforms as block_ops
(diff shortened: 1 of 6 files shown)


bigframes/core/indexes/index.py
```diff
@@ -149,6 +151,27 @@ def has_duplicates(self) -> bool:
     def _block(self) -> blocks.Block:
         return self._data._get_block()
 
+    @property
+    def T(self) -> Index:
+        return self.transpose()
+
+    def transpose(self) -> Index:
```
shobsi, line 158: Transpose is same as the input?

TrevorBergeron: Yeah, Transpose is just a no-op for Series and Index. Added for completness' sake, but not something that is terribly useful.

shobsi (approved):

PR #55, opened 2023-09-22
feat: label query jobs with an `bigframes-api-xx` key indicating the Python API(s) used to construct it

ashleyxuu:
Change-Id: I2a201ff6f35e4c945177b2f504b60a3744e86a0d

### DIFF
--- a/bigframes/core/__init__.py
+++ b/bigframes/core/__init__.py
@@ -472,7 +472,7 @@ class ArrayValue:
         sql = self._session.ibis_client.compile(count_expr)
         api_methods_len = len(self._api_methods)
 
-        # Initialize methods to add as an empty dictionary
+        # Initialize api-methods as an empty dictionary
         add_api_methods: Dict[str, str] = {}
         # We capture the latest label if it is out of the length limit of labels count
         if api_methods_len >= MAX_LABELS_COUNT:
@@ -483,7 +483,7 @@ class ArrayValue:
             sorted_items = sorted(self._api_methods.items())
             adeded_number = int(sorted_items[-1][1]) + 1
             add_api_key = "bigframes-api-" + str(adeded_number)
-            # Add the latest api to
+            # Add the latest api
             sorted_dict[add_api_key] = "shape"
             add_api_methods = sorted_dict
         elif api_methods_len == 0:
@@ -612,7 +612,7 @@ class ArrayValue:
             for col_in, agg_op, col_out in aggregations
         }
 
-        # Initialize methods to add as an empty dictionary
+        # Initialize api-methods as an empty dictionary
         add_api_methods: Dict[str, str] = {}
         api_methods_len = len(self._api_methods)
         if api_methods_len >= MAX_LABELS_COUNT:
@@ -623,7 +623,7 @@ class ArrayValue:
             sorted_items = sorted(self._api_methods.items())
             adeded_number = int(sorted_items[-1][1]) + 1
             add_api_key = "bigframes-api-" + str(adeded_number)
-            # Add the latest api to
+            # Add the latest api
             sorted_dict[add_api_key] = "shape"
             add_api_methods = sorted_dict
         elif api_methods_len == 0:
--- a/bigframes/core/__init__.py
+++ b/bigframes/core/__init__.py
@@ -474,8 +474,18 @@ class ArrayValue:
 
         # Initialize methods to add as an empty dictionary
         add_api_methods: Dict[str, str] = {}
+        # We capture the latest label if it is out of the length limit of labels count
         if api_methods_len >= MAX_LABELS_COUNT:
-            add_api_methods = self.api_method
+            # Pop out the first item
+            sorted_dict = dict(sorted(self._api_methods.items()))
+            sorted_dict.popitem()
+            # Get the new key number
+            sorted_items = sorted(self._api_methods.items())
+            adeded_number = int(sorted_items[-1][1]) + 1
+            add_api_key = "bigframes-api-" + str(adeded_number)
+            # Add the latest api to
+            sorted_dict[add_api_key] = "shape"
+            add_api_methods = sorted_dict
         elif api_methods_len == 0:
             add_api_key = "bigframes-api-0"
             add_api_methods = {add_api_key: "shape"}
@@ -606,7 +616,16 @@ class ArrayValue:
         add_api_methods: Dict[str, str] = {}
         api_methods_len = len(self._api_methods)
         if api_methods_len >= MAX_LABELS_COUNT:
-            add_api_methods = self.api_method
+            # Pop out the first item
+            sorted_dict = dict(sorted(self._api_methods.items()))
+            sorted_dict.popitem()
+            # Get the new key number
+            sorted_items = sorted(self._api_methods.items())
+            adeded_number = int(sorted_items[-1][1]) + 1
+            add_api_key = "bigframes-api-" + str(adeded_number)
+            # Add the latest api to
+            sorted_dict[add_api_key] = "shape"
+            add_api_methods = sorted_dict
         elif api_methods_len == 0:
             add_api_key = "bigframes-api-0"
             add_api_methods = {add_api_key: "aggregate"}
--- a/bigframes/core/__init__.py
+++ b/bigframes/core/__init__.py
@@ -45,6 +45,7 @@ if typing.TYPE_CHECKING:
 
 ORDER_ID_COLUMN = "bigframes_ordering_id"
 PREDICATE_COLUMN = "bigframes_predicate"
+MAX_LABELS_COUNT = 64
 
 
 @dataclass(frozen=True)
@@ -94,6 +95,7 @@ class ArrayValue:
         hidden_ordering_columns: Optional[Sequence[ibis_types.Value]] = None,
         ordering: ExpressionOrdering = ExpressionOrdering(),
         predicates: Optional[Collection[ibis_types.BooleanValue]] = None,
+        api_methods: Dict[str, str] = {},
     ):
         self._session = session
         self._table = table
@@ -120,6 +122,9 @@ class ArrayValue:
         self._hidden_ordering_column_names = {
             column.get_name(): column for column in self._hidden_ordering_columns
         }
+
+        self._api_methods = api_methods
+
         ### Validation
         value_col_ids = self._column_names.keys()
         hidden_col_ids = self._hidden_ordering_column_names.keys()
@@ -233,6 +238,10 @@ class ArrayValue:
             {**self._column_names, **self._hidden_ordering_column_names},
             self._ordering.all_ordering_columns,
         )
+
+    @property
+    def api_method(self) -> Dict[str, str]:
+        return self._api_methods
 
     def builder(self) -> ArrayValueBuilder:
         """Creates a mutable builder for expressions."""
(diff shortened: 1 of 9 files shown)


bigframes/core/__init__.py
```diff
@@ -94,6 +95,7 @@ def __init__(
         hidden_ordering_columns: Optional[Sequence[ibis_types.Value]] = None,
         ordering: ExpressionOrdering = ExpressionOrdering(),
         predicates: Optional[Collection[ibis_types.BooleanValue]] = None,
+        api_methods: Dict[str, str] = {},
```
TrevorBergeron, line 98: I'm worried about adding this to ArrayValue. Making api method tracking something that cuts up and down the stack will make it tricky to maintain.

bigframes/core/__init__.py
```diff
@@ -234,6 +239,10 @@ def _ibis_order(self) -> Sequence[ibis_types.Value]:
             self._ordering.all_ordering_columns,
         )
 
+    @property
+    def api_method(self) -> Dict[str, str]:
```
TrevorBergeron, line 243: Maybe a list of named tuples would be a good way to represent this?

TrevorBergeron (changes requested):

bigframes/core/__init__.py
ashleyxuu: I've considered it before, and I think it is a data structure choice. 1. `job_configs.labels` is a dictionary, so we do less work assigning variable there. 2. In python, we can sort the dictionary by key and value, so I don't think there is much difference than using a list of tuples for doing any operations.

bigframes/core/__init__.py
ashleyxuu: For APIs that will call down the to the lower level, like aggregate here, we pass it down to the ArrayValue. For other cases, we don't need to do so. What potential maintenance problems are you thinking about here? We don't need to consider any acyclic graph, so the stacking is just for debugging purpose.

tswast, line 98: Why is this a dictionary? Don't we just need a list?

bigframes/core/__init__.py
```diff
@@ -461,14 +470,37 @@ def shape(self) -> typing.Tuple[int, int]:
         width = len(self.columns)
         count_expr = self._to_ibis_expr(ordering_mode="unordered").count()
         sql = self._session.ibis_client.compile(count_expr)
-
+        api_methods_len = len(self._api_methods)
+
+        # Initialize api-methods as an empty dictionary
+        add_api_methods: Dict[str, str] = {}
+        # We capture the latest label if it is out of the length limit of labels count
+        if api_methods_len >= MAX_LABELS_COUNT:
```
tswast, line 478: This logic seems weirdly placed. I'd expect this on Session / where we start our BQ jobs.

bigframes/session.py
```diff
@@ -930,7 +931,7 @@ def _read_pandas(
 
         job_config = bigquery.LoadJobConfig(schema=schema)
         job_config.clustering_fields = cluster_cols
-        job_config.labels = {"bigframes-api": api_name}
+        job_config.labels = {"bigframes-io-api": api_name}
```
tswast, line 934: Why do we need a different key for this?

tswast (changes requested):
`feat: ` seems an odd classification. This isn't a new feature as far as end user's are concerned, right? It doesn't allow them to do anything new.

bigframes/core/__init__.py
ashleyxuu: Change it to list now.

ashleyxuu: Change to list done.

bigframes/core/__init__.py
ashleyxuu: Redo the logic here, and assign the key_name in `Session`.

bigframes/session.py
ashleyxuu: Change this back by removing the io.

bigframes/core/__init__.py
```diff
@@ -94,6 +95,7 @@ def __init__(
         hidden_ordering_columns: Optional[Sequence[ibis_types.Value]] = None,
         ordering: ExpressionOrdering = ExpressionOrdering(),
         predicates: Optional[Collection[ibis_types.BooleanValue]] = None,
+        api_methods: List[str] = [],
```
tswast, line 98: Don't use mutable objects for default values. https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

In fact, ArrayValue is intended to be immutable. A tuple would be more appropriate. Sequence is read-only. We could use that for the type annotation if we want to also accept list arguments. https://docs.python.org/3/library/typing.html#typing.Sequence

bigframes/core/__init__.py
```diff
@@ -234,6 +239,10 @@ def _ibis_order(self) -> Sequence[ibis_types.Value]:
             self._ordering.all_ordering_columns,
         )
 
+    @property
+    def api_method(self) -> List[str]:
```
tswast, line 243: Likewise, let's use `Sequence` here.

bigframes/core/__init__.py
```diff
@@ -461,6 +470,16 @@ def shape(self) -> typing.Tuple[int, int]:
         width = len(self.columns)
         count_expr = self._to_ibis_expr(ordering_mode="unordered").count()
         sql = self._session.ibis_client.compile(count_expr)
+        api_methods_len = len(self._api_methods)
+
+        # Initialize methods to add as an empty list
+        add_api_methods = []
+        api_methods_len = len(self._api_methods)
+        if api_methods_len >= MAX_LABELS_COUNT:
```
tswast, line 478: I don't think this is the correct layer to check checking for this limit. Rather, better to apply the limit at the time in which it matters (the labels on the job).

Also, the limit will be more appropriate to be applied when we create the job because it's possible we have other labels on the job besides these that we need to keep attached.

bigframes/core/__init__.py
```diff
@@ -461,6 +470,16 @@ def shape(self) -> typing.Tuple[int, int]:
...
+            add_api_methods = self.api_method[1:]
+        else:
+            add_api_methods = self.api_method
+        add_api_methods.append("shape")
```
tswast, line 482: As Trevor said, this feels like very much the wrong layer at which to log. We should be logging public user-facing APIs not our internal methods.

tswast (changes requested):

bigframes/core/__init__.py
ashleyxuu: Move the logic to `Session._start_query`

bigframes/core/__init__.py
ashleyxuu: There are a lot of methods using aggregate. In this cl, I only added the methods in DataFrameGroupBy, and the basic in DataFrame/Series as example. We can follow up adding other methods in another cl.

bigframes/core/__init__.py
ashleyxuu: Done

bigframes/core/__init__.py
ashleyxuu: Done using Sequence.

bigframes/core/blocks.py
```diff
@@ -601,11 +605,13 @@ def _split(
...
     def _compute_dry_run(
-        self, value_keys: Optional[Iterable[str]] = None
+        self,
+        value_keys: Optional[Iterable[str]] = None,
+        api_name: str = "",
     ) -> bigquery.QueryJob:
         expr = self._apply_value_keys_to_expr(value_keys=value_keys)
         job_config = bigquery.QueryJobConfig(dry_run=True)
-        _, query_job = expr.start_query(job_config=job_config)
+        _, query_job = expr.start_query(job_config=job_config, api_name=api_name)
```
tswast, line 614: Shouldn't this take a list?

bigframes/session.py
```diff
@@ -1449,14 +1449,34 @@ def _start_query(
...
+                label_values = api_methods
+            else:
+                cur_labels: Sequence[str] = [*job_config.labels.values()]
+                total_label_values = list(cur_labels) + list(api_methods)
+                # If the total number of labels is under the limit of labels count
+                if len(job_config.labels) + len(api_methods) <= _MAX_LABELS_COUNT:
+                    label_values = total_label_values
+                # We capture the latest label if it is out of the length limit of labels count
+                else:
+                    label_values = total_label_values[-_MAX_LABELS_COUNT:]
```
tswast, line 1469: We'll need to make sure we save room for `len(job_config.labels)`

tswast: I recommend saving some of this label logic in it's own function, possibly in https://github.com/googleapis/python-bigquery-dataframes/blob/main/bigframes/core/io.py and writing some unit tests that account for having many existing labels and many API methods.

tswast (changes requested):

bigframes/session.py
ashleyxuu: Done refactoring the labels logic in `io.py` file and adding unit test.

bigframes/core/io.py
```diff
@@ -16,11 +16,43 @@
...
+from typing import Dict, Optional, Sequence, Tuple, Union
 
 import google.cloud.bigquery as bigquery
 
 IO_ORDERING_ID = "bqdf_row_nums"
+MAX_LABELS_COUNT = 64
+
+
+def create_job_configs_labels(
+    job_config: Optional[bigquery.job.QueryJobConfig],
```
tswast, line 28: Nit: This might be easier to test if we just pass in the other labels as a dictionary and return the new labels as a dictionary.

bigframes/core/io.py
```diff
@@ -16,11 +16,43 @@
...
+                label_values = list(cur_labels) + list(api_methods)[-added_api_len:]
+    else:
+        job_config = bigquery.QueryJobConfig()
+        label_values = api_methods
+
+    labels = {}
+    for i, label_value in enumerate(label_values):
+        label_key = "bigframes-api-" + str(i)
+        labels[label_key] = label_value
+    job_config.labels = labels
```
tswast, line 54: Mutating the job config will make testing more difficult.

bigframes/core/io.py
```diff
@@ -16,11 +16,43 @@
...
+                label_values = list(cur_labels) + list(api_methods)
+            # We capture the latest label if it is out of the length limit of labels count
+            else:
+                added_api_len = cur_labels_len + api_methods_len - MAX_LABELS_COUNT
+                label_values = list(cur_labels) + list(api_methods)[-added_api_len:]
+    else:
+        job_config = bigquery.QueryJobConfig()
+        label_values = api_methods
+
+    labels = {}
```
tswast, line 50: We should populate this with the original labels (if any)

tests/unit/core/test_io.py
```diff
@@ -13,12 +13,77 @@
...
+    assert labels is not None
+    assert len(labels) == 5
+    assert labels == expected_dict
+
+
+def test_create_job_configs_labels_length_limit_met():
+    mock_job_config = mock.create_autospec(bigquery.QueryJobConfig())
+    cur_labels = {}
+    for i in range(63):
+        key = f"bigframes-api{i}"
```
tswast, line 72: Ideally we'd have keys other than `bigframes-api` for the existing labels so that we make sure we aren't overwriting keys that need to be preserved.

bigframes/core/io.py
ashleyxuu: Pass existing labels as dictionary and return dictionary now.

bigframes/core/io.py
ashleyxuu: Pass existing labels as dictionary and return dictionary now. We test if the the output dictionary is expected.

tests/unit/core/test_io.py
ashleyxuu: Preserve the existing key labels now. When we add new labels, we start counting from the length of the existing labels, making sure that we won't override the keys.

tswast:
Closing in favor of the approach in https://github.com/googleapis/python-bigquery-dataframes/pull/145

PR #62, opened 2023-09-26
refactor: push down SQL generate logic in core.BqmlModelFactory

GarrettWu:
I want to make SQL generation as much as possible to stay in the ml.sql module. Will have other changes later.


ashleyxuu (approved):
Thanks for refactoring!

PR #63, opened 2023-09-26
feat: add df.unstack

TrevorBergeron:

### DIFF
--- a/tests/system/small/test_dataframe.py
+++ b/tests/system/small/test_dataframe.py
@@ -1232,11 +1232,6 @@ def test_combine_first(
     pd_df_b.columns = ["b", "a", "d"]
     pd_result = pd_df_a.combine_first(pd_df_b)
 
-    print("pandas")
-    print(pd_result.to_string())
-    print("bigframes")
-    print(bf_result.to_string())
-
     # Some dtype inconsistency for all-NULL columns
     pd.testing.assert_frame_equal(bf_result, pd_result, check_dtype=False)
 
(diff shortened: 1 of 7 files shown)


tests/system/small/test_dataframe.py
```diff
@@ -1705,6 +1700,26 @@ def test_df_stack(scalars_dfs):
     pd.testing.assert_series_equal(bf_result, pd_result, check_dtype=False)
 
 
+def test_df_unstack(scalars_dfs):
+    scalars_df, scalars_pandas_df = scalars_dfs
+    # To match bigquery dataframes
```
shobsi, line 1705: Didn't quite get this, why can't we use the original `scalars_pandas_df`?

tests/system/small/test_multiindex.py
```diff
@@ -729,6 +729,26 @@ def test_column_multi_index_stack(scalars_df_index, scalars_pandas_df_index):
     )
 
 
+def test_column_multi_index_unstack(scalars_df_index, scalars_pandas_df_index):
```
shobsi, line 732: [nit] would `test_multi_column_unstack` better represent this test?

tests/system/small/test_multiindex.py
```diff
@@ -866,6 +886,17 @@ def test_column_multi_index_reorder_levels(scalars_df_index, scalars_pandas_df_i
     pandas.testing.assert_frame_equal(bf_result, pd_result)
 
 
+def test_multi_index_unstack(hockey_df, hockey_pandas_df):
+    bf_result = (
+        hockey_df.set_index(["team_name", "season", "position"]).unstack().to_pandas()
+    )
+    pd_result = hockey_pandas_df.set_index(
+        ["team_name", "season", "position"]
+    ).unstack()
```
shobsi, line 895: I think it would be worth testing for all levels: `{-3, -2, -1, 0, 1, 2}`

TrevorBergeron: I don't support levels param yet. Will add levels support to stack() and unstack() in a later pr.

tests/system/small/test_multiindex.py
TrevorBergeron: My naming convention is based on the initial dataframe object, column_multi_index indicating that the initial column axis is a multi-index. Open to changing this convention, but it is inline with how I've named the tests so far.

tests/system/small/test_dataframe.py
shobsi: So if we remove

```python
scalars_pandas_df = scalars_pandas_df.copy()
```

the following will still apply, just on the original dataframe, no?

```python
scalars_pandas_df.columns = scalars_pandas_df.columns.astype("string[pyarrow]")
```

TrevorBergeron: tests share resources, so we create a copy to make sure our mutation doesn't affect other tests

shobsi: Ah, I was missing that, thanks!

shobsi (approved):

PR #65, opened 2023-09-26
fix: generate unique ids on join to avoid id collisions

TrevorBergeron:

### DIFF
--- a/bigframes/core/joins/single_column.py
+++ b/bigframes/core/joins/single_column.py
@@ -16,6 +16,7 @@
 
 from __future__ import annotations
 
+import itertools
 import typing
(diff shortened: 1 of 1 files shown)


milkshakeiii (approved):
Great!

bigframes/core/joins/single_column.py
```diff
@@ -26,6 +27,7 @@
 import bigframes.constants as constants
 import bigframes.core as core
 import bigframes.core.guid
```
milkshakeiii, line 29: I think this import can be removed?

TrevorBergeron: yeah, with some slight changes, have now removed

PR #68, opened 2023-09-28
feat: Allow df.drop to take an index object

milkshakeiii:


bigframes/dataframe.py
```diff
@@ -925,10 +927,53 @@ def drop(
             )
         if columns:
             block = block.drop_columns(self._sql_names(columns))
-        if not index and not columns:
+        if index is None and not columns:
             raise ValueError("Must specify 'labels' or 'index'/'columns")
         return DataFrame(block)
 
+    def _drop_by_index(self, index: indexes.Index) -> DataFrame:
```
TrevorBergeron, line 934: Can we do this with a block-level join and filter op? Implementing as a block-operation will make it easier to share logic with series. I also worry that this complex index manipulation has edge cases, in particular when it comes to column multi-indices.

bigframes/dataframe.py
```diff
@@ -906,16 +906,18 @@ def drop(
                 columns = labels
 
         block = self._block
-        if index:
+        if index is not None:
             level_id = self._resolve_levels(level or 0)[0]
 
-            if utils.is_list_like(index):
+            if utils.is_list_like(index) or isinstance(index, pandas.Index):
```
TrevorBergeron, line 912: Doesn't utils.is_list_like already match indices?

milkshakeiii: good point, only the fix above ("if index is not None:") is needed. This line is not.

bigframes/dataframe.py
milkshakeiii: Okay, I have refactored it to use block-level join and filter op. The code should be much improved now, or at least much shortened.

tswast (approved):

PR #76, opened 2023-10-02
feat: add equals methods to series/dataframe

TrevorBergeron:


bigframes/core/block_transforms.py
```diff
@@ -24,6 +24,39 @@
 import bigframes.operations.aggregations as agg_ops
 
 
+def equals(block1: blocks.Block, block2: blocks.Block):
```
ashleyxuu, line 27: nit: can add a return type hint here.

ashleyxuu (approved):

bigframes/core/block_transforms.py
TrevorBergeron: added

PR #87, opened 2023-10-04
feat: use default session connection

GarrettWu:

http://b/303283207


jiaxunwu (approved):

bigframes/session.py
```diff
@@ -321,7 +323,10 @@ def __init__(
             ),
         )
 
-        self._bq_connection = context.bq_connection
+        self._bq_connection = (
+            context.bq_connection
```
shobsi, line 327: We should transform this internally to format <project>.<location>.<connection-id> if user just provides <connection-id>. Ashley added this handling in remote function, may be we could generalize that: https://github.com/googleapis/python-bigquery-dataframes/blob/c0efec8956198247b27904345a795f09c80d3502/bigframes/remote_function.py#L767

bigframes/session.py
```diff
@@ -97,6 +97,8 @@
 _BIGQUERYCONNECTION_REGIONAL_ENDPOINT = "{location}-bigqueryconnection.googleapis.com"
 _BIGQUERYSTORAGE_REGIONAL_ENDPOINT = "{location}-bigquerystorage.googleapis.com"
 
+_BIGFRAMES_DEFAULT_CONNECTION_ID = "bigframes-default-connection"
```
shobsi, line 100: Should we add a test that captures the use of this? Will have to create the connection and  configure IAM on its  SA as a prerequisite.

GarrettWu: I can add one after http://cl/571816142.

bigframes/session.py
GarrettWu: We can make it happen, but right now remote function session is an optional field. Which makes supplementing connection project and location from session not always possible. Do we want to make remote function always take bpd.get_default_session() as back up?

shobsi: I think that's a good idea. Current remote_function session logic pre-dates the default session. we can use default session as a backup if session is not provided

GarrettWu: Synced offline, we can add bigframes default session to remote function, so remote function will always have a session, and add a helper to share the functionality of supplementing connections.


def helper(default_proj:str, default_loc:str, default_conn_id:str) -> str:

bigframes/session.py
GarrettWu: Added

bigframes/remote_function.py
```diff
@@ -695,9 +695,12 @@ def remote_function(
             persistent name.
 
     """
```
shobsi, line 697: [nit]

```python
    if not session:
        session = bpd.get_global_session()
```

bigframes/remote_function.py
```diff
@@ -740,11 +741,11 @@ def remote_function(
         dataset_ref = bigquery.DatasetReference.from_string(
             dataset, default_project=bigquery_client.project
         )
-    elif session:
+    else:
         dataset_ref = bigquery.DatasetReference.from_string(
             session._session_dataset_id, default_project=bigquery_client.project
         )
-    else:
+    if not dataset_ref:
```
shobsi, line 748: Looks like it is redundant now?

bigframes/clients.py
```diff
@@ -162,3 +164,25 @@ def _get_service_account_if_connection_exists(
             pass
 
         return service_account
+
+
+def get_connection_name_full(
```
shobsi, line 169: Would it be tidier to make it a class method in `BqConnectionManager`?

tests/system/small/test_remote_function.py
```diff
@@ -465,6 +465,36 @@ def square(x):
     assert_pandas_df_equal_ignore_ordering(bf_result, pd_result)
 
 
+@pytest.mark.flaky(retries=2, delay=120)
+def test_remote_function_default_connection(scalars_dfs, dataset_id):
+    @rf.remote_function([int], int, dataset=dataset_id)
```
shobsi, line 470: direct version is deprecated, recommend testing `session.remote_function` or `bigframes.pandas.remote_function` interface.

https://github.com/googleapis/python-bigquery-dataframes/blob/228aeba09782ae2421040c7601c15d4af92790b6/bigframes/remote_function.py#L609

tests/unit/test_clients.py
shobsi, line 1: Nice, thanks! Let's add one test for the `Invalid connection name format` scenario as well

bigframes/clients.py
shobsi: IMHO it would still make sense, `@classmethod` would mean that it does not depend on the instance level state, but conceptually it would fit in nicely - BqConnectionManager sounds like the entity which is supposed to know the low level details of a connection and can provide helper functions about the same.

bigframes/clients.py
```diff
@@ -162,3 +164,25 @@ def _get_service_account_if_connection_exists(
...
+    connection_name: Optional[str], default_project: str, default_location: str
```
shobsi, line 170: awkward to have optional argument followed by mandatory arguments, would be nice to have it in the end

bigframes/ml/llm.py
```diff
@@ -48,7 +49,14 @@ def __init__(
         connection_name: Optional[str] = None,
     ):
         self.session = session or bpd.get_global_session()
-        self.connection_name = connection_name or self.session._bq_connection
+
+        connection_name = connection_name or self.session._bq_connection
```
shobsi, line 53: I thought about this a bit, we are resolving between 3 potential sources of connection:

1. provided by the user to llm, i.e. `connection_name` arg
2.  defined by the user in the session, i.e. `self.session._bq_connection`
3. the default defined in `clients.py`

How about we set the default in  `Session.__init__` like below?

```python
self._bq_connection = context.bq_connection or "bigframes-default-connection"
```

We are assuming other session defaults there, such as `self._location = "US"`.

bigframes/clients.py
shobsi, line 169: [nit] since it takes a connection and returns a connection after resolving the format, would prefer naming it `resolve_full_connection_name`

tests/system/small/test_remote_function.py
shobsi: I think it's okay, the goal of this test is to assert that default connection takes effect irrespective of the session.

We can add a separate large test `test_remote_function_direct_defaults_to_global_session` specifically for that change.

bigframes/clients.py
GarrettWu: Optional only means it can be value None. It is OK to put it at the front, as it is the most important param of the function.

Not really "optional", which means it has a default value. Then it has to be at the end.

bigframes/clients.py
GarrettWu: That could work. Will address in another PR.

GarrettWu: Will address in another PR.

bigframes/ml/llm.py
GarrettWu: Oh yea, I did this way at first place, but then I found remote_function doesn't always have a session object. So I had to put the default away from session. Then we added default session... 

We can move it to session.

PR #92, opened 2023-10-06
refactor: all ArrayValue ops return only ArrayValue

TrevorBergeron:


bigframes/core/joins/name_resolution.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class JoinNameRemapper:
+    def __init__(self, namespace: str) -> None:
+        self._namespace = namespace
+
+    def __call__(
+        self, left_column_ids: Sequence[str], right_column_ids: Sequence[str]
+    ) -> Tuple[Mapping[str, str], Mapping[str, str]]:
+        """When joining column ids from different namespaces, this function defines how names are remapped. Map only non-hidden ids."""
```
tswast, line 26: I see here you warn to "Map only non-hidden IDs" but we are passing in some hidden columns in bigframes/core/joins/row_identity.py

tswast (commented):
I like it! One question.

bigframes/core/joins/name_resolution.py
TrevorBergeron: Adjusted code comment to reflect the real constraint: hidden ids should be mapped in a separate namespace as they are compile context-dependent, while value columns names must be fixed regardless of compilation modes.

tswast (approved):

PR #94, opened 2023-10-06
fix: create session dataset for remote functions only when needed

shobsi:
With this change BigFrames will not create a dataset upfront at the time of session creation, but instead leave it to the components which need the dataset to create it.


bigframes/remote_function.py
```diff
@@ -202,10 +202,22 @@ def create_bq_remote_function(
...
         logger.info(f"Creating BQ remote function: {create_function_ddl}")
+
+        # Make sure the dataset exists
+        dataset = bigquery.Dataset(
+            bigquery.DatasetReference.from_string(
+                self._bq_dataset, default_project=self._gcp_project_id
+            )
+        )
+        dataset.location = self._bq_location
+        self._bq_client.create_dataset(dataset, exists_ok=True)
```
jiaxunwu, line 215: What happens if the dataset exists while the user doesn't have dataset creation permission?

bigframes/remote_function.py
```diff
@@ -202,10 +202,22 @@ def create_bq_remote_function(
             OPTIONS (
               endpoint = "{endpoint}"
             )"""
+
         logger.info(f"Creating BQ remote function: {create_function_ddl}")
+
+        # Make sure the dataset exists
```
jiaxunwu, line 208: Any tests needed here?

TrevorBergeron (approved):

bigframes/remote_function.py
shobsi: Let me test that. Btw, this is happening in the remote function flow, where a lot more permissions are required, so requiring  `bigquery.datasets.create` shouldn't bother us

shobsi: > What happens if the dataset exists while the user doesn't have dataset creation permission?

tested that they still get the error

bigframes/remote_function.py
shobsi: There are large system tests where we provide a different dataset explicitly than the one that was created in the session. https://github.com/googleapis/python-bigquery-dataframes/blob/main/tests/system/large/test_remote_function.py

jiaxunwu (approved):

bigframes/remote_function.py
```diff
@@ -465,17 +477,20 @@ def get_remote_function_specs(self, remote_function_name):
...
+                    # TODO(shobs): Use first class properties when they are available
+                    # https://github.com/googleapis/python-bigquery/issues/1552
+                    rf_options = routine._properties.get("remoteFunctionOptions")
+                    if rf_options:
+                        http_endpoint = rf_options.get("endpoint")
+                        bq_connection = rf_options.get("connection")
+                        if bq_connection:
+                            bq_connection = os.path.basename(bq_connection)
+                    break
+        except google.api_core.exceptions.NotFound:
```
tswast, line 492: It's not 100% clear to me in what situations this can occur. Please add a comment explaining when this NotFound can happen.

I'm guessing it's because the dataset is not found and `list_routines` delays the API request until we start iterating?

tswast:
```suggestion
        except google.api_core.exceptions.NotFound:
            # The dataset might not exist, in which case the http_endpoint doesn't, either.
            # Note: list_routines doesn't make an API request until we iterate on the response object.
```

tswast (approved):
LGTM with one suggested change.

PR #101, opened 2023-10-12
feat!: rename `bigframes.pandas.reset_session` to `close_session`

GarrettWu:
BEGIN_COMMIT_OVERRIDE
feat!: rename `bigframes.pandas.reset_session` to `close_session` (#101)

Release-As: 0.9.0
END_COMMIT_OVERRIDE

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -63,7 +63,7 @@ auto-populates ``bf.options.bigquery.location`` if the user starts with
 directly or in a SQL statement.
 
 If you want to reset the location of the created DataFrame or Series objects,
-you can reset the session by executing ``bigframes.pandas.reset_session()``.
+you can close the session by executing ``bigframes.pandas.close_session()``.
 After that, you can reuse ``bigframes.pandas.options.bigquery.location`` to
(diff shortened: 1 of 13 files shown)


shobsi (approved):

tswast (approved):

tswast:
Since we need to manually specify the version number anyway (https://github.com/googleapis/python-bigquery-dataframes/pull/99#issuecomment-1759792419), let's label this as a breaking change. https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer

tswast (changes requested):
Let's mark as a breaking change.

Alternatively, we can leave an alias for reset_session but mark it as deprecated.

GarrettWu:
> Let's mark as a breaking change.
> 
> Alternatively, we can leave an alias for reset_session but mark it as deprecated.

I only need to say feat! instead of feat in commit message, right? Done.

tswast (approved):

PR #110, opened 2023-10-13
refactor: ArrayValue is now a tree that defers conversion to ibis

TrevorBergeron:


bigframes/core/__init__.py
```diff
@@ -49,470 +40,196 @@
...
-
-    grouping_keys: typing.Sequence[str] = tuple()
-    ordering: typing.Sequence[OrderingColumnReference] = tuple()
-    preceding: typing.Optional[int] = None
-    following: typing.Optional[int] = None
-    min_periods: int = 0
-
-
-# TODO(swast): We might want to move this to it's own sub-module.
 class ArrayValue:
```
tswast, line 43: Maybe keep a docstring explaining this represents an ordered heterogenous data for now?

bigframes/core/__init__.py
```diff
@@ -49,470 +40,196 @@
...
-        self._hidden_ordering_column_names = {
-            column.get_name(): column for column in self._hidden_ordering_columns
-        }
-        ### Validation
-        value_col_ids = self._column_names.keys()
-        hidden_col_ids = self._hidden_ordering_column_names.keys()
+    @classmethod
+    def from_pandas(cls, pd_df: pandas.DataFrame):
+        # Need to make immutable and hashable
+        local_array = tuple(tuple(val) for _, val in pd_df.items())  # column-major
```
tswast, line 67: This is going to be quite slow. To be consistent with how we do load jobs, let's create a Parquet file (in memory) using https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_parquet.html

Something like:

```
import io

buffer = io.BytesIO()
df.to_parquet(buffer)
local_array = buffer. getvalue()
```

bigframes/core/__init__.py
```diff
@@ -49,470 +40,196 @@
...
-                    bigframes.dtypes.ibis_dtype_to_bigframes_dtype(column_type)
-                )
-                # TODO(swast): Ibis memtable doesn't use backticks in struct
-                # field names, so spaces and other characters aren't allowed in
-                # the memtable context. Blocked by
-                # https://github.com/ibis-project/ibis/issues/7187
-                column = f"col_{column_index}"
-            new_schema.append((column, new_type))
+    def compile(self) -> compiled.CompiledArrayValue:
+        return compiled.ArrayValueCompiler().compile(self)
```
tswast, line 87: Rather than make this a stateless class, let's just have some functions in a module.

bigframes/core/__init__.py
```diff
@@ -49,470 +40,196 @@
...
 
-        # must set non-null column labels. these are not the user-facing labels
-        pd_df = pd_df.set_axis(
-            [column for column, _ in new_schema],
-            axis="columns",
-        )
-        keys_memtable = ibis.memtable(pd_df, schema=ibis.schema(new_schema))
+    def shape(self) -> typing.Tuple[int, int]:
```
tswast, line 89: Let's add something to our backlog to try and get rid of `shape` at this layer. I think it makes some caching more difficult in future.

bigframes/core/__init__.py
```diff
@@ -49,470 +40,196 @@
...
+        return (length, width)
 
-    @property
-    def columns(self) -> typing.Tuple[ibis_types.Value, ...]:
-        return self._columns
-
-    @property
-    def column_ids(self) -> typing.Sequence[str]:
-        return tuple(self._column_names.keys())
+    def _uniform_sampling(self, fraction: float) -> ArrayValue:
```
tswast, line 111: TODO: move to be by the other node operators. (Comment separating sections)

bigframes/core/__init__.py
```diff
@@ -49,470 +40,196 @@
...
-        # more node types for that to be useful), we create a builder class.
-        return ArrayValueBuilder(
-            self._session,
-            self._table,
-            columns=self._columns,
-            hidden_ordering_columns=self._hidden_ordering_columns,
-            ordering=self._ordering,
-            predicates=self._predicates,
+    def cached(self, cluster_cols: typing.Sequence[str]) -> ArrayValue:
+        """Write the ArrayValue to a session table and create a new block object that references it."""
```
tswast, line 157: Nit: "block" is a different class :-)

bigframes/core/__init__.py
```diff
@@ -49,470 +40,196 @@
...
         )
 
-    def _get_ibis_column(self, key: str) -> ibis_types.Value:
-        """Gets the Ibis expression for a given column."""
-        if key not in self.column_ids:
-            raise ValueError(
-                "Column name {} not in set of values: {}".format(key, self.column_ids)
-            )
-        return typing.cast(ibis_types.Value, self._column_names[key])
+    def _get_table_size(self, destination_table):
```
tswast, line 179: Who calls this? Maybe we can have a separate refactor PR to delete `_get_table_size()` from ArrayValue before we do this PR.

bigframes/core/__init__.py
```diff
@@ -49,470 +40,196 @@
...
-        if key not in self._hidden_ordering_column_names.keys():
-            raise ValueError(
-                "Column name {} not in set of values: {}".format(
-                    key, self._hidden_ordering_column_names.keys()
-                )
-            )
-        return typing.cast(ibis_types.Column, self._hidden_ordering_column_names[key])
+    def drop_columns(self, columns: Iterable[str]) -> ArrayValue:
+        return ArrayValue(
+            nodes.DropColumnsNode(child=self.node, columns=tuple(columns))
```
tswast, line 186: Add to backlog: just one Selection node type. (Though the ibis devs seemed to be going the other direction)

Might be good to follow substrait's architecture for this.

bigframes/core/blocks.py
```diff
@@ -1029,13 +1030,13 @@ def summarize(
             for col_id in column_ids
         ]
         columns = [
-            (col_id, [f"{col_id}-{stat.name}" for stat in stats])
+            (col_id, tuple([f"{col_id}-{stat.name}" for stat in stats]))
```
tswast, line 1033:
```suggestion
            (col_id, tuple(f"{col_id}-{stat.name}" for stat in stats))
```

bigframes/core/compile/compiled.py
```diff
@@ -0,0 +1,1122 @@
...
+from bigframes.core.window_spec import WindowSpec
+import bigframes.dtypes
+import bigframes.operations as ops
+import bigframes.operations.aggregations as agg_ops
+
+ORDER_ID_COLUMN = "bigframes_ordering_id"
+PREDICATE_COLUMN = "bigframes_predicate"
+
+
+# TODO(swast): We might want to move this to it's own sub-module.
```
tswast, line 48: TODO is now done


```suggestion
```

bigframes/core/compile/compiler.py
```diff
@@ -0,0 +1,202 @@
...
+import bigframes.core.compile.single_column
+import bigframes.core.nodes as nodes
+
+if typing.TYPE_CHECKING:
+    import bigframes.core
+    import bigframes.session
+
+
+@dataclasses.dataclass(frozen=True)
+class ArrayValueCompiler:
```
tswast, line 32: This class isn't necessary.

When we need a hashable context (e.g. ordering), we can add that parameter when we need it.

bigframes/core/nodes.py
```diff
@@ -0,0 +1,233 @@
...
+if typing.TYPE_CHECKING:
+    import ibis.expr.types as ibis_types
+
+    import bigframes.core.ordering as orderings
+    import bigframes.session
+
+
+@dataclass(frozen=True)
+class BigFrameNode:
+    pass
```
tswast, line 38: "pass" isn't necessary since there is a body to this class. Let's add a docstring (maybe here's where we explain the need to be hashable for all subclasses?)

```suggestion
```

bigframes/core/nodes.py
```diff
@@ -0,0 +1,233 @@
...
+    @property
+    def deterministic(self) -> bool:
+        return True
+
+    @property
+    def child_nodes(self) -> typing.Sequence[BigFrameNode]:
+        return tuple([])
+
+    @property
+    def sessions(self):
```
tswast, line 49: This walks the whole tree. I'd like something that can short-circuit that.

bigframes/core/nodes.py
```diff
@@ -0,0 +1,233 @@
...
+        sessions = []
+        for child in self.child_nodes:
+            sessions.extend(child.sessions)
+        return sessions
```
tswast, line 53: Tuple to make hashable and cacheable.

tswast (changes requested):

bigframes/core/nodes.py
tswast: Could we get away with something like "first_session" since these should all be the same object ideally.

bigframes/core/nodes.py
tswast: Or just a singular session.

TrevorBergeron: Made it a single session or None.

bigframes/core/__init__.py
TrevorBergeron: added

bigframes/core/__init__.py
TrevorBergeron: done

bigframes/core/__init__.py
TrevorBergeron: done

bigframes/core/nodes.py
TrevorBergeron: done

bigframes/core/nodes.py
TrevorBergeron: made it max 1 session and throw error if >1

bigframes/core/compile/compiler.py
TrevorBergeron: done

bigframes/core/compile/compiled.py
TrevorBergeron: done

bigframes/core/__init__.py
TrevorBergeron: created task

bigframes/core/__init__.py
TrevorBergeron: created task

bigframes/core/__init__.py
TrevorBergeron: Instead of parquet, went with feather which is supposedly cheaper to write

bigframes/core/__init__.py
TrevorBergeron: inlined this

tswast (approved):

PR #111, opened 2023-10-13
style: improve cancellation string

milkshakeiii:

### DIFF
--- a/bigframes/formatting_helpers.py
+++ b/bigframes/formatting_helpers.py
@@ -156,7 +156,8 @@ def wait_for_query_job(
     except KeyboardInterrupt:
         query_job.cancel()
         print(
-            f"Requested cancellation for {query_job.job_type.capitalize()} job {query_job.job_id} in location {query_job.location}..."
+            f"Requested cancellation for {query_job.job_type.capitalize()}"
+            f" job {query_job.job_id} in location {query_job.location}..."
         )
         # begin the cancel request before immediately rethrowing
         raise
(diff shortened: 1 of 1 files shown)


bigframes/formatting_helpers.py
```diff
@@ -155,7 +155,9 @@ def wait_for_query_job(
         raise
     except KeyboardInterrupt:
         query_job.cancel()
-        print(f"Requested cancelation for {query_job.job_type} {query_job.job_id}...")
+        print(
+            f"Requested cancellation for {query_job.job_type.capitalize()} job {query_job.job_id}..."
```
tswast, line 159: Nit: we should also include the job location. It's common for folks to give support this job ID but without the location it's not easy to look up where this job's logs are.

milkshakeiii: It looks like https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.job.QueryJob#google_cloud_bigquery_job_QueryJob_location is not very specific (for example: "US"). Do you think that's sufficient? Thanks.

tswast: Yes, that's correct. "US" means our US-multi-region. For other regions in the US, it'll be a string specific to that region, such as "us-central1"

tswast (approved):

bigframes/formatting_helpers.py
milkshakeiii: Nice, I hoped that was the case but the documentation was very minimal

"
location
str: Location where the job runs.
"

Ha.

Thanks.

PR #133, opened 2023-10-23
feat: add bigframes.options.compute.maximum_bytes_billed option that sets maximum bytes billed on query jobs

orrbradford:
-implement context manager for global options
-maximum_bytes_billed only applies to query jobs. This limitation will be set per query. Operations that trigger multiple jobs may result in total usage beyond this setting


third_party/bigframes_vendored/pandas/_config/config.py
```diff
@@ -0,0 +1,44 @@
+import contextlib
```
tswast, line 1: Needs license header.

orrbradford: Done. Add a reference to the pandas file similar to other files in the parent directory. Let me know if this works

orrbradford:
kokoro rerun

bigframes/pandas/__init__.py
```diff
@@ -175,6 +175,13 @@ def merge(
     )
 
 
+options = config.options
```
tswast, line 178: This already exists at the bottom of this file. See: https://github.com/googleapis/python-bigquery-dataframes/blob/ac44ccd3936cdb28755d2bbe16377d489f08d5e5/bigframes/pandas/__init__.py#L462

bigframes/pandas/__init__.py
```diff
@@ -175,6 +175,13 @@ def merge(
...
+"""Global :class:`~bigframes._config.Options` to configure BigQuery DataFrames."""
+
+option_context = config.option_context
```
tswast, line 181: Need to add this to `__all__`: https://github.com/googleapis/python-bigquery-dataframes/blob/ac44ccd3936cdb28755d2bbe16377d489f08d5e5/bigframes/pandas/__init__.py#L472

Would also be good to defined below near the `options` definition.

bigframes/pandas/__init__.py
orrbradford: Done

bigframes/pandas/__init__.py
orrbradford: Done

bigframes/_config/compute_options.py
```diff
@@ -0,0 +1,35 @@
...
+# limitations under the License.
+
+"""Options for displaying objects."""
+
+import dataclasses
+from typing import Optional
+
+
+@dataclasses.dataclass
+class ComputeOptions:
```
tswast, line 22: Could you add an "autoclass" for this file.

Need to update the following files for this to actually get documented:

* https://github.com/googleapis/python-bigquery-dataframes/blob/main/docs/reference/bigframes/options.rst
* https://github.com/googleapis/python-bigquery-dataframes/blob/main/docs/templates/toc.yml

tswast (commented):
Looking good, just one thing I forgot about so that we get this in the docs.

bigframes/_config/compute_options.py
orrbradford: Done

tswast (approved):

PR #136, opened 2023-10-24
feat: add `DataFrame.to_pandas_batches()` to download large `DataFrame` objects

tswast:
Builds on https://github.com/googleapis/python-bigquery-dataframes/pull/132

Towards internal issue 280662868
🦕


bigframes/core/blocks.py
```diff
@@ -412,6 +412,27 @@ def to_pandas(
...
+            yield df
+
+    def _copy_index_to_pandas(self, df: pd.DataFrame):
+        """Set the index on pandas DataFrame to match this block.
+
+        Warning: This method modifies ``df`` inplace.
+        """
+        if self.index_columns:
+            df.set_index(list(self.index_columns), inplace=True)
+            df.index.names = self.index.names  # type: ignore
```
ashleyxuu, line 434: Just curious what's the error here?

ashleyxuu (approved):

bigframes/core/blocks.py
tswast: bigframes/core/blocks.py:434: error: Incompatible types in assignment (expression has type "Sequence[Hashable]", variable has type "list[str]")  [assignment]

tswast: Even though list[str] is a sequence of hashable objects, I think maybe type covariance can't kick in. I'll try with some casts.

tswast: Oh, I had it opposite. pandas says "names" is `list[str]`, but actually I don't think that's true.

tswast: Filed https://github.com/pandas-dev/pandas-stubs/issues/804 and commented with an explanation.

PR #137, opened 2023-10-24
docs: link to ML.EVALUATE BQML page for score() methods

ashleyxuu:

### DIFF
--- a/bigframes/ml/ensemble.py
+++ b/bigframes/ml/ensemble.py
@@ -507,6 +507,11 @@ class RandomForestRegressor(
     ):
         """Calculate evaluation metrics of the model.
 
+        .. note::
+
+            We're using BigQuery ML.EVALUATE function (https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-evaluate)
+            for evaluateing model metrics.
+
         Args:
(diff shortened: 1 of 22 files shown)


bigframes/ml/ensemble.py
```diff
@@ -507,6 +507,11 @@ def score(
     ):
         """Calculate evaluation metrics of the model.
 
+        .. note::
+
+            We're using BigQuery ML.EVALUATE function (https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-evaluate)
```
tswast, line 512: "We" doesn't make sense for our intended audience. Instead, something like "Output matches that of the BigQuery ML.EVALUTE function. See: LINK HERE for the outputs relevant to this model type."

Also, please link to the relevant output format, as that's what the user needs to know. That is to say:

https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-evaluate#regression_models

or 

https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-evaluate#classification_models

etc.

tswast (changes requested):

tswast (approved):
Love this. Thanks!

PR #145, opened 2023-10-25
feat: log most recent API calls as `recent-bigframes-api-xx` labels on BigQuery jobs

ashleyxuu:


ashleyxuu:
FYI, the logic for creating labels doesn't change as https://github.com/googleapis/python-bigquery-dataframes/pull/55. I just added the `utils/log_adatper` so we can add to the `api_methods` list globally when calling the methods. Therefore, we don't need to pass down `api_name` params in each layer.

bigframes/utils/log_adapter.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def method_logger(method):
+    """Decorator that adds logging functionality to a method."""
+
+    @functools.wraps(method)
+    def wrapper(*args, **kwargs):
+        api_method_name = str(method.__name__)
+        if not api_method_name.startswith("__"):
+            add_api_method(api_method_name)
```
tswast, line 25: Arun't you missing return value here?

ashleyxuu: Done.

tests/unit/session/test_io_bigquery.py
```diff
@@ -19,15 +19,122 @@
...
+    df.max()
+    api_methods = log_adapter._api_methods
+
+    labels = io_bq.create_job_configs_labels(
+        job_configs_labels=cur_labels, api_methods=api_methods
+    )
+    expected_dict = {
+        "bigframes-api": "read_pandas",
+        "source": "bigquery-dataframes-temp",
+        "bigframes-api-2": "head",
```
tswast, line 96: I don't really understand why the count starts at 2. I would think an existing "bigframes-api" label would be totally opaque to the logic we use to add the labels.

bigframes/utils/log_adapter.py
```diff
@@ -0,0 +1,37 @@
...
+    return decorated_cls
+
+
+def method_logger(method):
+    """Decorator that adds logging functionality to a method."""
+
+    @functools.wraps(method)
+    def wrapper(*args, **kwargs):
+        api_method_name = str(method.__name__)
+        if not api_method_name.startswith("_"):
```
tswast, line 24: This isn't entirely accurate. We do want to add when it starts and ends with two `__` characters. Those are the "dunder" methods and how Python overrides operators.

bigframes/session/_io/bigquery.py
```diff
@@ -17,15 +17,45 @@
...
+            added_api_len = cur_labels_len + api_methods_len - MAX_LABELS_COUNT
+            label_values = list(api_methods)[-added_api_len:]
+
+    for i, label_value in enumerate(label_values):
+        if job_configs_labels is not None:
+            label_key = "bigframes-api-" + str(i + len(job_configs_labels))
+        else:
+            label_key = "bigframes-api-" + str(i)
+        labels[label_key] = label_value
+    return labels
```
tswast, line 56: This all seems way more complicated than it needs to be. Some functional-style code could make this a lot more understandable. Note: You'll need to import `itertools` for `chain`, which is a built-in module.

```suggestion
    # If there is no label set
    if job_configs_labels is None:
        job_configs_labels = {}

    labels = list(itertools.chain(
        job_config_labels.keys(),
        (f"bigframes-api-{i}" for i in range(len(api_methods)))
    ))
    values = list(itertools.chain(job_config_labels.values(), api_methods))
    return dict(zip(labels[:MAX_LABELS_COUNT], values[:MAX_LABELS_COUNT]))
```

tests/unit/session/test_io_bigquery.py
```diff
@@ -19,15 +19,122 @@
...
+        "bigframes-api-3": "max",
+    }
+    assert labels is not None
+    assert len(labels) == 4
+    assert labels == expected_dict
+
+
+def test_create_job_configs_labels_length_limit_met():
```
tswast, line 104: You're missing a test where the length limit is met and there aren't any current labels.

tswast (changes requested):

bigframes/core/groupby/__init__.py
```diff
@@ -30,9 +30,11 @@
 import bigframes.operations as ops
 import bigframes.operations.aggregations as agg_ops
 import bigframes.series as series
+from bigframes.utils import log_adapter
```
tswast, line 33: Please remove the "utils" package. "core" already serves the same purpose and communicates that the modules are internal.

tswast (changes requested):

bigframes/core/groupby/__init__.py
tswast: go/no-utils

bigframes/core/__init__.py
```diff
@@ -39,6 +40,8 @@
 ORDER_ID_COLUMN = "bigframes_ordering_id"
 PREDICATE_COLUMN = "bigframes_predicate"
 
+__all__ = ["class_logger", "method_logger"]
```
tswast, line 43: Let's remove this. It doesn't really make sense to re-export these methods just for a slightly smaller import. If we were to have it, we'd want to include other classes as well, such as ArrayValue.

```suggestion
```

bigframes/core/log_adapter.py
```diff
@@ -0,0 +1,38 @@
+import functools
```
tswast, line 1: You're missing a license header.

bigframes/core/log_adapter.py
```diff
@@ -0,0 +1,38 @@
...
+
+from loguru import logger
+
+_log_file_path = None
+_logger = logger
```
tswast, line 6: This variable appears to be unused.

bigframes/core/log_adapter.py
```diff
@@ -0,0 +1,38 @@
+import functools
+
+from loguru import logger
+
+_log_file_path = None
```
tswast, line 5: This variable appears to be unused.

bigframes/core/log_adapter.py
```diff
@@ -0,0 +1,38 @@
...
+_logger = logger
+_api_methods = []
```
tswast, line 7: I'm confused by this. Wouldn't we want to store the `_api_methods` on the individual objects. What happens when we have two different objects? I don't think we want to log all methods ever in the API request, just the ones used to generate the object.

bigframes/core/log_adapter.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def class_logger(decorated_cls):
+    """Decorator that adds logging functionality to each method of the class."""
+    for attr_name, attr_value in decorated_cls.__dict__.items():
+        if callable(attr_value):
+            setattr(decorated_cls, attr_name, method_logger(attr_value))
+    return decorated_cls
```
tswast, line 15: This would be where we'd override the initializer, I think to create an `_api_methods` attribute, potentially combining the `_api_methods` from objects passed into the constructor.

bigframes/core/log_adapter.py
```diff
@@ -0,0 +1,38 @@
+import functools
+
+from loguru import logger
```
tswast, line 3: Doesn't look like we actually use loguru. Let's remove that dependency if so.

tswast (changes requested):

third_party/bigframes_vendored/cpython/Lib/functools.py
```diff
@@ -0,0 +1,101 @@
+"""functools.py - Tools for working with functions and callable objects
```
tswast, line 1: Why are we vendoring this? Just use the one built-in to Python, right?

bigframes/session/_io/bigquery.py
```diff
@@ -30,9 +31,28 @@
...
+    api_methods: Sequence[str],
+) -> Dict[str, str]:
+    # If there is no label set
+    if job_configs_labels is None:
+        job_configs_labels = {}
+
+    labels = list(
+        itertools.chain(
+            job_configs_labels.keys(),
+            (f"bigframes-api-{i}" for i in range(len(api_methods))),
```
tswast, line 49: I wonder if we actually want to reverse the `api_methods` list, or make sure we always add the latest API to the front?

bigframes/core/log_adapter.py
tswast, line 7: Need to protect this with a lock.

bigframes/core/log_adapter.py
```diff
@@ -0,0 +1,38 @@
...
+            return result
+        except Exception as e:
+            raise e
+
+    return wrapper
+
+
+def add_api_method(method: str):
+    global _api_methods
+    _api_methods.append(method)
```
tswast, line 38: Make sure we use a lock before modifying this. Push to the front and keep the list of APIs less than our maximum length.

bigframes/session/__init__.py
```diff
@@ -1506,8 +1508,14 @@ def _start_query(
...
         Starts query job and waits for results.
         """
+        api_methods = log_adapter._api_methods
         job_config = self._prepare_job_config(job_config)
+        job_config.labels = bigframes_io.create_job_configs_labels(
+            job_configs_labels=job_config.labels, api_methods=api_methods
+        )
         query_job = self.bqclient.query(sql, job_config=job_config)
+        # Clear out the global api logger
+        log_adapter._api_methods = []
```
tswast, line 1518: This all needs to be protected by a lock. Alternatively, let's make a `log_adapter.get_and_reset_api_methods()` that locks correctly and returns a copy of the previous list of api methods.

bigframes/session/_io/bigquery.py
tswast, line 49: Since these might not actually relate to the actual job being run, let's rename the labels to `recent-bigframes-api-{i}`.

tswast (changes requested):

bigframes/core/log_adapter.py
```diff
@@ -0,0 +1,65 @@
...
+    def wrapper(*args, **kwargs):
+        api_method_name = str(method.__name__)
+        # Track regular and "dunder" methods
+        if api_method_name.startswith("__") or not api_method_name.startswith("_"):
+            add_api_method(api_method_name)
+        try:
+            result = method(*args, **kwargs)
+            return result
+        except Exception as e:
+            raise e
```
tswast, line 45: Since we aren't doing anything with the exception, we don't really need to re-raise, right? We can add this back in future if we want to start logging client-side exceptions.

```suggestion
        return method(*args, **kwargs)
```

bigframes/session/_io/bigquery.py
```diff
@@ -17,17 +17,37 @@
...
 IO_ORDERING_ID = "bqdf_row_nums"
+MAX_LABELS_COUNT = 64
 TEMP_TABLE_PREFIX = "bqdf{date}_{random_id}"
 
 
+def create_job_configs_labels(
+    job_configs_labels: Optional[Dict[str, str]],
+    api_methods: Sequence[str],
+) -> Dict[str, str]:
+    # If there is no label set
```
tswast, line 37: This comment is just describing _what_ the code does, not _why_. See: go/comments

```suggestion
```

tests/unit/core/test_log_adapter.py
```diff
@@ -0,0 +1,44 @@
+import pytest
```
tswast, line 1: Need license header.

tswast (changes requested):

tests/unit/test_compute_options.py
```diff
@@ -26,5 +26,6 @@ def test_maximum_bytes_option():
         for call in session.bqclient.method_calls:
             _, _, kwargs = call
             num_query_calls += 1
-            assert kwargs["job_config"].maximum_bytes_billed == 10000
+            if "job_config" in kwargs:
+                assert kwargs["job_config"].maximum_bytes_billed == 10000
```
tswast, line 30: I find this concerning. Please revert.

tswast (approved):

PR #149, opened 2023-10-26
feat: add pd.get_dummies

milkshakeiii:
- Emulate most aspects of the pandas get_dummies interface
- Tests and doctest examples
- Performance bottleneck is BigQuery column count in most cases.


third_party/bigframes_vendored/pandas/core/reshape/encoding.py
```diff
@@ -0,0 +1,64 @@
...
+        dictionary as with prefix.
+
+      dummy_na (bool, default False):
+        Add a column to indicate NaNs, if False NaNs are ignored.
+
+      columns (list-like, default None):
+        Column names in the DataFrame to be encoded. If columns is None
+        then only the columns with string dtype will be converted.
+
+      sparse (bool, default False):
```
shobsi, line 49: We should omit arguments that we don't implement

tests/system/small/test_pandas.py
```diff
@@ -45,6 +45,56 @@ def test_concat_series(scalars_dfs):
...
+        scalars_df,
+        prefix=prefix,
+        prefix_sep=prefix_sep,
+        dummy_na=dummy_na,
+        columns=columns,
+        drop_first=drop_first,
+    )
+    pd_result = pd.get_dummies(
+        scalars_pandas_df,
+        prefix=prefix,
```
shobsi, line 70: [nit] would be more assured (and concise) to define once and use twice:

```
args = {
        'prefix' : prefix,
        'prefix_sep' : prefix_sep,
        'dummy_na' : dummy_na,
        'columns' : columns,
        'drop_first' : drop_first,
}

bf_result = bpd.get_dummies(scalars_df, **args)
pd_result = pd.get_dummies(scalars_pandas_df, **args)
```

tests/system/small/test_pandas.py
```diff
@@ -45,6 +45,56 @@ def test_concat_series(scalars_dfs):
...
+        prefix_sep=prefix_sep,
+        dummy_na=dummy_na,
+        columns=columns,
+        drop_first=drop_first,
+    )
+
+    pd.testing.assert_frame_equal(
+        bf_result.to_pandas(), pd_result, check_dtype=False
+    )  # dtype differences are expected
```
shobsi, line 79: I prefer the approach where we modify the pd_result as per the expected type, and then assert more strongly e.g. https://github.com/googleapis/python-bigquery-dataframes/blob/8e4451841ba09099b0ed5433f9102511741dfbed/tests/system/small/test_dataframe.py#L434

tests/system/small/test_pandas.py
```diff
@@ -45,6 +45,56 @@ def test_concat_series(scalars_dfs):
     pd.testing.assert_series_equal(bf_result, pd_result)
 
 
+@pytest.mark.parametrize(
+    ("args"),
```
shobsi, line 49: Can directly specify `("prefix', "prefix_sep", "dummy_na", "columns", "drop_first")`

third_party/bigframes_vendored/pandas/core/reshape/encoding.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    Convert categorical variable into dummy/indicator variables.
+
+    Each variable is converted in as many 0/1 variables as there are
+    different values. Columns in the output are each named after a value;
+    if the input is a DataFrame, the name of the original variable is
+    prepended to the value.
+
+    Args:
+      data (array-like, Series, or DataFrame):
```
shobsi, line 29: Need to update this to say only Series and DataFrame. 

In general, check if the documentation matches the level of implementation

third_party/bigframes_vendored/pandas/core/reshape/encoding.py
milkshakeiii: Hm, I suppose arguments that now and always will do nonthing are different from arguments where we currently only support one value or similar. Removed, thanks!

third_party/bigframes_vendored/pandas/core/reshape/encoding.py
milkshakeiii: Thanks for catching that. Fixed.

tests/system/small/test_pandas.py
milkshakeiii: Sounds good, I added a loop to change the mismatched datatypes.

tests/system/small/test_pandas.py
milkshakeiii: Good suggestion, this makes the body of the tests simpler and although the decorator is a little bit verbose it's also clear.

tests/system/small/test_pandas.py
milkshakeiii: Good point, thanks!  I took your other suggestion about using a dictionary, so the tuple isn't needed at all anymore.

bigframes/pandas/__init__.py
```diff
@@ -134,6 +138,136 @@ def cut(
...
+        default_dummy_types = [pandas.StringDtype]
+        columns = [
+            col_id
+            for col_id in block.value_columns
+            if block.expr.get_column_type(col_id) in default_dummy_types
+        ]
+
+    prefix_given = prefix is not None
+
+    def parse_prefix_kwarg(kwarg, kwarg_name):
```
TrevorBergeron, line 177: nit: would be helpful to have a return type annotation on this method to make it more obvious on a glance what type its parsing the argument into

bigframes/pandas/__init__.py
```diff
@@ -134,6 +138,136 @@ def cut(
...
+        elif isinstance(kwarg, str):
+            return [kwarg] * len(columns)
+        elif isinstance(kwarg, dict):
+            return [kwarg[column] for column in columns]
+        elif kwarg is None:
+            return None
+        else:
+            raise TypeError(f"{kwarg_name} kwarg must be a string, list, or dictionary")
+
+    prefix_sep = parse_prefix_kwarg(prefix_sep, "prefix_sep")
```
TrevorBergeron, line 194: nit: this and "prefix" are plural, so maybe "prefixes" and "prefix_seps"?

bigframes/pandas/__init__.py
```diff
@@ -134,6 +138,136 @@ def cut(
...
+    if prefix_sep is None:
+        prefix_sep = ["_"] * len(columns)
```
TrevorBergeron, line 196: nit: might be more straightforward to have this earlier on `prefix_sep = parse_prefix_kwarg(prefix_sep or "_", 'prefix_sep')`.

bigframes/pandas/__init__.py
```diff
@@ -134,6 +138,136 @@ def cut(
...
+        - len(block.value_columns)
+        - len(block.index_columns)
+        - 1
+    ) // len(columns)
+    columns_ids = []
+    full_prefixes_with_duplicity = []
+    for i in range(len(columns)):
+        label = columns[i]
+        full_prefix = prefix[i] + prefix_sep[i]
+        for col_id in block.label_to_col_id[label]:
```
TrevorBergeron, line 212: does this match pandas behavior for duplicate labels? a test or two would add confidence

bigframes/pandas/__init__.py
```diff
@@ -134,6 +138,136 @@ def cut(
...
+            columns_ids.append(col_id)
+            full_prefixes_with_duplicity.append(full_prefix)
+
+    columns_values = block._get_unique_values(columns_ids, max_unique_value)
```
TrevorBergeron, line 216: `_get_unique_values` with multiple `column_ids` gets the unique tuples of values. Here I think we just care about each column independently?

TrevorBergeron: the main issue being that if each of 3 columns has 100 unique values, that might be fine here, as that is just 300 columns after this operation. However, the _get_unique_values op could produce 1000000 unique 3-tuples, which starts getting cumbersome to download.

bigframes/pandas/__init__.py
```diff
@@ -134,6 +138,136 @@ def cut(
...
+
+    result = block.drop_columns(columns_ids)
+    for i in range(columns_values.nlevels):
+        level = columns_values.get_level_values(i).sort_values()
+        column_label = full_prefixes_with_duplicity[i]
+        column_id = columns_ids[i]
+
+        already_seen = set()
+        for value in level:
```
TrevorBergeron, line 225: possible to extract the inner loop into a helper method? perhaps `create_dummy_column`?

bigframes/pandas/__init__.py
```diff
@@ -134,6 +138,136 @@ def cut(
...
+            if pandas.isna(value):
+                continue
+            if value in already_seen:
```
TrevorBergeron, line 228: If we have generated the unique values properly, we should hit this

bigframes/pandas/__init__.py
```diff
@@ -134,6 +138,136 @@ def cut(
...
+            new_block, new_id = new_block.apply_unary_op(
+                new_id, ops.BinopPartialRight(ops.fillna_op, False)
+            )
+
+            new_block = new_block.select_column(new_id)
+            new_block = new_block.with_column_labels([new_column_label])
+            result_index, _ = result.index.join(
+                new_block.index, how="left", block_identity_join=True
+            )
+            result = result_index._block
```
TrevorBergeron, line 252: might be better just to select all the ids at once at the end instead of relying so heavily on the implicit joiner to avoid doing a 'real' join

milkshakeiii:
Thanks for the review! Working on addressing these comments today.

bigframes/pandas/__init__.py
milkshakeiii: Added a test, thanks.

third_party/bigframes_vendored/pandas/core/reshape/tile.py
```diff
@@ -1,6 +1,6 @@
 # Contains code from https://github.com/pandas-dev/pandas/blob/main/pandas/core/reshape/tile.py
 """
-Quantilization functions and related stuff
+Quantilization functions and related routines
```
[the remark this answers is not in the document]
milkshakeiii: Just realized this comment including the informal wording is copied straight from pandas, haha. To be honest I don't think I've ever thought about the word "quantilization" before so I guess I just trust whoever wrote this pandas comment, and just removed the word "stuff" in order to meet best practices for comment tone.

bigframes/pandas/__init__.py
milkshakeiii: Added

bigframes/pandas/__init__.py
milkshakeiii: Done

bigframes/pandas/__init__.py
milkshakeiii: Done

bigframes/pandas/__init__.py
milkshakeiii: I changed this to 
`columns_values = [block._get_unique_values([col_id], max_unique_value) for col_id in columns_ids]`. As you say, that should be better than having exponential growth.

bigframes/pandas/__init__.py
milkshakeiii: Removed

bigframes/pandas/__init__.py
milkshakeiii: Got it, apply_unary_op already does what I want it to do (add just 1 column to the block), so I can avoid any sort of join call here.

bigframes/pandas/__init__.py
milkshakeiii: I've tested this out and unfortunately it doesn't help simplify the code. Almost every line here has a different local variable, mostly due to all the different customizations via kwargs.

TrevorBergeron (approved):
Logic looks good, just a few suggestions on cutting down the method length a bit

bigframes/pandas/__init__.py
```diff
@@ -134,6 +138,147 @@ def cut(
...
+        if pandas.api.types.is_list_like(kwarg):
+            return list(map(str, kwarg))
+        raise TypeError(f"{kwarg_name} kwarg must be a string, list, or dictionary")
+
+    prefix_seps = parse_prefix_kwarg(prefix_sep or "_", "prefix_sep")
+    prefix_seps = typing.cast(List, prefix_seps)
+    prefixes = parse_prefix_kwarg(prefix, "prefix")
+    if prefixes is None:
+        prefixes = column_labels
+    prefixes = typing.cast(List, prefixes)
```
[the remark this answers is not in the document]
milkshakeiii: Good point, that should help clarify that we're just parsing things into lists for this whole part.

bigframes/pandas/__init__.py
```diff
@@ -134,6 +138,147 @@ def cut(
...
+
+    dummy_columns_ids = []
+    dummy_columns_labels = []
+    for i in range(len(columns_values)):
+        level = columns_values[i].get_level_values(0).sort_values()
+        column_label = full_prefixes_with_duplicity[i]
+        column_id = columns_ids[i]
+
+        skip_next = drop_first
+        for value in level:
```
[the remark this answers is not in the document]
milkshakeiii: True, fixed.

bigframes/pandas/__init__.py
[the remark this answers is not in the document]
milkshakeiii: Done

bigframes/pandas/__init__.py
```diff
@@ -134,6 +138,147 @@ def cut(
...
+    dummified_col_ids = set(columns_ids)
+    carried_over_ids = []
+    carried_over_labels = []
+    for i in range(len(original_value_columns_ids)):
+        col_id = original_value_columns_ids[i]
+        if col_id not in dummified_col_ids:
+            carried_over_ids.append(col_id)
+            carried_over_labels.append(original_value_columns_labels[i])
+    block = block.select_columns(carried_over_ids + dummy_columns_ids)
+    block = block.with_column_labels(carried_over_labels + dummy_columns_labels)
```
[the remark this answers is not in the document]
milkshakeiii: Makes sense, did that, thanks.

bigframes/pandas/__init__.py
```diff
@@ -134,6 +138,147 @@ def cut(
...
+                # in these special cases, the columns are labeled with
+                # the scalar itself, not a string
+                new_column_label = value
+
+            new_block, new_id = block.apply_unary_op(
+                column_id, ops.BinopPartialLeft(ops.eq_op, value)
+            )
+            block, new_id = new_block.apply_unary_op(
+                new_id, ops.BinopPartialRight(ops.fillna_op, False)
+            )
```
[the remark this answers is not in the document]
milkshakeiii: Ah, useful, thanks.

PR #156, opened 2023-10-27
feat: add unordered sql compilation

TrevorBergeron:

### DIFF
--- a/bigframes/core/__init__.py
+++ b/bigframes/core/__init__.py
@@ -99,7 +99,9 @@ class ArrayValue:
     def shape(self) -> typing.Tuple[int, int]:
         """Returns dimensions as (length, width) tuple."""
         width = len(self._compile().columns)
-        count_expr = self._compile()._to_ibis_expr(ordering_mode="unordered").count()
+        count_expr = (
+            self._compile_unordered()._to_ibis_expr(ordering_mode="unordered").count()
+        )
 
         # Support in-memory engines for hermetic unit tests.
         if not self.node.session:
--- a/bigframes/core/__init__.py
+++ b/bigframes/core/__init__.py
@@ -23,7 +23,8 @@ import ibis
 import ibis.expr.types as ibis_types
 import pandas
 
-import bigframes.core.compile as compiled
+import bigframes.core.compile.compiled as compiled
+import bigframes.core.compile.compiler as compiler
 import bigframes.core.guid
 import bigframes.core.nodes as nodes
 from bigframes.core.ordering import OrderingColumnReference
@@ -77,7 +78,7 @@ class ArrayValue:
 
     @property
     def column_ids(self) -> typing.Sequence[str]:
-        return self.compile().column_ids
+        return self._compile().column_ids
 
     @property
     def session(self) -> Session:
@@ -87,15 +88,18 @@ class ArrayValue:
         return self.node.session[0] if required_session else get_global_session()
 
     def get_column_type(self, key: str) -> bigframes.dtypes.Dtype:
-        return self.compile().get_column_type(key)
+        return self._compile().get_column_type(key)
 
-    def compile(self) -> compiled.CompiledArrayValue:
-        return compiled.compile_node(self.node)
+    def _compile(self) -> compiled.OrderedIR:
+        return compiler.compile_ordered(self.node)
+
+    def _compile_unordered(self) -> compiled.UnorderedIR:
+        return compiler.compile_unordered(self.node)
 
     def shape(self) -> typing.Tuple[int, int]:
         """Returns dimensions as (length, width) tuple."""
-        width = len(self.compile().columns)
-        count_expr = self.compile()._to_ibis_expr("unordered").count()
+        width = len(self._compile().columns)
+        count_expr = self._compile()._to_ibis_expr(ordering_mode="unordered").count()
 
         # Support in-memory engines for hermetic unit tests.
         if not self.node.session:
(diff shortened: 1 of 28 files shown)


bigframes/core/compile/__init__.py
```diff
@@ -13,9 +13,9 @@
 # limitations under the License.
 
 from bigframes.core.compile.compiled import CompiledArrayValue
-from bigframes.core.compile.compiler import compile_node
+from bigframes.core.compile.compiler import compile_ordered
 
 __all__ = [
-    "compile_node",
+    "compile_ordered",
```
tswast, line 19: Maybe for consistency, include the unordered one here too

bigframes/core/compile/compiled.py
```diff
@@ -45,76 +42,250 @@
...
-class CompiledArrayValue:
-    """Immutable BigQuery DataFrames expression tree.
+class CompiledArrayValue(typing.Protocol):
+    @property
+    def column_ids(self) -> typing.Sequence[str]:
+        ...
 
-    Note: Usage of this class is considered to be private and subject to change
-    at any time.
+    def to_sql(self) -> str:
```
tswast, line 50: Does this need to take more parameters?

bigframes/core/compile/compiled.py
```diff
@@ -164,112 +331,108 @@ def mem_expr_from_pandas(
...
+        to reflect this in the ibis expression.
 
-    def drop_columns(self, columns: Iterable[str]) -> CompiledArrayValue:
-        # Must generate offsets if we are dropping a column that ordering depends on
-        expr = self
-        for ordering_column in set(columns).intersection(
-            [col.column_id for col in self._ordering.ordering_value_columns]
-        ):
-            expr = self._hide_column(ordering_column)
+        * "offset_col": Zero-based offsets are generated as a column, this will
```
tswast, line 376: Can remove these since no ordering mode supported.

bigframes/core/compile/compiled.py
```diff
@@ -164,112 +331,108 @@ def mem_expr_from_pandas(
...
-    def _ibis_order(self) -> Sequence[ibis_types.Value]:
-        """Returns a sequence of ibis values which can be directly used to order a table expression. Has direction modifiers applied."""
-        return _convert_ordering_to_table_values(
-            {**self._column_names, **self._hidden_ordering_column_names},
-            self._ordering.all_ordering_columns,
-        )
+    def _to_ibis_expr(
+        self,
+        *,
+        expose_hidden_cols: bool = False,
```
tswast, line 365: Might be able to remove this parameter.

bigframes/core/compile/compiled.py
```diff
@@ -477,29 +487,148 @@ def project_ternary_op(
         col_id_3: str,
         op: ops.TernaryOp,
         output_column_id: str,
-    ) -> CompiledArrayValue:
-        """Creates a new expression based on this expression with ternary operation applied to three columns."""
+    ) -> UnorderedIR:
         value = op(
```
tswast, line 491: Looks like we can share some of these implementations across unordered / ordered with a `_maybe_hide_columns` or something "template pattern".

bigframes/core/compile/compiled.py
```diff
@@ -514,117 +643,621 @@ def aggregate(
...
+        )
+        if value_col_ids & hidden_col_ids:
+            raise ValueError(
+                f"Keys in both hidden and exposed list: {value_col_ids & hidden_col_ids}"
+            )
+        if not ordering_valid:
+            raise ValueError(f"Illegal ordering keys: {ordering.all_ordering_columns}")
+
+    @classmethod
+    def from_pandas(
```
tswast, line 836: The separation for ordered/unordered here might be premature optimatization. Maybe we can share these implementations for now?

bigframes/core/compile/__init__.py
TrevorBergeron: added

bigframes/core/compile/compiled.py
TrevorBergeron: No, if the caller called in unordered mode, it also know that it has no parameters to pass here - all parameters relate to ordering.

bigframes/core/compile/compiled.py
TrevorBergeron: This parameter is a bit tricky, as its mostly there to prevent ibis from doing to many re-projections. I don't want to risk removing this in this cl

bigframes/core/compile/compiled.py
TrevorBergeron: Thats true, changed to always initialize as ordered version, and it will just drop the ordering property if in unordered compilation mode.

bigframes/core/compile/compiled.py
TrevorBergeron: Templated out to an abstract `_set_or_replace_by_id` method that assigns ibis values to column ids.

tswast (approved):

bigframes/core/__init__.py
```diff
@@ -100,7 +100,7 @@ def _compile_unordered(self) -> compiled.UnorderedIR:
     def shape(self) -> typing.Tuple[int, int]:
         """Returns dimensions as (length, width) tuple."""
         width = len(self._compile().columns)
```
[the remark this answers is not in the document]
TrevorBergeron: done and done

milkshakeiii (approved):
Awesome

PR #158, opened 2023-10-31
feat: `to_gbq` without a destination table writes to a temporary table

tswast:
- [x] Code coverage does not decrease (if any source code was changed) In fact, should have improved some as I added unit tests for a couple related `DataFrame.to_gbq` input validation branches.
- [x] Appropriate docs were updated (if necessary) Yes, including code samples that act as system tests :-)

Fixes internal issue 307942243 🦕

### DIFF
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -223,6 +223,17 @@ class Session(
         query_job.result()  # blocks until finished
         self._session_id = query_job.session_info.session_id
 
+        # The anonymous dataset is used by BigQuery to write query results and
+        # session tables. BigQuery DataFrames also writes temp tables directly
+        # to the dataset, no BigQuery Session required. Note: there is a
+        # different anonymous dataset per location. See:
+        # https://cloud.google.com/bigquery/docs/cached-results#how_cached_results_are_stored
+        query_destination = query_job.destination
+        self._anonymous_dataset = bigquery.DatasetReference(
+            query_destination.project,
+            query_destination.dataset_id,
+        )
+
         self.bqclient.default_query_job_config = bigquery.QueryJobConfig(
             connection_properties=[
                 bigquery.ConnectionProperty("session_id", self._session_id)
--- a/tests/unit/test_pandas.py
+++ b/tests/unit/test_pandas.py
@@ -116,7 +116,7 @@ def test_pandas_attribute():
     assert bpd.ArrowDtype is pd.ArrowDtype
 
 
-def test_close_session_after_bq_session_ended(monkeypatch):
+def test_close_session_after_bq_session_ended(monkeypatch: pytest.MonkeyPatch):
     bqclient = mock.create_autospec(google.cloud.bigquery.Client, instance=True)
     bqclient.project = "test-project"
     session = resources.create_bigquery_session(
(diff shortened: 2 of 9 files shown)


third_party/bigframes_vendored/pandas/core/frame.py
```diff
@@ -125,29 +125,49 @@ def to_numpy(self, dtype=None, copy=False, na_value=None, **kwargs) -> np.ndarra
...
             >>> # destination_table = PROJECT_ID + "." + DATASET_ID + "." + TABLE_NAME
             >>> df.to_gbq("bigframes-dev.birds.test-numbers", if_exists="replace")
+            'bigframes-dev.birds.test-numbers'
+
+        Write a DataFrame to a temporary BigQuery table in the anonymous dataset.
+
+            >>> df = bpd.DataFrame({'col1': [1, 2], 'col2': [3, 4]})
+            >>> destination = df.to_gbq(ordering_id="ordering_id")
+            >>> # The table created can be read outside of the current session.
+            >>> bpd.reset_session()  # For demonstration, only.
```
shobsi, line 153: Is my impression correct that we prefer `close_session`, and `reset_session` was added back only for backward compatibility?

third_party/bigframes_vendored/pandas/core/frame.py
```diff
@@ -125,29 +125,49 @@ def to_numpy(self, dtype=None, copy=False, na_value=None, **kwargs) -> np.ndarra
...
-            destination_table (str):
+            destination_table (Optional[str]):
                 Name of table to be written, in the form ``dataset.tablename``
                 or ``project.dataset.tablename``.
 
-            if_exists (str, default 'fail'):
+                If no ``destination_table`` is set, a new temporary table is
+                created in the BigQuery anonymous dataset.
+
+            if_exists (Optional[str], default 'fail'):
```
shobsi, line 170: Should update the default to ``None``

tests/unit/test_pandas.py
shobsi, line 1: Is the change in this file intended? a type annotation and a const change

bigframes/session/__init__.py
```diff
@@ -223,6 +223,17 @@ def _create_and_bind_bq_session(self):
         query_job.result()  # blocks until finished
         self._session_id = query_job.session_info.session_id
 
+        # The anonymous dataset is used by BigQuery to write query results and
+        # session tables. BigQuery DataFrames also writes temp tables directly
+        # to the dataset, no BigQuery Session required. Note: there is a
```
shobsi, line 228: Could you please help elaborate "no BigQuery Session required", isn't a session always there in BigQuery DataFrames?

tswast: It is, but I'm working on removing that. This is a step in that direction. :-)

tests/unit/test_pandas.py
tswast: Yes. This test was failing, so this change helped me debug why. The `SELECT 1` that was there before made me think the session opening logic was running, but it was really the query from this test.

third_party/bigframes_vendored/pandas/core/frame.py
tswast: Done.

shobsi (approved):

PR #160, opened 2023-10-31
test: migrate e2e presubmit tests to bigframes-load-testing project

shobsi:
BEGIN_COMMIT_OVERRIDE
test: add tools and process to set up a GCP project for running test suite (#160)

chore: migrate e2e tests to bigframes-load-testing project (#160)
END_COMMIT_OVERRIDE

Fixes internal issue 307809767 🦕


tests/system/utils.py
```diff
@@ -143,3 +143,75 @@ def convert_pandas_dtypes(df: pd.DataFrame, bytes_col: bool):
...
+        actual: Actual Pandas DataFrame
+
+        expected: Expected Pandas DataFrame
+
+        kwargs: kwargs to use in `pandas.testing.assert_series_equal` per column
+    """
+    # Compare the index, columns and values separately, as the polarity of the
+    # PCA vectors can be arbitrary
+    pd.testing.assert_index_equal(
+        actual.index, expected.index.astype(actual.index.dtype)
```
GarrettWu, line 164: Thank you for adding this. May worth mentioned index types are ignored in function docstring.

tests/system/utils.py
```diff
@@ -143,3 +143,75 @@ def convert_pandas_dtypes(df: pd.DataFrame, bytes_col: bool):
...
+    pd.testing.assert_index_equal(actual.columns, expected.columns)
+    for column in expected.columns:
+        try:
+            pd.testing.assert_series_equal(actual[column], expected[column], **kwargs)
+        except AssertionError:
+            if column not in {"numerical_value", "categorical_value"}:
+                raise
+
+            # Allow for sign difference per numeric/categorical column
+            if column == "numerical_value":
```
GarrettWu, line 175: I guess the reason we compare every column is because non-numerical columns can't apply abs() operation. 

1, For numerical columns, it is still neater to call abs() then compare than try compare and catch failures.
2, IMO it is more readable to construct the normalized DFs first then just compare them equal. Rather than write a customized compare method. Also it is easier to debug, we can directly compare the normalized DFs and see the diff.

tests/system/utils.py
```diff
@@ -143,3 +143,75 @@ def convert_pandas_dtypes(df: pd.DataFrame, bytes_col: bool):
...
+                        newarr.append(newelement)
+                    return sorted(newarr, key=lambda d: d["category"])
+
+                actual_ = actual[column].apply(normalize_array_of_objects, args=(True,))
+                expected_ = expected[column].apply(normalize_array_of_objects)
+
+            pd.testing.assert_series_equal(actual_, expected_, **kwargs)
+
+
+def assert_pandas_df_equal_pca(actual, expected, **kwargs):
```
GarrettWu, line 197: Is it that this can be replace by assert_pandas_df_equal_pca_components?

tests/system/utils.py
shobsi: The main reason not doing `abs` is that the valid PCs can differ in polarity of the whole vector, e.g. 

this is acceptable: [1, -2, 3] vs [-1, 2, -3]

but this is not: [1, -2, 3] vs [1, 2, 3]

`abs` will pass even the latter.

GarrettWu: It won't be a problem, every axis is perpendicular and independent to each other. The later is a valid output too.

shobsi: As discussed offline, below is a real example of a failing test. the second column is exactly the same, while the first and the third columns differ in polarity. With `abs` we will relax more than we need to, which would pass cases even if only some (not all) columns differ in polarity, which we don't want.

---------------------------

expected
            principal_component_1  principal_component_2  principal_component_3
tag_number                                                                     
1633                    -1.314041              -0.889106              -0.704345
1672                    -0.855813              -1.259753               0.322555
1690                    -1.848786              -0.983304              -0.095759
actual
            principal_component_1  principal_component_2  principal_component_3
tag_number                                                                     
1633                     1.314041              -0.889106               0.704345
1672                     0.855813              -1.259753              -0.322555
1690                     1.848786              -0.983304               0.095759

tests/system/utils.py
shobsi: Let me improve the docstring to reflect that `assert_pandas_df_equal_pca_components` compares the principle components themselves, while `assert_pandas_df_equal_pca` compares the projection of rows in the PC dimensions.

GarrettWu (approved):
Approved for changes of the ml module.

.kokoro/presubmit/e2e.cfg
```diff
@@ -3,5 +3,10 @@
 # Only run this nox session.
 env_vars: {
     key: "NOX_SESSION"
-    value: "unit_prerelease system_prerelease system_noextras e2e notebook samples"
```
shobsi, line 6: removing only temporarily to see if the rest of the tests pass in the new project

shobsi: restored

PR #166, opened 2023-11-02
Add at least one output per notebook.

MarkDaoust:
Hi, I'm working with @aliciawilliams on pulling some of these to the developer site. 

They're not really compatible with our import & test system. 

But if we include at least one output per notebook that will skip the tests, and allow the import to function.


tswast:
Could we add some kind of test for this so that we don't accidentally revert this change next time we refresh these notebooks?

tswast:
The notebooks have changed significantly since this PR was first open and we have a different way of testing them in our e2e tests. Closing as obsolete.

PR #109, opened 2023-10-13
fix: use table clone instead of system time for `read_gbq_table`

tswast:
BEGIN_COMMIT_OVERRIDE
chore(reverted): use table clone instead of system time for read_gbq_table
END_COMMIT_OVERRIDE

Towards internal issue 301645192 to have lower-cost default ordering.

Blocked by internal issue 305235665

Closes internal issue 301459572

PR #171, opened 2023-11-04
revert: use table clone instead of system time for `read_gbq_table` (#109)

tswast:
This reverts commit 031f253890f8a212309097554c3462980654201d.


jiaxunwu (approved):

ashleyxuu (approved):

tswast:
For context, the reason we reverted #109 is that we very quickly hit the 1000 table clones per table limit. https://cloud.google.com/bigquery/quotas#table_clones

PR #172, opened 2023-11-04
fix: match pandas behavior when assigning listlike to empty dfs

milkshakeiii:
Previously, we weren't able to assign to empty dfs at all, which is certainly not right. This change also matches the behavior for when there is a multi-index but with empty index columns. The pandas behavior is to throw an error with a somewhat odd error message. (I made our error message explicit.)

Note: this bug also occurs for assigning series to empty dfs. However, in that case, fixing it would come at the cost of adding a query to check if there are no rows. That's a big cost, so I haven't fixed the series case.

### DIFF
--- a/bigframes/dataframe.py
+++ b/bigframes/dataframe.py
@@ -1100,50 +1100,48 @@ class DataFrame(vendored_pandas_frame.DataFrame):
             copy[k] = v(copy)
             return copy
         elif utils.is_list_like(v):
-            given_rows = len(v)
-            actual_rows = len(self)
-            assigning_to_empty_df = len(self.columns) == 0 and actual_rows == 0
-            if not assigning_to_empty_df and given_rows != actual_rows:
-                raise ValueError(
-                    f"Length of values ({given_rows}) does not match length of index ({actual_rows})"
-                )
-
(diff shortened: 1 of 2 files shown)


bigframes/dataframe.py
```diff
@@ -1100,37 +1100,48 @@ def _assign_single_item(
...
-            # since the original list came from memory and so is probably < MAX_INLINE_DF_SIZE
 
-            new_column_block = local_df._block
-            original_index_column_ids = self._block.index_columns
-            self_block = self._block.reset_index(drop=False)
-            result_index, (get_column_left, get_column_right) = self_block.index.join(
+        local_df = bigframes.dataframe.DataFrame(
+            {k: v}, session=self._get_block().expr.session
+        )
+        # local_df is likely (but not guarunteed) to be cached locally
```
tswast, line 1119: Nit: spelling

```suggestion
        # local_df is likely (but not guaranteed) to be cached locally
```

tswast (approved):

bigframes/dataframe.py
milkshakeiii: Thank you!

PR #176, opened 2023-11-06
feat: add 'cross' join support

TrevorBergeron:

### DIFF
--- a/tests/system/small/test_dataframe.py
+++ b/tests/system/small/test_dataframe.py
@@ -886,6 +886,26 @@ def test_df_isin_dict(scalars_dfs):
     pandas.testing.assert_frame_equal(bf_result, pd_result.astype("boolean"))
 
 
+def test_df_cross_merge(scalars_dfs):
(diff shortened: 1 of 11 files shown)


tests/system/small/test_dataframe.py
```diff
@@ -1740,6 +1755,8 @@ def test_join_duplicate_columns_raises_not_implemented(scalars_dfs):
 
 @all_joins
 def test_join_param_on(scalars_dfs, how):
+    if how == "cross":
+        pytest.skip("cross join not supported with 'on' param")
```
shobsi, line 1759: probably more meaningful to assert with `pytest.raises` in this case

shobsi (commented):
LGTM with one minor comment

tests/system/small/test_dataframe.py
TrevorBergeron: done

shobsi (approved):

PR #178, opened 2023-11-07
chore: update formatter in CONTRIBUTING.rst

tswast:


tswast:
Looks like owlbot is still controlling this file. I'll make the change there.

CONTRIBUTING.rst
```diff
@@ -96,9 +96,9 @@ On Debian/Ubuntu::
 Coding Style
 ************
 - We use the automatic code formatter ``black``. You can run it using
-  the nox session ``blacken``. This will eliminate many lint errors. Run via::
+  the nox session ``format``. This will eliminate many lint errors. Run via::
 
-   $ nox -s blacken
```
parthea: Will the owlbot file be updated to use `format` instead of `blacken` to reflect the changes here?

https://github.com/googleapis/python-bigquery-dataframes/blob/37914a4077c681881491f5c36d1a9c9f4255e18f/owlbot.py#L114

tswast: Good point. I should really do that too. That way we get more than just "black", we'll also get the import sorting.

tswast (commented):
remove for loop to format code samples. These are managed by the root noxfile.

parthea (approved):

PR #184, opened 2023-11-08
fix: increase recursion limit, cache compilation tree hashes

TrevorBergeron:


bigframes/pandas/__init__.py
```diff
@@ -647,6 +648,9 @@ def read_gbq_function(function_name: str):
 close_session = global_session.close_session
 reset_session = global_session.close_session
 
+# SQL Compilation uses recursive algorithms on deep trees
+# 10M tree depth should be sufficient to generate any sql that is under bigquery limit
+sys.setrecursionlimit(10000000)
```
tswast, line 653: Let's get the current so that we aren't decreasing the limit (though I expect that to be unlikely)

bigframes/core/nodes.py
```diff
@@ -65,6 +65,10 @@ def session(self):
             return sessions[0]
         return None
 
+    @functools.cached_property
+    def _node_hash(self):
```
tswast, line 69: Let's document this (needed for cached_property and to avoid infinite loop)

tests/system/small/test_dataframe.py
```diff
@@ -3376,3 +3376,10 @@ def test_df_dot_operator_series(
         bf_result,
         pd_result,
     )
+
+
+def test_recursion_limit(scalars_df_index):
```
tswast, line 3381: Might be able to make this a unit test, but as discussed, it is good to exercise ibis here too.

bigframes/core/nodes.py
TrevorBergeron: added comment explaining need for this hash impl

bigframes/pandas/__init__.py
TrevorBergeron: done, now doing max for 10000000 and current recursion limit

tswast (approved):

PR #186, opened 2023-11-08
feat: support `date_series.astype("string[pyarrow]")` to cast DATE to STRING

ajpotts01:
Adds support for `date_series.astype("string[pyarrow]") to cast DATE to STRING.

As discussed with @tswast in a BigFrames feedback submission - column in source table was `date`, but wanted BigFrames to read it and save in another table as `string`. Thanks for the assist and encouragement.


tswast (approved):
Thanks so much!

PR #191, opened 2023-11-09
feat: read_gbq creates order deterministically without table copy

TrevorBergeron:

### DIFF
--- a/bigframes/dataframe.py
+++ b/bigframes/dataframe.py
@@ -2719,7 +2719,8 @@ class DataFrame(vendored_pandas_frame.DataFrame):
         return self._block
 
     def _cached(self) -> DataFrame:
-        return DataFrame(self._block.cached())
+        self._set_block(self._block.cached())
+        return self
 
     _DataFrameOrSeries = typing.TypeVar("_DataFrameOrSeries")
 
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -36,6 +36,7 @@ from typing import (
     Tuple,
     Union,
 )
+import uuid
 import warnings
 
 import google.api_core.client_info
@@ -517,6 +518,14 @@ class Session(
         column(s), then return those too so that ordering generation can be
         avoided.
         """
+        if table_ref.dataset_id.upper() == "_SESSION":
+            # _SESSION tables aren't supported by the tables.get REST API.
+            return (
+                self.ibis_client.sql(
+                    f"SELECT * FROM `_SESSION`.`{table_ref.table_id}`"
+                ),
+                None,
+            )
         table_expression = self.ibis_client.table(
             table_ref.table_id,
             database=f"{table_ref.project}.{table_ref.dataset_id}",
@@ -829,7 +838,7 @@ class Session(
         job_config.clustering_fields = cluster_cols
         job_config.labels = {"bigframes-api": api_name}
 
-        load_table_destination = bigframes_io.random_table(self._anonymous_dataset)
+        load_table_destination = self._create_session_table()
         load_job = self.bqclient.load_table_from_dataframe(
             pandas_dataframe_copy,
             load_table_destination,
@@ -1129,6 +1138,13 @@ class Session(
                 "It is recommended to use engine='bigquery' "
                 "for large files to avoid loading the file into local memory."
             )
+
+    def _create_session_table(self) -> bigquery.TableReference:
+        table_name = f"{uuid.uuid4().hex}"
+        dataset = bigquery.Dataset(
+            bigquery.DatasetReference(self.bqclient.project, "_SESSION")
+        )
+        return dataset.table(table_name)
 
     def _create_empty_temp_table(
         self,
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -538,7 +538,7 @@ class Session(
         table = self.bqclient.get_table(table_ref)
 
         if enforce_region:
-            if table.location != self._location:
+            if table.location.casefold() != self._location.casefold():
                 raise ValueError(
                     f"Current session is in {self._location} but dataset '{table.project}.{table.dataset_id}' is located in {table.location}"
                 )
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -82,6 +82,7 @@ import bigframes.version
 # Even though the ibis.backends.bigquery.registry import is unused, it's needed
 # to register new and replacement ops with the Ibis BigQuery backend.
 import third_party.bigframes_vendored.ibis.backends.bigquery.registry  # noqa
+import third_party.bigframes_vendored.ibis.expr.operations as vendored_ibis_ops
 import third_party.bigframes_vendored.pandas.io.gbq as third_party_pandas_gbq
 import third_party.bigframes_vendored.pandas.io.parquet as third_party_pandas_parquet
 import third_party.bigframes_vendored.pandas.io.parsers.readers as third_party_pandas_readers
@@ -1444,14 +1445,10 @@ def _can_cluster_bq(field: bigquery.SchemaField):
 def _convert_to_string(column: ibis_types.Column) -> ibis_types.StringColumn:
     # Some of these probably don't work
     col_type = column.type()
-    if col_type.is_array():
-        result = column.cast(ibis_dtypes.String(nullable=True))
-    elif col_type.is_struct():
-        result = column.cast(ibis_dtypes.String(nullable=True))
+    if col_type.is_array() or col_type.is_struct():
+        result = vendored_ibis_ops.ToJsonString(column).to_expr()
     elif col_type.is_geospatial():
         result = typing.cast(ibis_types.GeoSpatialColumn, column).as_text()
-    elif col_type.is_json():
-        result = column.cast(ibis_dtypes.String(nullable=True))
     elif col_type.is_string():
         result = column
     else:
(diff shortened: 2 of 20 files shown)


bigframes/session/__init__.py
```diff
@@ -504,11 +505,12 @@ def read_gbq_table(
             api_name="read_gbq_table",
         )
 
-    def _read_gbq_table_to_ibis_with_total_ordering(
+    def _get_snapshot_sql_and_primary_key(
         self,
         table_ref: bigquery.table.TableReference,
         *,
         api_name: str,
+        enforce_region: bool = False,
```
tswast, line 513: Could you add some docstring for this? It's not 100% clear to me what this means just from the variable name.

bigframes/session/__init__.py
```diff
@@ -666,142 +639,64 @@ def _read_gbq_table(
...
-            index_cols=index_cols,
-            index_labels=index_labels,
-            hidden_cols=hidden_cols,
-            ordering=ordering,
-            is_total_ordering=is_total_ordering,
-            api_name=api_name,
+            array_value = self._create_total_ordering(table_expression)
+
+        if col_order:
+            array_value = array_value.select_columns(tuple(col_order))
```
tswast, line 662: As we discussed in chat, we should really do the column filter before making the hash-based ordering. Any chance we can do that now? Otherwise, let's file an issue and add a TODO.

bigframes/session/__init__.py
```diff
@@ -1559,3 +1446,17 @@ def _can_cluster_bq(field: bigquery.SchemaField):
         "BOOL",
         "BOOLEAN",
     )
+
+
+def _convert_to_string(column: ibis_types.Column) -> ibis_types.StringColumn:
+    # Some of these probably don't work
```
tswast, line 1452: Should we check? Sounds like we need some targeted tests to cover the branches in this function.

tests/system/small/ml/test_forecasting.py
```diff
@@ -36,6 +36,12 @@ def test_model_predict(time_series_arima_plus_model):
     expected["forecast_timestamp"] = expected["forecast_timestamp"].astype(
         pd.ArrowDtype(pa.timestamp("us", tz="UTC"))
     )
+
+    print("pandas")
+    print(expected.to_string())
+    print("bigframes")
+    print(predictions.to_string())
```
tswast, line 43: Remove these debugging statements.

tests/system/small/test_progress_bar.py
```diff
@@ -52,14 +52,6 @@ def test_progress_bar_scalar(penguins_df_default_index: bf.dataframe.DataFrame,
     assert_loading_msg_exist(capsys.readouterr().out)
 
 
-def test_progress_bar_read_gbq(session: bf.Session, penguins_table_id: str, capsys):
```
tswast, line 55: Presumably this was causing us some trouble because of modifying global state?

Could you instead mark the test with pytest skip and add a TODO to re-enable once we add thread local context managers (internal issue 308657813)

tswast (changes requested):
Thanks! A few changes, but otherwise looking good. Always nice when `session.py` can get a little bit simpler.

bigframes/session/__init__.py
```diff
@@ -1297,34 +1167,51 @@ def _create_empty_temp_table(
...
-            cluster_cols=list(index_cols) + [default_ordering_name],
-            api_name=api_name,
+        full_row_str = (
+            str_values[0].concat(*str_values[1:])
+            if len(str_values) > 1
+            else str_values[0]
         )
-        table = self.ibis_client.table(
-            f"{table_ref.project}.{table_ref.dataset_id}.{table_ref.table_id}"
+        full_row_hash = full_row_str.hash().name(ordering_hash_part)
```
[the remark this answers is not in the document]
TrevorBergeron: Hmm, might be a bit early to be attempting to make our deterministic ordering independent of execution platform. Would be nice for testing I guess, but not only does the hash have to be the same, we have to serialize the rows to string exactly the same.

bigframes/session/__init__.py
```diff
@@ -1297,34 +1167,51 @@ def _create_empty_temp_table(
...
+        return core.ArrayValue.from_ibis(
+            self,
+            table_with_ordering,
+            columns,
+            hidden_ordering_columns=hidden_columns,
+            ordering=ordering,
         )
-        return table, ordering
 
     def _ibis_to_session_table(
```
tswast, line 1216: We can remove this method now, I believe.

TrevorBergeron: Still used by ArrayValue.cached(). Renaming to the more accurate `_ibis_to_temp_table` since the session dataset isn't being used anymore.

bigframes/session/__init__.py
TrevorBergeron: Should all work now, added a test specifically for the json case, other datatypes are covered by existing tests.

tests/system/small/ml/test_forecasting.py
TrevorBergeron: removed

tests/system/small/test_progress_bar.py
TrevorBergeron: read_gbq doesn't copy the table anymore, so no loading messages either

bigframes/session/__init__.py
TrevorBergeron: yeah, definitely want to minimize the set of columns hashed (even if it affects the ordering). done

bigframes/session/__init__.py
```diff
@@ -666,142 +639,64 @@ def _read_gbq_table(
...
+
+        value_columns = [col for col in array_value.column_ids if col not in index_cols]
+        block = blocks.Block(
+            array_value,
+            index_columns=index_cols,
+            column_labels=value_columns,
+            index_labels=index_cols,
         )
+        if max_results:
+            block = block.slice(stop=max_results)
```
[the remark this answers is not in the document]
TrevorBergeron: The problem with LIMIT is that it is non-deterministic in the absence of ordering. We would have to immediately cache the sample.

bigframes/session/__init__.py
```diff
@@ -666,142 +639,64 @@ def _read_gbq_table(
...
-        return self._read_gbq_with_ordering(
-            table_expression=table_expression,
-            col_order=col_order,
-            index_cols=index_cols,
-            index_labels=index_labels,
-            hidden_cols=hidden_cols,
-            ordering=ordering,
-            is_total_ordering=is_total_ordering,
-            api_name=api_name,
+            array_value = self._create_total_ordering(table_expression)
```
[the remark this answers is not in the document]
TrevorBergeron: done

tests/system/small/test_progress_bar.py
tswast: Ah, good point! Woohoo! 🎉

bigframes/dataframe.py
```diff
@@ -2719,7 +2719,8 @@ def _get_block(self) -> blocks.Block:
         return self._block
 
     def _cached(self) -> DataFrame:
-        return DataFrame(self._block.cached())
+        self._set_block(self._block.cached())
+        return self
```
tswast, line 2723: Wunderbar! This should help with taking better advantage of our cached data in the cases where we do call this automatically.

bigframes/session/__init__.py
```diff
@@ -534,6 +534,11 @@ def _read_gbq_table_to_ibis_with_total_ordering(
         # the same assumption and use these columns as the total ordering keys.
         table = self.bqclient.get_table(table_ref)
 
+        if table.location.casefold() != self._location.casefold():
```
tswast, line 537: Today I learned `casefold()`, thanks.

"Casefolded strings may be used for caseless matching." https://docs.python.org/3/library/stdtypes.html#str.casefold

bigframes/session/__init__.py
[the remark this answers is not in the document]
tswast: Ack. Filed issue 310257606 to track adding non-deterministic sampling methods in `read_gbq`.

bigframes/session/__init__.py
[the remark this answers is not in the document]
tswast: That's true. I recall having a very difficult time of that in the https://github.com/GoogleCloudPlatform/professional-services-data-validator project. Inconsistent floating point to string across different databases cost me a day or two.

tswast (approved):

tswast:
doctest failures:

```
FAILED third_party/bigframes_vendored/pandas/io/gbq.py::third_party.bigframes_vendored.pandas.io.gbq.GBQIOMixin.read_gbq
FAILED third_party/bigframes_vendored/pandas/io/parquet.py::third_party.bigframes_vendored.pandas.io.parquet.ParquetIOMixin.read_parquet
FAILED bigframes/pandas/__init__.py::bigframes.pandas.read_gbq
FAILED bigframes/pandas/__init__.py::bigframes.pandas.read_parquet
FAILED bigframes/session/__init__.py::bigframes.session.Session.read_gbq_table
FAILED bigframes/pandas/__init__.py::bigframes.pandas.read_gbq_query
FAILED bigframes/session/__init__.py::bigframes.session.Session.read_gbq_query
FAILED bigframes/pandas/__init__.py::bigframes.pandas.read_gbq_table
```

Let's remove the `df.head(2)` calls from samples where the data source doesn't have a guaranteed ordering.

PR #219, opened 2023-11-18
feat: add info and memory_usage methods to dataframe

TrevorBergeron:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,7 @@ dependencies = [
     "requests >=2.27.1",
     "scikit-learn >=1.2.2",
     "sqlalchemy >=1.4,<3.0dev",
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -50,6 +50,7 @@
     "requests >=2.27.1",
     "scikit-learn >=1.2.2",
     "sqlalchemy >=1.4,<3.0dev",
+    "tabulate >= 0.9",
```
GarrettWu, line 53: plz justify the License of the new dependency.

GarrettWu (approved):

setup.py
TrevorBergeron: uses permissive MIT license.

PR #222, opened 2023-11-18
fix: dataframes to_gbq now creates dataset if it doesn't exist

Genesis929:

Fixes b/295247521🦕

### DIFF
--- a/bigframes/dataframe.py
+++ b/bigframes/dataframe.py
@@ -2532,7 +2532,7 @@ class DataFrame(vendored_pandas_frame.DataFrame):
             )
 
         try:
-            self._session.bqclient.get_dataset()
+            self._session.bqclient.get_dataset(destination_dataset)
         except google.api_core.exceptions.NotFound:
             self._session.bqclient.create_dataset(destination_dataset, exists_ok=True)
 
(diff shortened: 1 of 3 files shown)


bigframes/dataframe.py
```diff
@@ -2400,11 +2407,12 @@ def to_gbq(
                 f"Valid options include None or one of {dispositions.keys()}."
             )
 
+        self._session.bqclient.create_dataset(destination_dataset, exists_ok=True)
```
shobsi, line 2410: Thanks for adding me. This would always require the user to have `datasets.create` permission. A better way would be to follow the pattern

```
try:
  bqclient.get_dataset()
except NotFound:
  bqclient.create_dataset()
```

This works even if the user just has `datasets.get` permission given the dataset already exists.

See for reference  https://github.com/googleapis/python-bigquery-dataframes/blob/9cde708bb4a94d3ba35ecdf298cc80bc5680e7b4/bigframes/remote_function.py#L216C1-L224C1

Genesis929: Done.

shobsi (approved):

PR #229, opened 2023-11-23
fix: add df snapshots lookup for `read_gbq`

ashleyxuu:

Fixes interbal issue 312493159 🦕

### DIFF
--- a/bigframes/pandas/__init__.py
+++ b/bigframes/pandas/__init__.py
@@ -518,6 +518,7 @@ def read_gbq_query(
     index_col: Iterable[str] | str = (),
     col_order: Iterable[str] = (),
     max_results: Optional[int] = None,
+    use_cache: bool = True,
 ) -> bigframes.dataframe.DataFrame:
     _set_default_session_location_if_possible(query)
     return global_session.with_default_session(
@@ -526,6 +527,7 @@ def read_gbq_query(
         index_col=index_col,
         col_order=col_order,
         max_results=max_results,
+        use_cache=use_cache,
     )
 
 
@@ -656,7 +658,7 @@ reset_session = global_session.close_session
 # Use __all__ to let type checkers know what is part of the public API.
 __all___ = [
     # Functions
-    "clear_cache" "concat",
+    "concat",
     "merge",
     "read_csv",
     "read_gbq",
--- a/bigframes/pandas/__init__.py
+++ b/bigframes/pandas/__init__.py
@@ -122,6 +122,16 @@ def concat(
 concat.__doc__ = vendored_pandas_concat.concat.__doc__
 
 
+def clear_cache(table: str):
+    return global_session.with_default_session(
+        bigframes.session.Session.clear_cache,
+        table=table,
+    )
+
+
+clear_cache.__doc__ = bigframes.session.Session.clear_cache.__doc__
+
+
 def cut(
     x: bigframes.series.Series,
     bins: int,
@@ -652,7 +662,7 @@ reset_session = global_session.close_session
 # Use __all__ to let type checkers know what is part of the public API.
 __all___ = [
     # Functions
-    "concat",
+    "clear_cache" "concat",
     "merge",
     "read_csv",
     "read_gbq",
--- a/bigframes/pandas/__init__.py
+++ b/bigframes/pandas/__init__.py
@@ -122,16 +122,6 @@ def concat(
 concat.__doc__ = vendored_pandas_concat.concat.__doc__
 
 
-def clear_cache(table: str):
-    return global_session.with_default_session(
-        bigframes.session.Session.clear_cache,
-        table=table,
-    )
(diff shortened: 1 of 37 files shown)


bigframes/session/__init__.py
```diff
@@ -213,6 +214,14 @@ def __hash__(self):
         # Stable hash needed to use in expression tree
         return hash(str(self._anonymous_dataset))
 
+    def clear_cache(self, table: str):
```
jiaxunwu, line 217: Shall we let users define the snapshot time in read_gbq instead of exposing this function? What's this function used for?

bigframes/session/__init__.py
```diff
@@ -177,6 +177,7 @@ def __init__(
         # Now that we're starting the session, don't allow the options to be
         # changed.
         context._session_started = True
+        self._df_snapshot: Dict[bigquery.TableReference, datetime.datetime] = {}
```
jiaxunwu, line 180: Just curious, is TableReference hashable to be used as the dict key?

jiaxunwu: Then could you help check if TableReference will change when there is new data coming in or anything happened to the layout (e.g., partitioning).

Can we use the full table ID/path as the key to be insensitive to such changes?

bigframes/pandas/__init__.py
```diff
@@ -652,7 +656,7 @@ def read_gbq_function(function_name: str):
 # Use __all__ to let type checkers know what is part of the public API.
 __all___ = [
     # Functions
-    "concat",
+    "clear_cache" "concat",
```
tswast, line 659: Looks like we had some leftover code here to revert.

```suggestion
    "concat",
```

bigframes/session/__init__.py
```diff
@@ -246,6 +247,7 @@ def read_gbq(
         index_col: Iterable[str] | str = (),
         col_order: Iterable[str] = (),
         max_results: Optional[int] = None,
+        use_cache: bool = True,
```
tswast, line 250: Can we pass this to `_read_gbq_query` too, please? It can be used to populate this property:

https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.job.QueryJobConfig#google_cloud_bigquery_job_QueryJobConfig_use_query_cache

bigframes/session/__init__.py
```diff
@@ -495,14 +488,26 @@ def _get_snapshot_sql_and_primary_key(
...
+                    ).result()
+                )[0][0]
+                self._df_snapshot[table_ref] = snapshot_timestamp
+        else:
+            snapshot_timestamp = list(
+                self.bqclient.query(
+                    "SELECT CURRENT_TIMESTAMP() AS `current_timestamp`",
+                    job_config=job_config,
+                ).result()
+            )[0][0]
```
tswast, line 508: I think we want to update the snapshot timestamp in this code path too. I believe that's how BQ works where it'll still populate the cache but not read from it.

bigframes/pandas/__init__.py
ashleyxuu: Done.

bigframes/session/__init__.py
ashleyxuu: Done.

bigframes/session/__init__.py
```diff
@@ -407,6 +414,7 @@ def _read_gbq_query(
                         ],
                         "job_id": [query_job.job_id if query_job else "unknown"],
                         "location": [query_job.location if query_job else "unknown"],
+                        "use_query_cache": [use_cache],
```
tswast, line 417: I don't think this is particually useful for DDM or DML queries. Let's remove.
```suggestion
```

More important is to pass `use_cache` to `_query_to_destination` and populate the QueryJobConfig.

bigframes/session/__init__.py
tswast: Unresolving. You didn't populate the `QueryJobConfig.use_query_cache` property as I requested.

bigframes/session/__init__.py
```diff
@@ -495,14 +494,19 @@ def _get_snapshot_sql_and_primary_key(
 
         job_config = bigquery.QueryJobConfig()
         job_config.labels["bigframes-api"] = api_name
-        current_timestamp = list(
-            self.bqclient.query(
-                "SELECT CURRENT_TIMESTAMP() AS `current_timestamp`",
-                job_config=job_config,
-            ).result()
-        )[0][0]
+        job_config.use_query_cache = use_cache
```
tswast, line 497: Not the `QueryJobConfig` I wanted you to update. Remove, please. This is the query we use to determine the snapshot time, so if anything we'd want to always set this to False.

tswast (changes requested):

tswast (approved):
Thanks!

PR #230, opened 2023-11-23
feat: support `Series.dot` on a DataFrame input

shobsi:

Fixes internal issue 312559139 🦕


Genesis929 (dismissed):

bigframes/series.py
```diff
@@ -680,6 +680,17 @@ def rdivmod(self, other) -> Tuple[Series, Series]:  # type: ignore
         return (self.rfloordiv(other), self.rmod(other))
 
     def __matmul__(self, other):
```
TrevorBergeron, line 682: Can we add `__rmatmul__`?

bigframes/series.py
```diff
@@ -680,6 +680,17 @@ def rdivmod(self, other) -> Tuple[Series, Series]:  # type: ignore
...
+        if isinstance(other, bigframes.dataframe.DataFrame):
+            return Series(
+                [
+                    pandas.NA if other[col].isna().any() else (self * other[col]).sum()
+                    for col in other.columns
+                ],
+                index=other.columns,
+                name=self.name,
+            )
```
TrevorBergeron, line 691: This will perform several queries, as each aggregate `.any()` operation initiates a query. I think we can do this more efficiently.

TrevorBergeron (changes requested):

bigframes/series.py
shobsi: Let me take that up as a follow up. Seems there is transpose involved so I need to understand the behavior better b/313748747

bigframes/series.py
shobsi: Send an optimization, PTAL. It doesn't apply to multi-index yet due to b/313747368, left a TODO.

OKEAMAH (approved):

bigframes/series.py
shobsi: `__rmatmul__` was already added in the repo, closing this thread.

PR #232, opened 2023-11-27
docs: add examples for dataframe.kurt, dataframe.std, dataframe.count

Genesis929:

dataframe.kurt: https://screenshot.googleplex.com/9zaLhagKdCNVu4F
dataframe.std: https://screenshot.googleplex.com/5maUCBQozbCsQCy
dataframe.count: https://screenshot.googleplex.com/7Cbgcd5BxXbageq

### DIFF
--- a/third_party/bigframes_vendored/pandas/core/frame.py
+++ b/third_party/bigframes_vendored/pandas/core/frame.py
@@ -2843,6 +2843,7 @@ class DataFrame(NDFrame):
             >>> df = bpd.DataFrame({"A": [1, 2, 3, 4, 5],
             ...                     "B": [3, 4, 3, 2, 1],
             ...                     "C": [2, 2, 3, 2, 2]})
+            >>> df
                 A	B	C
             0	1	3	2
             1	2	4	2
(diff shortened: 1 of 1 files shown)


third_party/bigframes_vendored/pandas/core/frame.py
```diff
@@ -2830,11 +2830,37 @@ def skew(self, *, numeric_only: bool = False):
...
+                A	B	C
+            0	1	3	2
+            1	2	4	2
+            2	3	3	3
+            3	4	2	2
+            4	5	1	2
+            <BLANKLINE>
+            [5 rows x 3 columns]
+
+        Calculating the kurtosis value of each column.
```
shobsi, line 2856: [nit] probably more appropriate to end the line with ":" instead of ".". Same elsewhere.

shobsi (approved):

third_party/bigframes_vendored/pandas/core/frame.py
ashleyxuu: [nit] Do we also want to add the examples for `numeric_only = True`? Same for elsewhere.

ashleyxuu (approved):

third_party/bigframes_vendored/pandas/core/frame.py
Genesis929: For numeric_only=True, in several Pandas APIs I reviewed, it seems that Pandas doesn't really provide an example for this. In the case of BigFrames, numeric_only=True almost does nothing, except it avoids an error that tells the user to set numeric_only=True. So I'm not quite sure if it's necessary.

PR #245, opened 2023-11-30
fix: fix value_counts column label for normalize=True

TrevorBergeron:

### DIFF
--- a/tests/system/small/test_dataframe.py
+++ b/tests/system/small/test_dataframe.py
@@ -3385,6 +3385,8 @@ def test_df_to_orc(scalars_df_index, scalars_pandas_df_index):
     ],
 )
 def test_df_value_counts(scalars_dfs, subset, normalize, ascending, dropna):
+    if pd.__version__.startswith("1."):
+        pytest.skip("pandas 1.x produces different column labels.")
     scalars_df, scalars_pandas_df = scalars_dfs
 
     bf_result = (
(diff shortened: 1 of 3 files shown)


tests/system/small/test_dataframe.py
```diff
@@ -3385,6 +3385,8 @@ def test_df_to_orc(scalars_df_index, scalars_pandas_df_index):
     ],
 )
 def test_df_value_counts(scalars_dfs, subset, normalize, ascending, dropna):
+    if pd.__version__.startswith("1."):
+        pytest.skip("pandas 1.x produces different column labels.")
```
shobsi, line 3389: insert update expected value and assert?

TrevorBergeron: I don't think its worth the tests complexity - we're validating against 2.x behavior and I think that is enough.

tests/system/small/test_dataframe.py
```diff
@@ -3396,10 +3398,6 @@ def test_df_value_counts(scalars_dfs, subset, normalize, ascending, dropna):
         subset, normalize=normalize, ascending=ascending, dropna=dropna
     )
 
-    # Older pandas version may not have these values, bigframes tries to emulate 2.0+
```
shobsi, line 3399: Let's remove this from test_series as well https://github.com/search?q=repo%3Agoogleapis%2Fpython-bigquery-dataframes%20%22Older%20pandas%20version%20may%20not%20have%20these%20values%22&type=code

TrevorBergeron: done

shobsi: Thanks, just one one more test `test_value_counts_w_cut` in test_series.py

TrevorBergeron: done

shobsi (approved):

shobsi (approved):

PR #255, opened 2023-12-05
refactor: move query execution from ArrayValue to Session

TrevorBergeron:


tswast (commented):
Looks like you'll need to fix some mocks in the unit tests. It appears to be trying to fetch real credentials.

TrevorBergeron:
> Looks like you'll need to fix some mocks in the unit tests. It appears to be trying to fetch real credentials.

This change removed a kludge we had to avoid invoking session at all for .shape. Added a new kludge to compensate. Longer term, we will want to figure out a better pattern for local execution in unit tests.

tswast (approved):

PR #265, opened 2023-12-09
fix: enfore pandas version requirement <2.1.4

shobsi:
There seem to be a breaking change in pandas release 2.1.4 that is failing tests using `pandas.read_json`. This change is pinning pandas dependency version to <2.1.4 until the proper fix.

Fixes internal issue 315539920 🦕


ashleyxuu (approved):
Thanks for helping debugging

PR #268, opened 2023-12-11
fix: pin prerelease tests to pandas 2.1.3 to unblock e2e tests

shobsi:
BEGIN_COMMIT_OVERRIDE
fix: exclude pandas 2.1.4 from prerelease tests to unblock e2e tests
END_COMMIT_OVERRIDE

Fixes internal issue 315852455 🦕

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -520,7 +520,7 @@ def prerelease(session: nox.sessions.Session, tests_path):
         "--upgrade",
         # TODO(shobs): Remove tying to version 2.1.3 after
         # https://github.com/pandas-dev/pandas/issues/56463 is resolved
-        "pandas==2.1.3",
+        "pandas!=2.1.4",
     )
     already_installed.add("pandas")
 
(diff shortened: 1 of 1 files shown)


tswast (approved):

noxfile.py
```diff
@@ -518,7 +518,9 @@ def prerelease(session: nox.sessions.Session, tests_path):
         "--prefer-binary",
         "--pre",
         "--upgrade",
-        "pandas",
+        # TODO(shobs): Remove tying to version 2.1.3 after
+        # https://github.com/pandas-dev/pandas/issues/56463 is resolved
+        "pandas==2.1.3",
```
parthea, line 523: Another possibility is to exclude the current version using `pandas!=2.1.4` so that we can run pre-release tests against the next release. It will be toilsome to keep updating the version but the process of updating the excluded versions could be used as a way to follow up on the issue.

shobsi: Thanks @parthea for the suggestion. Done so.

GarrettWu (approved):

PR #296, opened 2023-12-28
fix: Fix bug converting non-string labels to sql ids

TrevorBergeron:

### DIFF
--- a/bigframes/core/utils.py
+++ b/bigframes/core/utils.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import re
 import typing
 from typing import Hashable, Iterable, List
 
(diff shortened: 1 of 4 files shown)


tests/system/small/test_multiindex.py
```diff
@@ -16,7 +16,37 @@
...
+                ["Bonjour le monde!", "une chaîne de caractères"],
+                dtype=pandas.StringDtype(storage="pyarrow"),
+            ),
+        ],
+        names=["1columns 1", "1new_index 2"],
+    )
+    pandas_df = pandas.DataFrame(
+        [[1, 2], [3, 4]], index=index, columns=columns, dtype=pandas.Int64Dtype()
+    )
+    bf_df = bpd.DataFrame(pandas_df)
```
shobsi, line 47: Are we sure pandas itself doesn't do any normalization on the labels? Can we assert that before this conversion?

bigframes/core/utils.py
```diff
@@ -84,26 +85,42 @@ def get_standardized_ids(
...
 
-    ids = idx_ids + col_ids
+    ids = disambiguate_ids(idx_ids + col_ids)
+
+    idx_ids, col_ids = ids[: len(idx_ids)], ids[len(idx_ids) :]
+
+    return col_ids, idx_ids
+
+
+def label_to_identifier(label: typing.Hashable, strict: bool = False) -> str:
```
shobsi, line 103: would be nice to have unit tests on these utils methods

tests/system/small/test_multiindex.py
TrevorBergeron: pandas directly uses the passed index object without modification.

bigframes/core/utils.py
TrevorBergeron: Since utils are not part of the public API, I would prefer to test them via the public APIs that use them. I tried to include some unusual labels in the test cases.

shobsi: Sounds good, let's add these two kinds of labels to test the strict mode:

1. Leading spaces followed by alpha-numeric
2. Leading spaces followed by `_`

shobsi (approved):

PR #301, opened 2024-01-03
fix: DF.drop tuple input as multi-index

GarrettWu:

Fixes b/317131784

### DIFF
--- a/bigframes/dataframe.py
+++ b/bigframes/dataframe.py
@@ -1062,12 +1062,31 @@ class DataFrame(vendored_pandas_frame.DataFrame):
             level_id = self._resolve_levels(level or 0)[0]
 
             if utils.is_list_like(index):
-                block, inverse_condition_id = block.apply_unary_op(
-                    level_id, ops.IsInOp(index, match_nulls=True)
-                )
-                block, condition_id = block.apply_unary_op(
-                    inverse_condition_id, ops.invert_op
-                )
+                # Only tuple is treated as multi-index value combinations
+                if isinstance(index, tuple):
+                    if level is not None:
+                        raise ValueError("Multi-index tuple can't specify level.")
+                    condition_id = None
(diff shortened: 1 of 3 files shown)


bigframes/dataframe.py
```diff
@@ -1062,12 +1062,31 @@ def drop(
...
+                        raise ValueError("Multi-index tuple can't specify level.")
+                    condition_id = None
+                    for i, idx in enumerate(index):
+                        level_id = self._resolve_levels(i)[0]
+                        block, condition_id_cur = block.apply_unary_op(
+                            level_id, ops.partial_right(ops.ne_op, idx)
+                        )
+                        if condition_id:
+                            block, condition_id = block.apply_binary_op(
+                                condition_id, condition_id_cur, ops.or_op
```
shobsi, line 1077: Wondering if this should be `ops.and_op`. Does the example in https://b.corp.google.com/issues/317131784#comment1 gets fixed after this change? Would be nice to include that as an inline test case.

GarrettWu: It is or. We keep the rows based on the condition to be True, which is: none of the columns matches the drop value. The example is fixed (I pasted the link in the bug, since the we shouldn't paste link on github). It will be redundant to add the test case as the test suite covered.

tests/system/small/test_multiindex.py
```diff
@@ -213,6 +214,7 @@ def test_multi_index_drop(scalars_df_index, scalars_pandas_df_index, labels, lev
     bf_result = bf_frame.drop(labels=labels, axis="index", level=level).to_pandas()
     pd_result = pd_frame.drop(labels=labels, axis="index", level=level)
 
+    print("over")
```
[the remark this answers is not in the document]
GarrettWu: you caught me :)

shobsi (approved):

PR #302, opened 2024-01-03
feat: Allow manually set clustering_columns in dataframe.to_gbq

Genesis929:

### DIFF
--- a/bigframes/dataframe.py
+++ b/bigframes/dataframe.py
@@ -2789,7 +2789,7 @@ class DataFrame(vendored_pandas_frame.DataFrame):
             raise ValueError("Duplicates are not supported in clustering_columns")
 
         all_possible_columns = (
-            set(self.columns) | set(self.index.names) if index else set(self.columns)
+            (set(self.columns) | set(self.index.names)) if index else set(self.columns)
         )
         missing_columns = set(clustering_columns) - all_possible_columns
         if missing_columns:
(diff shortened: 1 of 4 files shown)


tests/system/small/test_dataframe_io.py
```diff
@@ -317,6 +317,51 @@ def test_to_gbq_w_None_column_names(
     )
 
 
+@pytest.mark.parametrize(
+    "clustering_columns",
+    [
+        pytest.param(["int64_col", "geography_col"]),
+        pytest.param(
+            ["float64_col"],
+            marks=pytest.mark.xfail(raises=google.api_core.exceptions.BadRequest),
```
shobsi, line 326: Worth adding one more testcase with duplicate clustering column names throwing the exception

bigframes/dataframe.py
```diff
@@ -2756,6 +2767,52 @@ def _apply_unary_op(self, operation: ops.UnaryOp) -> DataFrame:
...
+            return mapped_columns
+
+        if not clustering_columns:
+            return []
+
+        if len(list(clustering_columns)) != len(set(clustering_columns)):
+            raise ValueError("Duplicates are not supported in clustering_columns")
+
+        all_possible_columns = (
+            set(self.columns) | set(self.index.names) if index else set(self.columns)
```
shobsi, line 2792: [nit] not clear if `|` appplies first or the `if-else`, maybe an explicit parentheses would be nice

tests/system/small/test_dataframe_io.py
Genesis929: Done.

bigframes/dataframe.py
Genesis929: Done.

shobsi (approved):

PR #303, opened 2024-01-04
feat: add Series dt.tz and dt.unit properties

TrevorBergeron:

### DIFF
--- a/tests/system/small/operations/test_datetimes.py
+++ b/tests/system/small/operations/test_datetimes.py
@@ -32,8 +32,6 @@ DATE_COLUMNS = [
 )
 @skip_legacy_pandas
 def test_dt_day(scalars_dfs, col_name):
-    if pd.__version__.startswith("1."):
-        pytest.skip("Pyarrow datetime objects not support in pandas 1.x.")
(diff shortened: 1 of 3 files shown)


tests/system/small/operations/test_datetimes.py
```diff
@@ -16,16 +16,22 @@
...
 @pytest.mark.parametrize(
     ("col_name",),
-    DATETIME_COL_NAMES,
+    DATE_COLUMNS,
 )
-def test_day(scalars_dfs, col_name):
+@skip_legacy_pandas
+def test_dt_day(scalars_dfs, col_name):
     if pd.__version__.startswith("1."):
         pytest.skip("Pyarrow datetime objects not support in pandas 1.x.")
```
Genesis929, line 36: The @skip_legacy_pandas seems already skipping pandas ver 1.x,  I think this "if" will always be false? Same with test_dt_date

TrevorBergeron: good catch, removed the redundant checks

Genesis929 (approved):

PR #318, opened 2024-01-12
feat: add efficient peek dataframe preview

TrevorBergeron:


bigframes/dataframe.py
```diff
@@ -1066,6 +1066,14 @@ def head(self, n: int = 5) -> DataFrame:
     def tail(self, n: int = 5) -> DataFrame:
         return typing.cast(DataFrame, self.iloc[-n:])
 
+    def peek(self, n: int = 5) -> pandas.DataFrame:
+        maybe_result = self._block.try_peek(n)
+        if maybe_result is None:
+            raise NotImplementedError(
+                "Cannot peek efficiently when data has aggregates, joins or window functions applied."
```
tswast, line 1073: It's not always going to be obvious when the data has aggregates, windows, or especially not joins (due to implicit joins on index).

Could we instead call `_cached()` in this case and then peek?

tswast (commented):
I wonder if we can also implement this "peekable" logic (and _cached() if not) in the case of head()?

This would seem to align more with our customer's intuition, even if it causes some wasted work due to ordering.

Also, we should definitely document that they may want to use `peek()` instead of `head()` if they don't require a consistent ordering in as many places as we can.

bigframes/dataframe.py
TrevorBergeron: Added a force=True param to peek. This will cause the dataframe to execute and cache the block if it is not peekable. Are you sure this is the best default behavior? One of my goals with peek was for users to avoid fully computing the dataframe.

tswast: In general, I'd say the pandas API skews towards the most usable default, not the most efficient one. That said, our main motivation here is to address the feedback of how expensive `head()` can be, so I see the argument for having `force=False` the default.

bigframes/dataframe.py
```diff
@@ -1066,6 +1066,36 @@ def head(self, n: int = 5) -> DataFrame:
     def tail(self, n: int = 5) -> DataFrame:
         return typing.cast(DataFrame, self.iloc[-n:])
 
+    def peek(self, n: int = 5, *, force: bool = True) -> pandas.DataFrame:
```
tswast, line 1069:
```suggestion
    def peek(self, n: int = 5, *, force: bool = False) -> pandas.DataFrame:
```

Let's do `force=False` as the default for now. If we get complaints, we can reconsider.

TrevorBergeron: done, and amended docstring.

tswast (approved):

tswast (changes requested):

bigframes/dataframe.py
```diff
@@ -1066,6 +1066,36 @@ def head(self, n: int = 5) -> DataFrame:
...
+            NotImplementedError: If force=False and data cannot be efficiently peeked.
+        """
+        maybe_result = self._block.try_peek(n)
+        if maybe_result is None:
+            if force:
+                self._cached()
+                maybe_result = self._block.try_peek(n)
+                assert maybe_result is not None
+            else:
+                raise NotImplementedError(
```
tswast, line 1094: ValueError or a custom subclass of it may be more appropriate for this.

TrevorBergeron: changed to ValueError

tswast (approved):

PR #320, opened 2024-01-16
perf: define tree rewriter to simplify projections

TrevorBergeron:


TrevorBergeron:
Rewriting is causing more compilation work right now. Will put on hold until schema can be derive directly from tree rather than from compiling.

PR #321, opened 2024-01-17
refactor: simplify filter and join nodes

TrevorBergeron:


bigframes/core/blocks.py
```diff
@@ -1542,19 +1542,34 @@ def merge(
...
-            left_join_ids,
-            other.expr,
-            right_join_ids,
-            how=how,
-        )
-        get_column_left, get_column_right = join_names.JOIN_NAME_REMAPPER(
-            self.expr.column_ids, other.expr.column_ids
+        left_mappings = [
+            join_defs.JoinColumnMapping(
+                join_defs.JoinSide.LEFT, id, guid.generate_guid()
```
tswast, line 1547: Please use keyword arguments for these JoinColumnMapping constructor arguments. It's difficult for me to check correctness. We might even want to force the use of keyword arguments with the use of `*` syntax.

TrevorBergeron: ok, removed kw_only=True as that isn't allowed in old python versions, but still made all actual constructor invocations use kw args

tswast (approved):
Thanks!

PR #334, opened 2024-01-19
feat: Add Index constructor, copy, get_level_values, to_series

TrevorBergeron:
fix mypy error


bigframes/core/indexes/index.py
```diff
@@ -33,16 +34,60 @@
...
-        self._data = data
+    def __init__(
+        self,
+        data=None,
+        dtype=None,
+        *,
+        name=None,
+    ):
+        import bigframes.dataframe as df
+        import bigframes.series as series
```
Genesis929, line 58: 'm a bit curious as to why one would import a module inside a method.

TrevorBergeron: Its to avoids a circular dependency. DataFrame.py and Series.py already depend on Index.py, so cannot have top-level import of those from Index.py.

Genesis929 (approved):

PR #335, opened 2024-01-20
feat: update cut to work without labels = False and show intervals as dict

Genesis929:

Code example updates:
https://screenshot.googleplex.com/ApQyMsLXCByk7z8
https://screenshot.googleplex.com/5jFfvntpbX5djv3

### DIFF
--- a/third_party/bigframes_vendored/pandas/core/reshape/tile.py
+++ b/third_party/bigframes_vendored/pandas/core/reshape/tile.py
@@ -38,6 +38,15 @@ def cut(
 
     Cut with an integer (equal-width bins):
 
+        >>> bpd.cut(s, bins=4)
+            0    {'left_exclusive': -0.01, 'right_inclusive': 2.5}
+            1    {'left_exclusive': -0.01, 'right_inclusive': 2.5}
+            2      {'left_exclusive': 2.5, 'right_inclusive': 5.0}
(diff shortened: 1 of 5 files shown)


third_party/bigframes_vendored/pandas/core/reshape/tile.py
```diff
@@ -38,6 +38,15 @@ def cut(
...
     Cut with an integer (equal-width bins):
 
+        >>> bpd.cut(s, bins=4)
+            0    {'left_exclusive': -0.01, 'right_inclusive': 2.5}
+            1    {'left_exclusive': -0.01, 'right_inclusive': 2.5}
+            2      {'left_exclusive': 2.5, 'right_inclusive': 5.0}
+            3     {'left_exclusive': 7.5, 'right_inclusive': 10.0}
+            dtype: struct<left_exclusive: double, right_inclusive: double>[pyarrow]
+
+    Cut with an integer (equal-width bins) and labels=False:
```
shobsi, line 48: It seems `labels=True` is also acceptable. Can we include a code sample for that?

Genesis929: Was adding this to avoid the confusing case in pandas that labels can be false but can not be true, so makes labels = True works the same as labels = None. But it would appear this may be more confusing. Changed the code back so that now labels won't be able to be True.

shobsi (approved):

PR #338, opened 2024-01-22
refactor: add output type annotations to scalar ops

TrevorBergeron:

### DIFF
--- a/bigframes/dtypes.py
+++ b/bigframes/dtypes.py
@@ -23,7 +23,9 @@ from typing import Any, Dict, Iterable, Literal, Tuple, Union
 import geopandas as gpd  # type: ignore
 import google.cloud.bigquery as bigquery
 import ibis
+from ibis.backends.bigquery.datatypes import BigQueryType
 import ibis.expr.datatypes as ibis_dtypes
+from ibis.expr.datatypes.core import dtype as python_type_to_bigquery_type
 import ibis.expr.types as ibis_types
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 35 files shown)


bigframes/dtypes.py
```diff
@@ -42,6 +44,12 @@
     pd.ArrowDtype,
     gpd.array.GeometryDtype,
 ]
+ExpressionType = Union[Dtype, None]
```
shobsi, line 47: [nit] typing.Optional[Dtype] might be more concise

bigframes/operations/__init__.py
```diff
@@ -126,86 +139,101 @@ def _convert_expr_input(
...
         bases=(UnaryOp,),
         frozen=True,
     )()
 
 
-def create_binary_op(name: str) -> BinaryOp:
+def create_binary_op(
+    name: str, type_rule: op_typing.OpTypeRule = op_typing.Supertype()
+) -> BinaryOp:
+    def output_type(self, *input_types: dtypes.ExpressionType) -> dtypes.ExpressionType:
```
shobsi, line 159: This seems repetitive all create_*_op methods. Would it make sense to define outside once and reuse? Or even define in a base `class Op` and derive the *naryOp classes from that?

TrevorBergeron: deduplicated into OpTypeRule class itself. Maybe should jsut have OpTypeRule be a Callable anyways...

bigframes/dtypes.py
TrevorBergeron: Done

shobsi (approved):

PR #344, opened 2024-01-24
feat: Improve error message for drive based BQ table reads

shobsi:
This change will append a helpful documentation link to troubleshoot drive permission denied error https://screenshot.googleplex.com/96rLqtEUcrxzd39.

Fixes internal issue 304826407 🦕

### DIFF
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -249,33 +249,27 @@ class Session(
 
         query_or_table = self._filters_to_query(query_or_table, columns, filters)
 
-        try:
-
-            if _is_query(query_or_table):
-                return self._read_gbq_query(
-                    query_or_table,
-                    index_col=index_col,
-                    columns=columns,
-                    max_results=max_results,
-                    api_name="read_gbq",
-                    use_cache=use_cache,
-                )
-            else:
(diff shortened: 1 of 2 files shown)


bigframes/session/__init__.py
```diff
@@ -249,27 +249,33 @@ def read_gbq(
...
+                return self._read_gbq_table(
+                    query_or_table,
+                    index_col=index_col,
+                    columns=columns,
+                    max_results=max_results,
+                    api_name="read_gbq",
+                    use_cache=use_cache,
+                )
+        except google.api_core.exceptions.Forbidden as ex:
+            if "Permission denied while getting Drive credentials." in ex.message:
```
tswast, line 276: This could be a bit brittle. Let's only include the minimum to see if it's a Drive error.

```suggestion
            if "Drive credentials" in ex.message:
```

shobsi: Done, thanks! I also moved the exception handling deeper so that it serves mroe APIs. PTAL.

tswast (approved):

tswast:
Would be good to add a unit test for this logic to double-check that appending to the message like this doesn't cause a different exception.

PR #345, opened 2024-01-24
feat: limited support of lamdas in `Series.apply`

shobsi:
BEGIN_COMMIT_OVERRIDE
feat: limited support of lambdas in `Series.apply` (#345)
END_COMMIT_OVERRIDE

- [x] Appropriate docs were updated https://screenshot.googleplex.com/6ZEiKXPz8LWMTRf

Partially fixes internal issue 295964341 🦕


third_party/bigframes_vendored/pandas/core/series.py
```diff
@@ -1152,6 +1152,35 @@ def apply(
             >>> names = bpd.Series(["Alice", "Bob"])
             >>> hashes = names.apply(get_hash)
 
+        There is a limited support of simple functions and lambdas which can be
+        operated directly (without converting into a `remote_function`) on the
+        BigQuery DataFrames objects. This approach takes advantage of a nuance
+        in the way BigQuery DataFrames objects are modeled internally and works
+        only if the function body contains only arithmatic or logical operators.
```
tswast, line 1159: I'd rather we rephrase this as "ufunc" emulation support, as defined in the pandas docs. https://pandas.pydata.org/docs/reference/api/pandas.Series.apply.html

bigframes/core/compile/scalar_op_compiler.py
```diff
@@ -658,10 +658,6 @@ def isin_op_impl(x: ibis_types.Value, op: ops.IsInOp):
 
 @scalar_op_compiler.register_unary_op(ops.RemoteFunctionOp, pass_op=True)
 def remote_function_op_impl(x: ibis_types.Value, op: ops.RemoteFunctionOp):
-    if not hasattr(op.func, "bigframes_remote_function"):
```
tswast, line 661: This is the wrong layer at which to do this. Please instead call this on the actual Series object.

That way numpy ufuncs will also work as expected.

tswast (changes requested):

bigframes/core/compile/scalar_op_compiler.py
tswast: Also, please add tests for passing in numpy ufuncs directly in addition to the existing lambda tests.

third_party/bigframes_vendored/pandas/core/series.py
shobsi: Updated, PTAL.

bigframes/core/compile/scalar_op_compiler.py
shobsi: Added tests with select numpy ufuncs supported by bigframes

bigframes/series.py
```diff
@@ -1213,6 +1213,14 @@ def apply(self, func) -> Series:
         # Reproject as workaround to applying filter too late. This forces the filter
         # to be applied before passing data to remote function, protecting from bad
         # inputs causing errors.
+        if not callable(func):
+            raise ValueError(
+                "Only a ufunc (a NumPy function that applies to the entire Series) or a remote function that only works on single values are supported."
+            )
+
+        if not hasattr(func, "bigframes_remote_function"):
+            return func(self)
```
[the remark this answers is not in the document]
shobsi: Done, thanks!

bigframes/series.py
```diff
@@ -1216,6 +1222,22 @@ def apply(self, func) -> Series:
...
+
+        if not hasattr(func, "bigframes_remote_function"):
+            try:
+                return func(self)
+            except Exception as ex:
+                # This could happen if any of the operators in func is not
+                # supported on a Series. Let's guide the customer to use a
+                # remote function instead
+                if hasattr(ex, "message"):
+                    ex.message += "\n{_remote_function_recommendation_message}"
```
tswast, line 1238: This needs to be an `f` string.

bigframes/series.py
```diff
@@ -1216,6 +1222,22 @@ def apply(self, func) -> Series:
...
         # to be applied before passing data to remote function, protecting from bad
         # inputs causing errors.
+        if not callable(func):
+            raise ValueError(
+                "Only a ufunc (a NumPy function that applies to the entire Series) or a remote function that only works on single values are supported."
+            )
+
+        if not hasattr(func, "bigframes_remote_function"):
+            try:
+                return func(self)
```
tswast, line 1232: Since it may result in incorrect values if this isn't a true vectorized function, let's check for `by_row=False`. If `by_row="compat"` (default) then raise and suggest either remote_function or by_row=False.

bigframes/series.py
shobsi: Thanks for catching, corrected.

bigframes/series.py
shobsi: Done, PTAL.

bigframes/series.py
```diff
@@ -1210,12 +1216,41 @@ def _groupby_values(
...
+        if not callable(func):
+            raise ValueError(
+                "Only a ufunc (a function that applies to the entire Series) or a remote function that only works on single values are supported."
+            )
+
+        if not hasattr(func, "bigframes_remote_function"):
+            # It is not a remote function
+            # Then it must be a vectorized function that applies to the Series
+            # as a whole
+            assert (
```
tswast, line 1240: AssertionError is a bit of an odd one to raise. Usually that means some invariant that we never expect to happen has violated. Please raise `ValueError` instead.

shobsi: Done, PTAL, thanks.

tswast (approved):
Love it, thanks!

PR #349, opened 2024-01-25
perf: DataFrame.head materialization now materializes base DataFrame

TrevorBergeron:


bigframes/core/block_manager.py
```diff
@@ -0,0 +1,106 @@
+# Copyright 2023 Google LLC
```
tswast, line 1: Nit: We're in 2024 now.

PR #358, opened 2024-01-29
chore: pin pytest version

tswast:

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -31,7 +31,8 @@ import nox.sessions
 BLACK_VERSION = "black==22.3.0"
 ISORT_VERSION = "isort==5.12.0"
 
-# pytest-retry is not yet compatible with python 8.x.
+# pytest-retry is not yet compatible with pytest 8.x.
+# https://github.com/str0zzapreti/pytest-retry/issues/32
 PYTEST_VERSION = "pytest<8.0.0dev"
 SPHINX_VERSION = "sphinx==4.5.0"
 LINT_PATHS = ["docs", "bigframes", "tests", "third_party", "noxfile.py", "setup.py"]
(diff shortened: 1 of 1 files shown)


tswast (commented):
set max version instead of exact pin

GarrettWu (approved):

noxfile.py
```diff
@@ -30,6 +30,7 @@
 
 BLACK_VERSION = "black==22.3.0"
 ISORT_VERSION = "isort==5.12.0"
+PYTEST_VERSION = "pytest<=7.4.4"
```
parthea, line 33: nit: add a comment to capture the specific reason for adding an upper limit to `pytest` (with a link to the corresponding issue to follow up) . If there was a bad release of pytest, consider excluding the specific bad release.

tswast (commented):
comment why we set max version

noxfile.py
tswast: Added! As far as I can tell from recent test failures it's that https://github.com/str0zzapreti/pytest-retry isn't compatible with 8.x, which was released this weekend.

noxfile.py
```diff
@@ -30,6 +30,9 @@
 
 BLACK_VERSION = "black==22.3.0"
 ISORT_VERSION = "isort==5.12.0"
+
+# pytest-retry is not yet compatible with python 8.x.
```
parthea, line 34:
```suggestion
# pytest-retry is not yet compatible with IPython 8.x.
```

parthea: Please could you add a link to the upstream issue if it exists?

tswast: Filed and linked here, thanks: https://github.com/str0zzapreti/pytest-retry/issues/32

parthea (approved):

PR #360, opened 2024-01-29
refactor: generalize aggregation to handle 0,1, or 2 inputs

TrevorBergeron:


bigframes/core/compile/aggregate_compiler.py
```diff
@@ -21,31 +21,63 @@
...
 
-def compile_unary_aggregate(
-    op: agg_ops.AggregateOp, input: ibis_types.Column
+
+def compile_aggregate(
+    aggregate: ex.Aggregation,
+    bindings: typing.Dict[str, ibis_types.Value],
 ) -> ibis_types.Value:
-    return compile_agg(op, input)
+    if isinstance(aggregate, ex.UnaryAggregation):
```
tswast, line 38: Note to self: Is there another class for 0 argument aggregations (e.g. row count) that we need to handle here?

bigframes/core/compile/aggregate_compiler.py
```diff
@@ -382,6 +418,19 @@ def _(
     )
 
 
+@compile_binary_agg.register
+def _(
```
tswast, line 422: Not a change introduced in this PR, but it's a bit odd to see functions with name `_` outside of a closure function. Even though the name isn't really relevant, as we're relying on type-based dispatching, I'd expect more than a 1-character name.

https://google.github.io/styleguide/pyguide.html#3161-names-to-avoid

tswast (approved):

bigframes/core/compile/aggregate_compiler.py
TrevorBergeron: The code examples for @singledispatch in https://docs.python.org/3/library/functools.html use the `_` naming for overrides. Also the linter wants unique name for each implementation if it is not `_`. I can change though if this goes against our style guidelines.

bigframes/core/compile/aggregate_compiler.py
TrevorBergeron: CL title is a lie, I haven't actually defined the 0-arg aggregation ops yet. They still pretend to be 1-arg aggregations. Will convert 0-arg ops (rank is another one) in a follow-up pr.

PR #361, opened 2024-01-30
feat: add DataFrame.eval, DataFrame.query

TrevorBergeron:


bigframes/eval.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2024 Google LLC
```
tswast, line 1: Nit: This module name could get confusing if someone tries to call `bigframes.eval` instead of `bigframes.pandas.eval`.

I'm not sure how best to handle this, but maybe there's a sub-package we can use for this module like `core`?

bigframes/eval.py
```diff
@@ -0,0 +1,56 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import dataclasses
+from typing import Optional
+
+import pandas
+import pandas.core.computation.parsing as pandas_eval_parsing
```
tswast, line 19: This is a private method. We'll need to be very careful with this if we want to go with this implementation.

Please instead pull this function into `third_party` so we can avoid surprises.

bigframes/eval.py
```diff
@@ -0,0 +1,56 @@
...
+
+class EvalSeries(series.Series):
+    """Slight modified series that works better with pandas.eval"""
+
+    def __init__(self, underlying: series.Series):
+        super().__init__(data=underlying._block)
+
+    @property
+    def values(self):
+        """Returns fake numpy array with only dtype property so that eval can determine schema without actually downloading the data."""
```
tswast, line 55: Neat!

tswast (changes requested):

bigframes/eval.py
TrevorBergeron: yeah, sounds good, moved to core.

bigframes/eval.py
TrevorBergeron: Done. Was even considering vendoring the entire eval implementation. That way, I could get rid of the series.values.dtype calls and also be more resilient to any pandas changes.

bigframes/core/eval.py
```diff
@@ -0,0 +1,57 @@
...
+
+import pandas
+
+import bigframes.dataframe as dataframe
+import bigframes.dtypes
+import bigframes.series as series
+import third_party.bigframes_vendored.pandas.core.computation.parsing as vendored_pandas_eval_parsing
+
+
+def eval(df: dataframe.DataFrame, expr: str, target: Optional[dataframe.DataFrame]):
```
tswast, line 26: Docstring? I'm especially interested in what the `target` argument does.

bigframes/eval.py
tswast: > Was even considering vendoring the entire eval implementation.

Let's do that. Even though `eval` is public, it does spook me to pass in anything but pandas objects into it.

bigframes/core/eval.py
TrevorBergeron: added

bigframes/eval.py
TrevorBergeron: Have now vendored the pandas `eval` implementation

third_party/bigframes_vendored/pandas/core/computation/engines.py
```diff
@@ -0,0 +1,121 @@
...
+    ----------
+    expr : Expr
+        Terms can contain
+    """
+    names = expr.names
+    overlap = names & _ne_builtins
+
+    if overlap:
+        s = ", ".join([repr(x) for x in overlap])
+        raise NumExprClobberingError(
```
tswast, line 38: I imagine we don't actually use the NumExpr package (https://github.com/pydata/numexpr) right? Maybe we can delete this function?

third_party/bigframes_vendored/pandas/core/computation/eval.py
```diff
@@ -0,0 +1,401 @@
...
+      * If an invalid engine is passed.
+    ImportError
+      * If numexpr was requested but doesn't exist.
+
+    Returns
+    -------
+    str
+        Engine name.
+    """
+    from pandas.core.computation.check import NUMEXPR_INSTALLED
```
tswast, line 46: Ditto, I doubt we actually want to use numexpr, right?

third_party/bigframes_vendored/pandas/core/computation/eval.py
```diff
@@ -0,0 +1,401 @@
...
+            level + 1,
+            global_dict=global_dict,
+            local_dict=local_dict,
+            resolvers=resolvers,
+            target=target,
+        )
+
+        parsed_expr = Expr(expr, engine=engine, parser=parser, env=env)
+
+        if engine == "numexpr" and (
```
tswast, line 322: I'm wondering if we want to catch this and raise immediately?

Aside: I could imagine adding other "engines" in future. My initial thought for this issue was for a "sql" engine to materialize to a table and/or allow injecting arbitrary SQL, but that gets tricky since our labels don't always match the column IDs.

third_party/bigframes_vendored/pandas/core/computation/engines.py
tswast: I wonder how much the "python" engine has changed over the years? If it's been pretty stable, then I think there's probably more benefit in cutting the vendored code to just the "python" engine code. If it's pretty actively changing, then keeping this file as close to as in sync with upstream makes more sense.

third_party/bigframes_vendored/pandas/core/computation/ops.py
```diff
@@ -0,0 +1,605 @@
...
+    "arctan",
+    "arccosh",
+    "arcsinh",
+    "arctanh",
+    "abs",
+    "log10",
+    "floor",
+    "ceil",
+)
+_binary_math_ops = ("arctan2",)
```
tswast, line 51: Aside: are there any of these unary math ops that we're missing in our series implementation?

third_party/bigframes_vendored/pandas/core/computation/ops.py
```diff
@@ -0,0 +1,605 @@
...
+    """
+    Compute the vectorized membership of ``x in y`` if possible, otherwise
+    use Python.
+    """
+    try:
+        return x.isin(y)
+    except AttributeError:
+        if is_list_like(x):
+            try:
+                return y.isin(x)
```
tswast, line 256: It's a little strange that they flipped the order here. Like if "x" is not a series, then shouldn't it be trying to see if "y" is a Series of lists and x is a value in that list?

third_party/bigframes_vendored/pandas/core/computation/ops.py
```diff
@@ -0,0 +1,605 @@
...
+        return issubclass(t, (datetime, np.datetime64))
+
+
+def _in(x, y):
+    """
+    Compute the vectorized membership of ``x in y`` if possible, otherwise
+    use Python.
+    """
+    try:
+        return x.isin(y)
```
tswast, line 252: Aside: if y is a Series of an ARRAY column, would this do an element-wise "is in" check?

third_party/bigframes_vendored/pandas/core/computation/ops.py
```diff
@@ -0,0 +1,605 @@
...
+    def __call__(self, env) -> MathCall:
+        operand = self.operand(env)
+        # error: Cannot call function of unknown type
+        return self.func(operand)  # type: ignore[operator]
+
+    def __repr__(self) -> str:
+        return pprint_thing(f"{self.op}({self.operand})")
+
+    @property
+    def return_type(self) -> np.dtype:
```
tswast, line 571: Funny to see this pattern here too. We're gonna have so many expression trees, lol.

third_party/bigframes_vendored/pandas/core/computation/parsing.py
```diff
@@ -0,0 +1,196 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Hashable, Iterator
+
+# A token value Python's tokenizer probably will never use.
+BACKTICK_QUOTED_STRING = 100
+
+
+def create_valid_python_identifier(name: str) -> str:
```
tswast, line 20: Aside: I wonder if we should use this function when creating column names? Could make them more interpretable.

third_party/bigframes_vendored/pandas/core/frame.py
```diff
@@ -4904,6 +4904,50 @@ def value_counts(
         """
         raise NotImplementedError(constants.ABSTRACT_METHOD_ERROR_MESSAGE)
 
+    def eval(self, expr: str) -> DataFrame:
+        """
+        Evaluate a string describing operations on DataFrame columns.
+
+        Operates on columns only, not specific rows or elements.  This allows
+        `eval` to run arbitrary code, which can make you vulnerable to code
+        injection if you pass user input to this function.
```
tswast, line 4913: Examples?

third_party/bigframes_vendored/pandas/core/frame.py
```diff
@@ -4904,6 +4904,50 @@ def value_counts(
...
+                The expression string to evaluate.
+
+        Returns:
+            DataFrame
+        """
+        raise NotImplementedError(constants.ABSTRACT_METHOD_ERROR_MESSAGE)
+
+    def query(self, expr: str) -> DataFrame | None:
+        """
+        Query the columns of a DataFrame with a boolean expression.
```
tswast, line 4926: Examples?

tswast (commented):
Awesome! A few questions to resolve before we merge.

bigframes/dataframe.py
```diff
@@ -1489,6 +1489,17 @@ def sort_values(
             )
         return DataFrame(self._block.order_by(ordering))
 
+    def eval(self, expr: str) -> DataFrame:
+        import bigframes.core.eval as bf_eval
+
+        return bf_eval.eval(self, expr, target=self)
```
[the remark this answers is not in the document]
TrevorBergeron: eval uses `.copy()` on the target if `inplace=False`

tests/system/small/test_dataframe.py
```diff
@@ -3792,6 +3792,44 @@ def test_df_to_orc(scalars_df_index, scalars_pandas_df_index):
...
+    ("expr",),
+    [
+        ("int64_col > int64_too",),
+        ("bool_col",),
+        ("((int64_col - int64_too) % @local_var) == 0",),
+    ],
+)
+def test_df_query(scalars_dfs, expr):
+    # local_var is referenced in expressions
+    local_var = 3  # NOQA
```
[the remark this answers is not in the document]
TrevorBergeron: Yeah, its kind of scary, you tell it how many stack frames to look up and it will bring all those variables in scope for the evaluation.

third_party/bigframes_vendored/pandas/core/computation/parsing.py
TrevorBergeron: Could borrow this approach with sql special characters

third_party/bigframes_vendored/pandas/core/computation/ops.py
TrevorBergeron: We are missing `expm1`, `log1p`, `floor`, `ceil` and `arctan2`. Will see what I can do about adding those asap.

third_party/bigframes_vendored/pandas/core/computation/eval.py
TrevorBergeron: Went ahead and remove remaining references to "numexpr". Attempts to use this engine will result in exception.

third_party/bigframes_vendored/pandas/core/computation/eval.py
TrevorBergeron: removed

third_party/bigframes_vendored/pandas/core/computation/engines.py
TrevorBergeron: Seems the python engine itself is pretty stable. Went ahead and removed remaining NumExpr stuff

third_party/bigframes_vendored/pandas/core/computation/ops.py
TrevorBergeron: I think its assuming first that both x and y are series, and then assuming y is a scalar and x is a series, and finally giving up and assuming that both are scalars. I don't think a series of lists is in consideration here.

third_party/bigframes_vendored/pandas/core/computation/ops.py
TrevorBergeron: I don't think so

third_party/bigframes_vendored/pandas/core/frame.py
TrevorBergeron: added

third_party/bigframes_vendored/pandas/core/frame.py
TrevorBergeron: added

tswast (approved):

PR #365, opened 2024-02-01
fix: correct the numeric literal dtype

chelsea-lin:
- [ ] Make sure to open an issue as a internal bug

Fixes internal bug: b/323070618 🦕


bigframes/dtypes.py
```diff
@@ -378,11 +378,13 @@ def literal_to_ibis_scalar(
         scalar_expr = ibis.literal(literal, ibis_dtypes.float64)
     elif scalar_expr.type().is_integer():
         scalar_expr = ibis.literal(literal, ibis_dtypes.int64)
+    elif scalar_expr.type().is_decimal():
```
tswast, line 381: This would break BIGNUMERIC, I think. We should probably have something that goes to the nearest type. https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types

Like `is_decimal` and precision < 76, then this otherwise set to the precision and scale for BIGNUMERIC.

chelsea-lin: good point! thanks!

TrevorBergeron (approved):

PR #370, opened 2024-02-06
feat: add ml.llm.GeminiTextGenerator model

GarrettWu:

### DIFF
--- a/bigframes/ml/llm.py
+++ b/bigframes/ml/llm.py
@@ -41,6 +41,8 @@ _EMBEDDING_GENERATOR_ENDPOINTS = (
     _EMBEDDING_GENERATOR_GECKO_MULTILINGUAL_ENDPOINT,
 )
 
+_GEMINI_PRO_ENDPOINT = "gemini-pro"
+
 _ML_GENERATE_TEXT_STATUS = "ml_generate_text_status"
 _ML_EMBED_TEXT_STATUS = "ml_embed_text_status"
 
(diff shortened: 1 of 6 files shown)


ashleyxuu (commented):
Please add the GeminiTextGenerator model in the README.rst as well : https://github.com/googleapis/python-bigquery-dataframes/blob/main/README.rst?plain=1#L190

bigframes/ml/llm.py
```diff
@@ -396,3 +398,175 @@ def to_gbq(
...
+        col_label = cast(blocks.Label, X.columns[0])
+        X = X.rename(columns={col_label: "prompt"})
+
+        options = {
+            "temperature": temperature,
+            "max_output_tokens": max_output_tokens,
+            "top_k": top_k,
+            "top_p": top_p,
+            "flatten_json_output": True,
+        }
```
ashleyxuu, line 547: I believe we don't give Python users the option to specify this `flatten_json_output` as we return a DataFrame format? I guess it is only applicable in SQL interface?

docs/templates/toc.yml
```diff
@@ -107,6 +107,8 @@
         uid: bigframes.ml.llm.PaLM2TextGenerator
       - name: PaLM2TextEmbeddingGenerator
         uid: bigframes.ml.llm.PaLM2TextEmbeddingGenerator
+      - name: GeminiTextGenerator
```
ashleyxuu, line 110: Should be sorted it in alphabetic order.

bigframes/ml/llm.py
GarrettWu: right. It isn't user-facing, but what we send to SQL. It is always True value.

docs/templates/toc.yml
GarrettWu: Done

GarrettWu:
> Please add the GeminiTextGenerator model in the README.rst as well : https://github.com/googleapis/python-bigquery-dataframes/blob/main/README.rst?plain=1#L190

Done

ashleyxuu (approved):

PR #381, opened 2024-02-14
refactor!: move model optional args to kwargs

GarrettWu:
To be more like sklearn, and make API more accurate. Those param shouldn't be called through positions.

### DIFF
--- a/bigframes/ml/imported.py
+++ b/bigframes/ml/imported.py
@@ -32,15 +32,17 @@ class TensorFlowModel(base.Predictor):
     """Imported TensorFlow model.
 
     Args:
+        model_path (str):
+            GCS path that holds the model files.
         session (BigQuery Session):
             BQ session to create the model
-        model_path (str):
-            GCS path that holds the model files."""
+    """
 
     def __init__(
         self,
+        model_path: str,
+        *,
         session: Optional[bigframes.Session] = None,
-        model_path: Optional[str] = None,
     ):
         self.session = session or bpd.get_global_session()
         self.model_path = model_path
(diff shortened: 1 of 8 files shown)


bigframes/ml/imported.py
```diff
@@ -32,15 +32,17 @@ class TensorFlowModel(base.Predictor):
...
     Args:
+        model_path (str):
+            GCS path that holds the model files.
         session (BigQuery Session):
             BQ session to create the model
-        model_path (str):
-            GCS path that holds the model files."""
+    """
 
     def __init__(
```
shobsi, line 41: Should we mark this breaking change? https://www.conventionalcommits.org/en/v1.0.0/#:~:text=in%20Semantic%20Versioning).-,BREAKING%20CHANGE,-%3A%20a%20commit

bigframes/ml/imported.py
```diff
@@ -134,7 +138,7 @@ def _create_bqml_model(self):
     def _from_bq(cls, session: bigframes.Session, model: bigquery.Model) -> ONNXModel:
         assert model.model_type == "ONNX"
 
-        onnx_model = cls(session=session, model_path=None)
+        onnx_model = cls(session=session, model_path="")
```
shobsi, line 141: Curious what difference this makes ("" over None)

bigframes/ml/imported.py
GarrettWu: It is marked as breaking, see the "!" in PR title.

bigframes/ml/imported.py
GarrettWu: Nothing, if loaded from an existing model, we don't have a way to retrieve the model file path, as the model has already been loaded. Both None and "" are placeholders, change the value only to satisfy mypy, that we changed the parameter to be non-optional.

shobsi (approved):

PR #387, opened 2024-02-22
fix: exceptions raised in `apply` from a `remote_function` now surface in the client

shobsi:
BEGIN_COMMIT_OVERRIDE
fix: code exception in `remote_function` now prevents retry and surfaces in the client (#387)
END_COMMIT_OVERRIDE

Fixes internal issue 309699263 🦕


tswast (approved):
LGTM.

I've filed b/327662690 as a follow-up for an automatic call to `_cached()` inside `apply()` so that these errors happen as soon as possible. Since a remote function might do something non-deterministic, we need to do that to keep our results deterministic anyway.

PR #391, opened 2024-02-23
chore: exclude pandas 2.2.x from testing

tswast:

Towards internal issue 322845272 🦕


chelsea-lin (approved):
LTGM

noxfile.py
```diff
@@ -551,7 +551,7 @@ def prerelease(session: nox.sessions.Session, tests_path):
         # https://github.com/googleapis/python-bigquery-dataframes/issues/341
         # https://github.com/googleapis/python-bigquery-dataframes/issues/337
         # are resolved
-        "pandas!=2.1.4, !=2.2.0rc0, !=2.2.0",
+        "pandas<2.1.4",
```
shobsi, line 554: JFYI we have been following `!=` approach to track the exact breaking changes with each release via the associated follow up items (in the comments above) based on the following feedback.

https://github.com/googleapis/python-bigquery-dataframes/pull/268#discussion_r1423205172

shobsi (approved):
If we have narrowed down a new breaking issue in 2.2.1, worth adding in the comments like other excluded versions

tswast:
> If we have narrowed down a new breaking issue in 2.2.1, worth adding in the comments like other excluded versions

I don't think it's anything new compared to 2.2.0 as far as I can tell.

PR #398, opened 2024-02-27
feat: add `Series.plot.hist()` method

chelsea-lin:
Fixes internal issue b/322178004🦕

### DIFF
--- a/bigframes/series.py
+++ b/bigframes/series.py
@@ -50,6 +50,7 @@ import bigframes.operations as ops
 import bigframes.operations.aggregations as agg_ops
 import bigframes.operations.base
 import bigframes.operations.datetimes as dt
+import bigframes.operations.plot as plot
 import bigframes.operations.strings as strings
 import bigframes.operations.structs as structs
 import third_party.bigframes_vendored.pandas.core.series as vendored_pandas_series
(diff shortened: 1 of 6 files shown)


bigframes/series.py
```diff
@@ -1524,6 +1525,49 @@ def sample(
             self._block._split(ns=ns, fracs=fracs, random_state=random_state)[0]
         )
 
+    def hist(
```
tswast, line 1528: Can we expose this as `Series.plot.hist`, instead? https://pandas.pydata.org/pandas-docs/version/2.1/reference/api/pandas.Series.plot.hist.html

I'm not 100% sure why pandas provides Series.plot.hist and Series.hist https://pandas.pydata.org/pandas-docs/version/2.1/reference/api/pandas.Series.hist.html but  I think Series.plot.hist is the newer method and will make our docs organization easier.

chelsea-lin: That was what I confused too! Thanks for your suggestions!

chelsea-lin: Updated it into `Series.plot.hist`. Please review it again!

orrbradford (commented):
LGTM! I'll to @tswast to approve

chelsea-lin:
We don't need this change anymore with the change: https://github.com/googleapis/python-bigquery-dataframes/pull/420

PR #401, opened 2024-02-28
feat: add configuration option to read_gbq

### LINKED ISSUE #384: Support for query parameterization
I was wondering if there were plans to expose the ability to pass bound parameters (either named or qmark style) when constructing constructing a bigframes DataFrame e.g. via `Session.read_gbq_query`? My understanding is they would need to be thread through to the [`queryParameters` field](https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfigurationQuery.FIELDS.query_parameters) on the `QueryJobConfig` that gets constructed.

I realize it's probably a bit of a niche use-case since this is a code-based API but for context we were looking to integrate support for bigframes in a product that allows users to build Python notebook-based interactive apps which may be used by external/untrusted users (not the ones writing the app/code) and so our SQL functionality relies on using parameterization whenever these kinds of inputs are involved.

Genesis929:

Fixes https://github.com/googleapis/python-bigquery-dataframes/issues/384 🦕


tests/system/small/test_session.py
```diff
@@ -20,6 +20,8 @@
 import typing
 from typing import List
 
+import google
+from google.api_core.exceptions import InternalServerError
```
shobsi, line 24: import module not class go/pystyle#imports

tests/system/small/test_session.py
```diff
@@ -353,6 +355,45 @@ def test_read_gbq_table_wildcard_with_filter(session: bigframes.Session):
...
+            "query": {
+                "useQueryCache": True,
+                "maximumBytesBilled": "1000000000",
+                "timeoutMs": 10000,
+            }
+        },
+        pytest.param(
+            {"query": {"useQueryCache": True, "timeoutMs": 50}},
+            marks=pytest.mark.xfail(
+                raises=google.api_core.exceptions.BadRequest,
```
shobsi, line 371: curious why timeout 50ms is a bad request?

tests/system/small/test_session.py
```diff
@@ -353,6 +355,45 @@ def test_read_gbq_table_wildcard_with_filter(session: bigframes.Session):
...
+            ),
+        ),
+        pytest.param(
+            {"query": {"useQueryCache": False, "maximumBytesBilled": "100"}},
+            marks=pytest.mark.xfail(
+                raises=InternalServerError,
```
shobsi, line 377: this too, why internal server error? would be nice to add a comment for the reason

tests/system/small/test_session.py
Genesis929: It's an error that the time set is too short, this is a quick way to prove that the parameter works.

bigframes/session/__init__.py
```diff
@@ -376,7 +386,7 @@ def _query_to_destination(
         query: str,
         index_cols: List[str],
         api_name: str,
-        use_cache: bool = True,
+        configuration: dict = {"query": {"useQueryCache": True}},
```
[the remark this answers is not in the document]
Genesis929: Done

tests/system/small/test_session.py
Genesis929: Done

tests/system/small/test_session.py
Genesis929: Updated reason.

shobsi: Thanks!

tests/system/small/test_session.py
shobsi: Makes sense, thanks!

shobsi (commented):
Please wait for all checks to pass before merging

shobsi (approved):

PR #403, opened 2024-02-29
feat: Support CMEK for BQ tables

shobsi:

Fixes internal issue #327083723 🦕


bigframes/session/clients.py
```diff
@@ -121,12 +123,27 @@ def bqclient(self):
             bq_info = google.api_core.client_info.ClientInfo(
                 user_agent=self._application_name
             )
+            default_query_job_config = None
+            default_load_job_config = None
+            if self._bq_kms_key_name:
+                default_query_job_config = bigquery.QueryJobConfig(
```
shobsi, line 129: Looks like this breaks the BQML flow, in which the kms_key_name needs to be set differently:

```
Cannot set kms key name in jobs with ML DDL statements. For CREATE TABLE statement, use OPTIONS(kms_key_name=...) instead
```

https://pantheon.corp.google.com/bigquery?project=bigframes-dev&j=bq:US:8e13b27c-b938-4d44-95ba-d0e2fcadf218&page=queryresults&ws=!1m5!1m4!1m3!1sbigframes-dev!2s8e13b27c-b938-4d44-95ba-d0e2fcadf218!3sUS

shobsi: Separated the bqml client with the main bq client.

bigframes/session/__init__.py
```diff
@@ -1496,23 +1513,24 @@ def read_gbq_function(
             session=self,
         )
 
-    def _start_query(
-        self,
+    @staticmethod
```
GarrettWu, line 1516: staticmethod should be avoided. https://g3doc.corp.google.com/eng/doc/devguide/py/style/index.md?cl=head

bigframes/session/__init__.py
```diff
@@ -1668,11 +1712,10 @@ def _start_generic_job(self, job: formatting_helpers.GenericJob):
         else:
             job.result()
 
+    @staticmethod
```
GarrettWu, line 1715: same here

bigframes/session/clients.py
```diff
@@ -98,39 +99,65 @@ def __init__(
...
-                client_options=bq_options,
-                credentials=self._credentials,
-                project=self._project,
-                location=self._location,
-            )
 
         return self._bqclient
 
+    @property
+    def bqmlclient(self):
```
GarrettWu, line 155: Why we need separate clients for BQ and BQML?

GarrettWu: Oh saw your comment

GarrettWu: Well still, it doesn't need two separate clients. Instead we can do sth different in BQML with the same bqclient. The simplest way, _start_query() and _start_query_bqml() doesn't need to target at 2 different clients. (Also IMO it is better to have only one method with parameters.) Further more, we may extract the needed bqclient methods in a BqClientManager, and for BQML, there can be a derived class from BqClientManager which overrides start_query method.

It will help us keep the clients in sync.

bigframes/session/__init__.py
shobsi: Ack. refactored in the recent patch, PTAL.

bigframes/session/__init__.py
shobsi: taken care, PTAL.

bigframes/session/clients.py
shobsi: FYI it doesn't help if we also set it in DDL OPTIONS, the API simply doesn't like kms_key_name being set in the job_config for DDL statements https://pantheon.corp.google.com/bigquery?project=bigframes-dev&j=bq:US:9dc79963-ff71-4d66-bdf4-ab336c789e80

bigframes/session/clients.py
shobsi: As discussed with you offline, sent a patch to:
1. Reverted the bqmlclient separate from bqclient
2. Moved setting the encryption key from default job config to call time job config 

This comes up with the trade-off that now developer and reviewers need to make sure the code paths are setting the cmek (if any) correctly. In this change we are adding common methods to help with that:
1. _prepare_query_job_config
2. _prepare_load_job_config
3. _prepare_job_job_config

bigframes/session/__init__.py
```diff
@@ -1518,36 +1531,83 @@ def read_gbq_function(
...
+            self.bqclient, sql, job_config, max_results
         )
 
-        try:
-            query_job = self.bqclient.query(sql, job_config=job_config)
-        except google.api_core.exceptions.Forbidden as ex:
-            if "Drive credentials" in ex.message:
-                ex.message += "\nCheck https://cloud.google.com/bigquery/docs/query-drive-data#Google_Drive_permissions."
-            raise
+    def _start_query_bqml(
```
GarrettWu, line 1595: What determines between using options or jobConfig isn't BQ or BQML, but the query is a CREATE or not. Suggest to rename to _start_query_create or sth. Other wise it's misleading.

GarrettWu (approved):

bigframes/session/clients.py
GarrettWu: Thank you for verifying.

bigframes/session/__init__.py
shobsi: renamed, thanks!

bigframes/session/clients.py
```diff
@@ -98,36 +99,44 @@ def __init__(
         self._location = location
         self._use_regional_endpoints = use_regional_endpoints
         self._credentials = credentials
+        self._bq_kms_key_name = bq_kms_key_name
 
         # cloud clients initialized for lazy load
         self._bqclient = None
+        self._bqmlclient = None
```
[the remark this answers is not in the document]
shobsi: good catch, thanks, removed.

PR #409, opened 2024-03-05
feat: copy pandas matplotlib code to third_party

chelsea-lin:
This change copies the pandas `MPLPlot` and `LinePlot` classes to the `third_party` folder, in preparation for implements to the `plot.hist` features  (internal bug: b/322178004)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,6 +58,7 @@ dependencies = [
     "tabulate >= 0.9",
     "ipywidgets >=7.7.1",
     "humanize >= 4.6.0",
+    "matplotlib >= 3.7.1",
 ]
 extras = {
     # Optional test dependencies packages. If they're missed, may skip some tests.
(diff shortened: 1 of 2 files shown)


shobsi:
> This change copies the pandas `MPLPlot` and `LinePlot` classes to the `third_party` folder, in preparation for implements to the `plot.hist` features (internal bug: b/322178004)

Wondering why we can't depend on them directly through package dependency?

chelsea-lin:
> > This change copies the pandas `MPLPlot` and `LinePlot` classes to the `third_party` folder, in preparation for implements to the `plot.hist` features (internal bug: b/322178004)
> 
> Wondering why we can't depend on them directly through package dependency?

It can avoid tight dependency on pandas internal classes to ensure long-term compatibility. Pandas only tries to guarantees stability only for its public API, but not for these internal classes.

chelsea-lin:
We don't need this PR anymore with the change: https://github.com/googleapis/python-bigquery-dataframes/pull/420

PR #412, opened 2024-03-05
fix: read_pandas inline respects location

GarrettWu:

Fixes b/327544164 🦕


tests/system/small/test_session.py
```diff
@@ -369,6 +369,17 @@ def test_read_pandas(session, scalars_dfs):
     pd.testing.assert_frame_equal(result, expected)
 
 
+def test_read_pandas_inline_respects_location():
+    options = bigframes.BigQueryOptions(location="europe-west1")
+    session = bigframes.Session(options)
+
+    df = session.read_pandas(pd.DataFrame([[1, 2, 3], [4, 5, 6]]))
```
shobsi, line 376: Could we add a test for inline data too, and verify that it creates the result tables in the intended location? `bpd.Dataframe([[1, 2, 3], [4, 5, 6]])`

GarrettWu: added

shobsi (approved):

bigframes/core/nodes.py
```diff
@@ -155,6 +155,7 @@ def __hash__(self):
 @dataclass(frozen=True)
 class ReadLocalNode(BigFrameNode):
     feather_bytes: bytes
+    session: bigframes.session.Session
```
TrevorBergeron, line 159: Local data is session-independent, we don't want to add a session constrain to the node. Don't worry about a dataframe/block not having session, that just means you can execute it anywhere, as all the data sources are local.

TrevorBergeron (changes requested):

bigframes/core/nodes.py
GarrettWu: The data itself is independent of session, yes. But when reading a local data, a specific session will be used. And when executing the query, we'd call that particular session. Do we have other options than keep a record here?

TrevorBergeron: A session will be used to execute the tree, yes, but the choice of session need not be constrained by the tree itself. You can check a tree to see if it has a required session, and otherwise, just use the default session to execute trees that don't depend on a specific session.

TrevorBergeron: Discussed offline, lets just make the session an optional field that we set when users have a specific session they used to read the local data with `session.read_gbq`.

GarrettWu: Updated.

TrevorBergeron (approved):

PR #413, opened 2024-03-05
feat: add engine parameter to `read_parquet`

tswast:
This makes the default behavior in `read_parquet()` consistent with the other `read_XYZ()` methods. Pandas is used to parse the file and ordering is preserved by default. Use `engine="bigquery"` for the previous behavior (not order preserving).

BEGIN_COMMIT_OVERRIDE
feat!: add engine parameter to `read_parquet` (#413)

BREAKING-CHANGE: `read_parquet` uses a "pandas" engine to parse files by default. Use `engine="bigquery"` for the previous behavior
Release-As: 0.24.0
END_COMMIT_OVERRIDE


chelsea-lin (approved):
LGTM

PR #417, opened 2024-03-06
docs: fix the note rendering for DataFrames methods: nlargest, nsmallest

ashleyxuu:

Fixes internal issue #328445384 🦕

### DIFF
--- a/third_party/bigframes_vendored/pandas/core/frame.py
+++ b/third_party/bigframes_vendored/pandas/core/frame.py
@@ -3935,6 +3935,11 @@ class DataFrame(NDFrame):
         ``df.sort_values(columns, ascending=False).head(n)``, but more
         performant.
 
+        .. note::
+            This function cannot be used with all column types. For example, when
+            specifying columns with `object` or `category` dtypes, ``TypeError`` is
+            raised.
+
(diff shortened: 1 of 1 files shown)


third_party/bigframes_vendored/pandas/core/frame.py
```diff
@@ -3935,6 +3935,11 @@ def nlargest(self, n: int, columns, keep: str = "first"):
         ``df.sort_values(columns, ascending=False).head(n)``, but more
         performant.
```
shobsi, line 3936: Looks like nothing gets rendered except the first line in the docstring. Does all that get fixed by this reordering? https://screenshot.googleplex.com/5fTe6kSNpdMqxXc

ashleyxuu: I believe you're referring to the new method overview page.

I'm fixing this one: https://screenshot.googleplex.com/4VPB3N8qWLWbmjX

shobsi: Ah, you're right, thanks!

shobsi (approved):

PR #423, opened 2024-03-08
test: do GCF cleanup in both presubmit and e2e tests

shobsi:

Fixes internal issue 328515697 🦕


tests/system/conftest.py
```diff
@@ -34,7 +35,20 @@
 import test_utils.prefixer
 
 import bigframes
-from tests.system.utils import convert_pandas_dtypes
+from tests.system.utils import (
+    convert_pandas_dtypes,
+    delete_cloud_function,
```
tswast, line 40: Same here. Import the "utils" package/module, not the individual functions.

tests/system/conftest.py
```diff
@@ -21,6 +21,7 @@
 import typing
 from typing import Dict, Optional
 
+from google.api_core.exceptions import NotFound, ResourceExhausted
```
tswast, line 24: Nit: Use import statements for packages and modules only, not for individual types, classes, or functions.

https://google.github.io/styleguide/pyguide.html#22-imports

The only exception is with the typing package.

tests/system/conftest.py
```diff
@@ -1040,3 +1054,53 @@ def floats_bf(session, floats_pd):
...
+def cleanup_cloud_functions(session, cloudfunctions_client, dataset_id_permanent):
+    """Clean up stale cloud functions."""
+    permanent_endpoints = get_remote_function_endpoints(
+        session.bqclient, dataset_id_permanent
+    )
+    delete_count = 0
+    for cloud_function in get_cloud_functions(
+        cloudfunctions_client,
+        session.bqclient.project,
+        session.bqclient.location,
```
tswast, line 1069: Don't we create functions in more than one location?

tests/system/utils.py
```diff
@@ -15,13 +15,19 @@
...
 import geopandas as gpd  # type: ignore
+import google.api_core.operation
+from google.cloud import bigquery, functions_v2
+from google.cloud.functions_v2.types import functions
 import numpy as np
 import pandas as pd
 import pyarrow as pa  # type: ignore
 import pytest
 
+from bigframes.functions.remote_function import get_remote_function_locations
```
tswast, line 29: Same here. Import packages/modules, not functions.

tests/system/conftest.py
shobsi: Nope, one cloud function for one BQ location. For multi-region we map US -> us-central1 and EU -> europe-west1 https://github.com/googleapis/python-bigquery-dataframes/blob/adadb0658c35142fed228abbd9baa42f9372f44b/bigframes/functions/remote_function.py#L72

tests/system/conftest.py
shobsi: Done, thanks.

tests/system/conftest.py
shobsi: Done, thanks.

tswast (approved):

PR #424, opened 2024-03-08
fix: move `third_party.bigframes_vendored` to `bigframes_vendored`

tswast:
This will avoid potential conflicts when run from a location that contains the commonly named `third_party` directory.

Note: I followed the instructions here: https://stackoverflow.com/a/17179022/101923 but I also had to add a `pyproject.toml` file to support editable installations (see: https://github.com/pypa/setuptools/issues/230#issuecomment-1473278299).

Fixes internal issue 328781348.
🦕


GarrettWu (dismissed):
Just curious. If we want to avoid third_party, why can't we just remove it from the path?

tswast:
> Just curious. If we want to avoid third_party, why can't we just remove it from the path?

We need it to signify that this code has a different copyright (isn't covered under Google CLA) per go/thirdparty/non-google3 policy.

tswast:
Note: please don't manually merge unless this has been synced to main. I'm using `automerge: exact` which should ensure this change only merges if synced to main when merged via automerge.

GarrettWu (approved):

PR #430, opened 2024-03-12
feat: support CMEK for `remote_function` cloud functions

shobsi:

Fixes internal issue #328547338 🦕


bigframes/functions/remote_function.py
```diff
@@ -699,6 +711,20 @@ def remote_function(
...
             necessary IAM permissions configured as described in
             https://cloud.google.com/functions/docs/reference/iam/roles#additional-configuration.
+        cloud_function_kms_key_name (str, Optional):
+            Customer managed encryption key to protect cloud functions and
+            related data at rest. This is of the format
+            projects/PROJECT_ID/locations/LOCATION/keyRings/KEYRING/cryptoKeys/KEY.
+            Read https://cloud.google.com/functions/docs/securing/cmek for
+            more details including granting necessary service accounts
+            access to the key.
+        cloud_function_docker_repository (str, Optional):
```
GarrettWu, line 721: It looks like the repository only uses for CMEK. I may wonder why normal CF doesn't have a place to store artifacts, or it doesn't have those?

GarrettWu (approved):

bigframes/functions/remote_function.py
shobsi: normal CF does use a place to store artifacts, but it is a different, un-encrypted repository. Ideally cloud function should have created a separate encrypted repo since we are providing the cmek key anyway, but for some reason it wants the user to control that and provide the path as documented in https://cloud.google.com/functions/docs/securing/cmek#before_you_begin.

PR #441, opened 2024-03-14
fix: warn when `read_gbq` / `read_gbq_table` uses the snapshot time cache

tswast:
I've verified that this warning appears from the command line as well as in notebooks.

warning: reading cached table

- [N/A] Appropriate docs were updated (if necessary)

Fixes internal issue 329545805 where `read_gbq` gives an old version of the table unexpectedly.
🦕

### DIFF
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -629,6 +629,24 @@ class Session(
         job_config.labels["bigframes-api"] = api_name
         if use_cache and table_ref in self._df_snapshot.keys():
             snapshot_timestamp = self._df_snapshot[table_ref]
+
+            # Cache hit could be unexpected. See internal issue 329545805.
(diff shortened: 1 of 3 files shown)


tswast:
For context, this is a follow-up to https://github.com/googleapis/python-bigquery-dataframes/pull/229 which addressed internal issue 312493159 where we would often get

```
BadRequest: 400 Conflicting 'FOR SYSTEM_TIME AS OF' expressions found for table ... If a 'FOR SYSTEM_TIME AS OF' expression is used, all references of a table should use the same TIMESTAMP value.
```

errors when folks called `read_gbq` on the same table more than once and combined the resulting dataframe with previous versions of the dataframe. This can possibly happen because they wanted to rerun the cells in the notebook.

milkshakeiii (approved):

PR #444, opened 2024-03-14
fix: remove unsupported to_json arguments

chelsea-lin:
BREAKING CHANGE: This change is removing unused `orient` and `lines` arguments from `DataFrame.to_json` method. Only `orient='records'` and `lines=True` is supported so far, and the default arguments is not supported. For better user experience, this change is removing these parameters for now. 

Fixes internal issue b/329531924 🦕


milkshakeiii (approved):

chelsea-lin:
As discuss offline, we would like to keep the same behavior as pandas does. Throwing a clear error message and documenting the supported parameters look okay so far.

PR #446, opened 2024-03-15
docs: add the docs for loc and iloc indexers

ashleyxuu:

### DIFF
--- a/third_party/bigframes_vendored/pandas/core/frame.py
+++ b/third_party/bigframes_vendored/pandas/core/frame.py
@@ -5114,7 +5114,7 @@ class DataFrame(NDFrame):
         """Purely integer-location based indexing for selection by position.
 
         Returns:
-            ILocDataFrameIndexer: Purely integer-location Indexers.
+            bigframes.core.indexers.ILocDataFrameIndexer: Purely integer-location Indexers.
         """
         raise NotImplementedError(constants.ABSTRACT_METHOD_ERROR_MESSAGE)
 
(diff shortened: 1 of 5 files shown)


tswast (changes requested):
Please also update the iloc, loc docstrings in https://github.com/googleapis/python-bigquery-dataframes/blob/91bd39e8b194ddad09d53fca96201eee58063bb9/third_party/bigframes_vendored/pandas/core/frame.py#L5113 and https://github.com/googleapis/python-bigquery-dataframes/blob/91bd39e8b194ddad09d53fca96201eee58063bb9/third_party/bigframes_vendored/pandas/core/series.py#L3315 to point to the corresponding indexer class in the `Returns:` section.

third_party/bigframes_vendored/pandas/core/frame.py
```diff
@@ -5111,7 +5111,20 @@ def replace(
 
     @property
     def iloc(self):
-        """Purely integer-location based indexing for selection by position."""
+        """Purely integer-location based indexing for selection by position.
+
+        Returns:
+            ILocDataFrameIndexer: Purely integer-location Indexers.
```
tswast, line 5117: I think this will need to be the fully qualified name so that it renders as a link. Same for the others. See internal issue 296390934.

```suggestion
            bigframes.core.indexers.ILocDataFrameIndexer:
                Purely integer-location Indexers.
```

tswast (changes requested):

tswast (approved):
Thanks!

PR #453, opened 2024-03-15
feat: support Series.dt.strftime

junyazhang:
Fixes internal bug: b/309817430 🦕

Staged doc: https://screenshot.googleplex.com/aYkRZJjm2ehx8E6

### DIFF
--- a/bigframes/operations/__init__.py
+++ b/bigframes/operations/__init__.py
@@ -204,6 +204,7 @@ minute_op = create_unary_op(name="minute", type_rule=op_typing.INTEGER)
 month_op = create_unary_op(name="month", type_rule=op_typing.INTEGER)
 quarter_op = create_unary_op(name="quarter", type_rule=op_typing.INTEGER)
 second_op = create_unary_op(name="second", type_rule=op_typing.INTEGER)
+strftime_op = create_unary_op(name="strftime", type_rule=op_typing.STRING)
 time_op = create_unary_op(name="time", type_rule=op_typing.INTEGER)
 year_op = create_unary_op(name="year", type_rule=op_typing.INTEGER)
 ## Trigonometry Ops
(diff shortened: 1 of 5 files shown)


bigframes/operations/__init__.py
```diff
@@ -204,6 +204,7 @@ def create_ternary_op(
 month_op = create_unary_op(name="month", type_rule=op_typing.INTEGER)
 quarter_op = create_unary_op(name="quarter", type_rule=op_typing.INTEGER)
 second_op = create_unary_op(name="second", type_rule=op_typing.INTEGER)
+strftime_op = create_unary_op(name="strftime", type_rule=op_typing.STRING)
```
TrevorBergeron, line 207: Sorry, this is a bit confusing, but `create_unary_op` creates a class. So this and the class definition below are redundant. Since we need the `date_format` parameter, we need the full class declaration, so lets delete this line.

bigframes/operations/__init__.py
```diff
@@ -415,6 +416,15 @@ def output_type(self, *input_types):
         return input_types[0]
 
 
+@dataclasses.dataclass(frozen=True)
+class StrftimeOp(UnaryOp):
+    name: typing.ClassVar[str] = "strftime"
+    date_format: str
+
+    def output_type(self, *input_types):
+        return input_types[0]
```
TrevorBergeron, line 425: output type will always be string. best to use `bigframes.dtypes.STRING_DTYPE` in case we switch our default string dtype

bigframes/operations/__init__.py
junyazhang: Oh, that makes sense! Thanks for explaining this :-)

bigframes/operations/__init__.py
junyazhang: Sounds good.

bigframes/operations/datetimes.py
```diff
@@ -17,17 +17,19 @@
 import datetime as dt
 from typing import Optional
 
-import bigframes_vendored.pandas.core.indexes.accessor as vendordt
-
 from bigframes.core import log_adapter
 import bigframes.operations as ops
 import bigframes.operations.base
 import bigframes.series as series
+import third_party.bigframes_vendored.pandas.core.arrays.datetimelike as vendored_pandas_datetimelike
```
chelsea-lin, line 24: should be `bigframes_vendored` after this change:
https://github.com/googleapis/python-bigquery-dataframes/commit/763edeb4f4e8bc4b8bb05a992dae80c49c245e25

TrevorBergeron (approved):
lgtm

chelsea-lin (approved):
LGTM overall.

bigframes/operations/datetimes.py
junyazhang: Good catch!

PR #458, opened 2024-03-18
feat: update plot sample to 1000 rows

tswast:
In making a line plot sample with Salem, I noticed that 100 rows loses some important shape information. Most screens are > 1000 pixels wide, so this seems a reasonable default.


bigframes/operations/_matplotlib/core.py
```diff
@@ -45,7 +45,7 @@ def generate(self) -> None:
 
     def _compute_plot_data(self, data):
         # TODO: Cache the sampling data in the PlotAccessor.
-        sampling_n = self.kwargs.pop("sampling_n", 100)
+        sampling_n = self.kwargs.pop("sampling_n", 1000)
```
tswast, line 48: Maybe 500 or something would be better? 640x480 was a very common resolution in the 1990s.

chelsea-lin: I investigated how sample size affects the shape of a dataset. My findings (see document: https://docs.google.com/document/d/1KaIF7zX-7seXsb-rohl56jYjhjFdLc-HlNOYfTM1Zfw/edit?tab=t.0) indicate that:

- Samples of size 500 (sampling_n=500) broadly reflect the same shape as a sample of 1000.
- A sample size of 1000 yields a closer approximation to the true underlying distribution.

Maybe we can proceed with a sample size of 1000? Visualize the denser data in the graph for a more informative representation?

chelsea-lin (approved):

PR #460, opened 2024-03-18
feat: option to use bq connection without check

shobsi:
BQ connection is currently used in BQ remote function and BQML Vertex AI integrations. It has a dependency on resource manager API, which is still going through the process of AXT compliance. We should provide a path in which BigFrames can work without resource manager interaction for the following two reasons:

1. AXT compliance: By having such path user can invoke a fully AXT compliant GCP interaction
1. This lowers the IAM bar for the user to use bigframes remote function and llm (which currently require getIamPolcy and setIamPolcy permissions) by managing the BQ connection outside of bigframes.

Fixes internal issue 330224555 🦕


bigframes/ml/llm.py
```diff
@@ -289,12 +295,16 @@ def __init__(
...
+            None
+            if self.session._skip_bq_connection_check
+            else clients.BqConnectionManager(
+                self.session.bqconnectionclient, self.session.resourcemanagerclient
+            )
         )
 
         connection_name = connection_name or self.session._bq_connection
-        self.connection_name = self._bq_connection_manager.resolve_full_connection_name(
+        self.connection_name = clients.BqConnectionManager.resolve_full_connection_name(
```
GarrettWu, line 307: classmethod should be limited. Just a module method is enough.

go/pystyle#decorators-decision

bigframes/_config/bigquery_options.py
```diff
@@ -33,19 +33,22 @@ class BigQueryOptions:
 
     def __init__(
         self,
+        *,
```
GarrettWu, line 36: Any chance we can separate the breaking change with others? Then the commit history will be more user-friendly.

bigframes/functions/remote_function.py
```diff
@@ -152,12 +149,13 @@ def create_bq_remote_function(
     ):
         """Create a BigQuery remote function given the artifacts of a user defined
         function and the http endpoint of a corresponding cloud function."""
-        self._bq_connection_manager.create_bq_connection(
-            self._gcp_project_id,
-            self._bq_location,
-            self._bq_connection_id,
-            "run.invoker",
-        )
+        if self._bq_connection_manager:
```
GarrettWu, line 152: I would think of creating a dummy ConnectionManager class, and when skip_bq_connection_check is True, use the dummy implementation instead of a real one. Then the same logic can apply to all the places, and the user (of the connection_manager) doesn't need to think about the implementations.

Well just a suggestion, up to you.

shobsi: Um, gave it a thought, feels weird to implement a class with a method create_bq_connection which doesn't create anything. I'd prefer to not do that. Let me know if you have strong opinion about it.

bigframes/ml/llm.py
shobsi: it was an existing classmethod, moved outside to the module level.

bigframes/_config/bigquery_options.py
shobsi: Reverted the signature change.

bigframes/functions/remote_function.py
GarrettWu: In that case we'd rename the create_bq_connection to sth like prepare_bq_connection. 

No need to put too much thoughts on it. Just a suggestion.

GarrettWu (approved):

PR #465, opened 2024-03-19
fix!: exclude remote models for .register()

GarrettWu:
BEGIN_COMMIT_OVERRIDE
fix!: exclude remote models for .register() (#465)

Release-As: 0.26.0
END_COMMIT_OVERRIDE

### DIFF
--- a/bigframes/ml/base.py
+++ b/bigframes/ml/base.py
@@ -90,6 +90,7 @@ class BaseEstimator(bigframes_vendored.sklearn.base.BaseEstimator, abc.ABC):
         return prettyprinter.pformat(self)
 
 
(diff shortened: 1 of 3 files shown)


ashleyxuu (approved):
Any reason we want to exclude remote models `.register()` for LLM class?

GarrettWu:
> Any reason we want to exclude remote models `.register()` for LLM class?

Those remote models can be registered, but can't be deployed in Vertex. There was a discrepancy in BQML that CREATE MODEL can't register on remote model types, but ALTER MODEL can. In theory, no reason to register an already deployed model endpoint.

PR #466, opened 2024-03-19
test: exclude gcsfs 2024.3.1 to unblock PRs

shobsi:

Fixes internal issue 330235782 🦕

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -611,6 +611,19 @@ def prerelease(session: nox.sessions.Session, tests_path):
     session.install("--no-deps", "pandas-gbq")
     already_installed.add("pandas-gbq")
 
+    # gcsfs has introduced breaking change in version 2024.3.1
+    # https://github.com/fsspec/gcsfs/pull/614
+    # Let's exclude that version in favor of the prerelease tests. We also need
+    # to install necessary dependencies with appropriate versions.
+    # TODO(shobs): Remove the exclusion after the issue is fixed:
+    # https://github.com/fsspec/gcsfs/issues/616
+    session.install("aiohttp!=4.0.0a0, !=4.0.0a1")
+    already_installed.add("aiohttp")
+    session.install("--no-deps", "fsspec==2024.2.0")
+    already_installed.add("fsspec")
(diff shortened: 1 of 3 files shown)


ashleyxuu (approved):

noxfile.py
```diff
@@ -611,6 +611,19 @@ def prerelease(session: nox.sessions.Session, tests_path):
     session.install("--no-deps", "pandas-gbq")
     already_installed.add("pandas-gbq")
 
+    # gcsfs has introduced breaking change in version 2024.3.1
+    # https://github.com/fsspec/gcsfs/pull/614
+    # Let's exclude that version in favor of the prerelease tests. We also need
+    # to install necessary dependencies with appropriate versions.
+    # TODO(shobs): Remove the exclusion after the issue is fixed:
+    # https://github.com/fsspec/gcsfs/issues/616
+    session.install("aiohttp!=4.0.0a0, !=4.0.0a1")
```
shobsi, line 620: I didn't want to go down the route of figuring out further dependencies and install them each with --no-deps. Is this acceptable @tswast ?

chelsea-lin: IMO, I don't think we should ask users to exclude this version of `gcsfs`. Here are the reasons:
- The issue is under discussions between "by design" or "updating default values". 
- The issue does not affect BigFrames product codes, while the affected tests are able to find a work around solution ([PR](https://github.com/googleapis/python-bigquery-dataframes/pull/468))

shobsi:
Since bigframes.pandas.read_* has dependency on pandas.read_*, the break of the latter in gcsfs 2024.3.1 [for wildcard gcs paths ](https://github.com/fsspec/gcsfs/issues/616) is theoretically a break of the former as well. But I tested offline that bigframes.pandas.read_* was never working with wildcard paths for other reasons. So we don't need to worry about the backward compatibility of bigframes itself with gcsfs version 2024.3.1. @chelsea-lin has already [worked around](https://github.com/googleapis/python-bigquery-dataframes/pull/468) the test failures by adjusting the wildcard paths in pandas.read_*.

PR #468, opened 2024-03-19
fix: renable to_csv and to_json related tests

chelsea-lin:


tests/system/small/test_dataframe_io.py
```diff
@@ -150,7 +149,7 @@ def test_to_csv_index(
     # read_csv will decode into bytes inproperly, convert_pandas_dtypes will encode properly from string
     dtype.pop("bytes_col")
     gcs_df = pd.read_csv(
-        path,
+        path.replace("*", "000000000000"),
```
shobsi, line 152: might make sense to move `FIRST_FILE = "000000000000"` from test_session.py to utils.py and reuse everywhere

chelsea-lin: Good idea. Addressed.

tests/system/small/test_dataframe_io.py
```diff
@@ -19,7 +19,11 @@
 import pyarrow as pa
 import pytest
 
-from tests.system.utils import assert_pandas_df_equal, convert_pandas_dtypes
+from tests.system.utils import (
```
shobsi, line 22: go/pystyle#imports Use import statements for packages and modules only, not for individual types, classes, or functions.

tests/system/small/test_dataframe_io.py
```diff
@@ -150,7 +153,7 @@ def test_to_csv_index(
     # read_csv will decode into bytes inproperly, convert_pandas_dtypes will encode properly from string
     dtype.pop("bytes_col")
     gcs_df = pd.read_csv(
-        path,
+        path.replace("*", FIRST_GCS_FILE_SUFFIX),
```
shobsi, line 156: [nit] would be neat to abstract it out in a function in utils:

```python
def get_first_file_from_wildcard(path):
    return path.replace("*", FIRST_GCS_FILE_SUFFIX)
```

tests/system/small/test_encryption.py
```diff
@@ -19,6 +19,7 @@
 
 import bigframes
 import bigframes.ml.linear_model
+from tests.system.utils import FIRST_GCS_FILE_SUFFIX
```
shobsi, line 22: go/pystyle#imports

tests/system/small/test_session.py
```diff
@@ -30,9 +30,7 @@
 import bigframes.dataframe
 import bigframes.dtypes
 import bigframes.ml.linear_model
-from tests.system.utils import skip_legacy_pandas
-
-FIRST_FILE = "000000000000"
+from tests.system.utils import FIRST_GCS_FILE_SUFFIX, skip_legacy_pandas
```
shobsi, line 33: go/pystyle#imports

tests/system/small/test_dataframe_io.py
chelsea-lin: Trying to fix most files in this PR but leave `series.py` as it is, to avoid expanding this PR with many unrelated changes.

tests/system/small/test_dataframe_io.py
chelsea-lin: Done.

tests/system/small/test_session.py
chelsea-lin: Done!

shobsi (approved):

PR #483, opened 2024-03-20
feat: support Series.dt.normalize

milkshakeiii:

### DIFF
--- a/tests/system/small/operations/test_datetimes.py
+++ b/tests/system/small/operations/test_datetimes.py
@@ -275,11 +275,10 @@ def test_dt_strftime_time():
 @skip_legacy_pandas
 def test_dt_normalize(scalars_dfs, col_name):
     scalars_df, scalars_pandas_df = scalars_dfs
-    bf_series: bigframes.series.Series = scalars_df[col_name]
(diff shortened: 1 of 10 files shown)


tests/system/small/operations/test_datetimes.py
```diff
@@ -266,3 +266,20 @@ def test_dt_strftime_time():
...
+
+@pytest.mark.parametrize(
+    ("col_name",),
+    DATETIME_COL_NAMES,
+)
+@skip_legacy_pandas
+def test_dt_normalize(scalars_dfs, col_name):
+    scalars_df, scalars_pandas_df = scalars_dfs
+    bf_series: bigframes.series.Series = scalars_df[col_name]
+    bf_result = bf_series.dt.normalize().to_pandas()
```
chelsea-lin, line 279: we don't define `bf_series`, but just call `scalars_df[col_name].dt.normalize().to_pandas()`. Would that work also?

chelsea-lin (approved):
lgtm

tests/system/small/operations/test_datetimes.py
milkshakeiii: True, bf_series is unnecessary. Fixed, thanks!

PR #490, opened 2024-03-21
refactor: convert ordering definitions to use flexible scalar express…

TrevorBergeron:
…ions

### DIFF
--- a/bigframes/core/compile/compiled.py
+++ b/bigframes/core/compile/compiled.py
@@ -52,6 +52,7 @@ PREDICATE_COLUMN = "bigframes_predicate"
 T = typing.TypeVar("T", bound="BaseIbisIR")
 
 op_compiler = op_compilers.scalar_op_compiler
+ibis_backend = ibis_bigquery.Backend()
 
 
 class BaseIbisIR(abc.ABC):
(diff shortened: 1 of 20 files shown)


bigframes/core/compile/compiled.py
```diff
@@ -50,6 +52,7 @@
 T = typing.TypeVar("T", bound="BaseIbisIR")
 
 op_compiler = op_compilers.scalar_op_compiler
+ibis_backend = ibis_bigquery.Backend()
```
tswast, line 55: I'm a bit curious what this is for. IIRC this might cause some authentication to happen, which would break unit tests, but maybe that's been fixed in Ibis?

bigframes/core/compile/compiled.py
```diff
@@ -214,7 +221,7 @@ def peek_sql(self, n: int):
         # Peek currently implemented as top level LIMIT op.
         # Execution engine handles limit pushdown.
         # In future, may push down limit/filters in compilation.
-        sql = ibis_bigquery.Backend().compile(self._to_ibis_expr().limit(n))
+        sql = ibis_backend.compile(self._to_ibis_expr().limit(n))
```
tswast, line 224: I don't think compilation will need credentials, but if it does this is potentially out of sync with the credentials used in the Session.

tswast (approved):

bigframes/core/compile/compiled.py
TrevorBergeron: Reverting this aspect of the PR, though didn't notice any issues with authentication.

PR #494, opened 2024-03-21
fix: plot.scatter `c` argument functionalities

chelsea-lin:
Fixes internal bug: b/330770901 🦕

### DIFF
--- a/bigframes/operations/_matplotlib/core.py
+++ b/bigframes/operations/_matplotlib/core.py
@@ -14,10 +14,15 @@
 
 import abc
 import typing
+import uuid
+
+import matplotlib.pyplot as plt
+import pandas as pd
+
+import bigframes.dtypes as dtypes
 
 DEFAULT_SAMPLING_N = 1000
 DEFAULT_SAMPLING_STATE = 0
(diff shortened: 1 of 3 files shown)


bigframes/operations/_matplotlib/core.py
```diff
@@ -78,3 +87,55 @@ class ScatterPlot(SamplingPlot):
     @property
     def _kind(self) -> typing.Literal["scatter"]:
         return "scatter"
+
+    def __init__(self, data, **kwargs) -> None:
+        super().__init__(data, **kwargs)
+
+        c = self.kwargs.get("c", None)
+        if self._is_sequence_arg(c) and len(c) != self.data.shape[0]:
```
TrevorBergeron, line 95: Can we handle alignment at an earlier level and just accept `c` being a column id present in the data?

chelsea-lin: As we discussed offline, we can ask users to add this list-like colors to the dataframe before plotting and pass the column name.

TrevorBergeron (approved):

PR #511, opened 2024-03-25
fix: Fix case where df.peek would fail to execute even with force=True

TrevorBergeron:

### DIFF
--- a/bigframes/core/tree_properties.py
+++ b/bigframes/core/tree_properties.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import functools
 
 import bigframes.core.nodes as nodes
 
@@ -31,7 +30,6 @@ def local_only(node: nodes.BigFrameNode) -> bool:
     return all(isinstance(node, nodes.ReadLocalNode) for node in node.roots)
 
 
-@functools.cache
 def peekable(node: nodes.BigFrameNode) -> bool:
     if local_only(node):
         return True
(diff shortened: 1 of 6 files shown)


bigframes/core/tree_properties.py
```diff
@@ -25,3 +29,12 @@ def is_trivially_executable(node: nodes.BigFrameNode) -> bool:
 
 def local_only(node: nodes.BigFrameNode) -> bool:
     return all(isinstance(node, nodes.ReadLocalNode) for node in node.roots)
+
+
+@functools.cache
```
tswast, line 34: Would this get unwieldy with local data? I worry about that Feather data getting stored forever.

TrevorBergeron: Yeah, not sure why I put that cache decorator. Would take a really long time for the cache to get too big though. Removed and will figure out the whole property caching thing later.

tswast (commented):
I'm worried about cached behavior and peelable getting out of sync again. Any chance we could refactor to avoid the possibility of the null value after we call cached?

TrevorBergeron:
> I'm worried about cached behavior and peelable getting out of sync again. Any chance we could refactor to avoid the possibility of the null value after we call cached?

Added 

> I'm worried about cached behavior and peelable getting out of sync again. Any chance we could refactor to avoid the possibility of the null value after we call cached?

Now forcing the materialization after caching.

tswast (approved):

PR #513, opened 2024-03-25
fix: assign NaN scalar to column error.

Genesis929:


tests/system/small/test_dataframe.py
```diff
@@ -605,17 +606,24 @@ def test_assign_new_column_w_loc(scalars_dfs):
     pd.testing.assert_frame_equal(bf_result, pd_result)
 
 
-def test_assign_new_column_w_setitem(scalars_dfs):
+@pytest.mark.parametrize(
+    ("scalar",),
+    [
+        (2.1,),
+        (np.nan,),
```
GarrettWu, line 613: Does it need to be np.nan? Or None is just fine, then you don't need to import numpy.

GarrettWu (approved):

tests/system/small/test_dataframe.py
Genesis929: Updated.

tswast (approved):

PR #516, opened 2024-03-26
perf: Automatically condense internal expression representation

TrevorBergeron:
… schema system.


bigframes/core/__init__.py
```diff
@@ -183,7 +183,7 @@ def project_to_id(self, expression: ex.Expression, output_id: str):
                 child=self.node,
                 assignments=tuple(exprs),
             )
-        )
+        ).rewrite_projection()
```
chelsea-lin, line 186: how to pick up these functions and ask for rewriting the projection? Can we add the `rewrite_projection` to the ArrayValue constructor? In case other developers who will add new function here and miss the `rewrite_projection`  while missing context.

bigframes/core/rewrite.py
```diff
@@ -126,8 +137,10 @@ def maybe_join(
         new_columns = remap_names(join_def, lselection, rselection)
 
         # Reconstruct ordering
+        reverse_root = self.reverse_root
         if join_type == "right":
             new_ordering = right.ordering
+            reverse_root = right.reverse_root
         elif join_type == "outer":
```
chelsea-lin, line 144: do we need to redefine the `reverse_root` for the outer join case?

bigframes/core/compile/compiled.py
```diff
@@ -1062,18 +1062,24 @@ def _bake_ordering(self) -> OrderedIR:
...
+        new_ordering = ExpressionOrdering(
+            tuple(new_exprs),
+            self._ordering.integer_encoding,
+            self._ordering.string_encoding,
+            self._ordering.total_ordering_columns,
+        )
         return OrderedIR(
             self._table,
             columns=self.columns,
-            hidden_ordering_columns=[*self._hidden_ordering_columns, *new_baked_cols],
```
chelsea-lin, line 1075: Is it intension to omit the existing `_hidden_ordering_columns` here?

chelsea-lin (changes requested):

bigframes/core/rewrite.py
TrevorBergeron: outer join we treat as left order taking precedence, so it inherits reverse_root from the self.

bigframes/core/compile/compiled.py
TrevorBergeron: yes, we are rewriting the hidden columns here, so we are recreating the set of hidden columns and discarding the old set.

bigframes/core/__init__.py
TrevorBergeron: moved to constructor in new revision.

TrevorBergeron: on second thought, this isn't really possible, as ArrayValue is an immutable dataclass, and mutation (by rewriting) is blocked even in the initializer

chelsea-lin: I see. Thanks for trying that out! Maybe adding some comments for why we can add the `rewrite_projection` here? So other developers can have more context when they add something similar?

chelsea-lin (approved):
LGTM

PR #529, opened 2024-03-27
deps: include `pyarrow` as a dependency

tswast:
Test with a minimum version of pyarrow to confirm compatibility.

🦕

### DIFF
--- a/testing/constraints-3.9.txt
+++ b/testing/constraints-3.9.txt
@@ -12,7 +12,7 @@ google-cloud-resource-manager==1.10.3
 google-cloud-storage==2.0.0
 ibis-framework==8.0.0
 pandas==1.5.0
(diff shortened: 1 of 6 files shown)


milkshakeiii (approved):

testing/constraints-3.9.txt
```diff
@@ -12,6 +12,7 @@ google-cloud-resource-manager==1.10.3
 google-cloud-storage==2.0.0
 ibis-framework==8.0.0
 pandas==1.5.0
+pyarrow==3.0.0
```
milkshakeiii, line 15: can this be higher? looks like 
E   AttributeError: type object 'pyarrow.lib.RecordBatch' has no attribute 'from_pydict'
is due to 3.0.0 being too old

PR #532, opened 2024-03-27
refactor: Add type constraints to internal op definitions.

TrevorBergeron:

### DIFF
--- a/bigframes/operations/__init__.py
+++ b/bigframes/operations/__init__.py
@@ -161,102 +161,165 @@ def _convert_expr_input(
 
 
 # Operation Factories
-def create_unary_op(
-    name: str, type_rule: op_typing.OpTypeRule = op_typing.INPUT_TYPE
-) -> UnaryOp:
+def create_unary_op(name: str, type_signature: op_typing.UnaryTypeSignature) -> UnaryOp:
     return dataclasses.make_dataclass(
         name,
-        [("name", typing.ClassVar[str], name), ("output_type", typing.ClassVar[typing.Callable], type_rule.as_method)],  # type: ignore
+        [("name", typing.ClassVar[str], name), ("output_type", typing.ClassVar[typing.Callable], type_signature.as_method)],  # type: ignore
         bases=(UnaryOp,),
         frozen=True,
     )()
 
 
 def create_binary_op(
-    name: str, type_rule: op_typing.OpTypeRule = op_typing.Supertype()
+    name: str, type_signature: op_typing.BinaryTypeSignature
 ) -> BinaryOp:
     return dataclasses.make_dataclass(
         name,
-        [("name", typing.ClassVar[str], name), ("output_type", typing.ClassVar[typing.Callable], type_rule.as_method)],  # type: ignore
+        [("name", typing.ClassVar[str], name), ("output_type", typing.ClassVar[typing.Callable], type_signature.as_method)],  # type: ignore
         bases=(BinaryOp,),
-        frozen=True,
-    )()
-
-
-def create_ternary_op(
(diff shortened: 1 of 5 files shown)


bigframes/operations/type.py
```diff
@@ -22,54 +23,175 @@
 
 
 @dataclasses.dataclass
-class OpTypeRule:
-    def output_type(self, *input_types: ExpressionType) -> ExpressionType:
-        raise NotImplementedError("Abstract typing rule has no output type")
+class TypeSignature(abc.ABC):
```
chelsea-lin, line 26: Please add a docstring and give some context here?

bigframes/operations/type.py
```diff
@@ -22,54 +23,175 @@
...
+        lcd_type = bigframes.dtypes.lcd_etype(left_type, right_type)
+        if lcd_type == bigframes.dtypes.INT_DTYPE:
+            return bigframes.dtypes.FLOAT_DTYPE
+        return lcd_type
 
 
 @dataclasses.dataclass
-class Fixed(OpTypeRule):
-    out_type: ExpressionType
+class Numeric(BinaryTypeSignature):
```
chelsea-lin, line 126: `BinaryNumeric`? Why we have both this and `BinaryRealNumeric`?

bigframes/operations/type.py
```diff
@@ -22,54 +23,175 @@
...
+    def output_type(
+        self, left_type: ExpressionType, right_type: ExpressionType
+    ) -> ExpressionType:
+        if not bigframes.dtypes.is_numeric(left_type):
+            raise TypeError(f"Type {left_type} is not numeric")
+        if not bigframes.dtypes.is_numeric(right_type):
+            raise TypeError(f"Type {right_type} is not numeric")
+        lcd_type = bigframes.dtypes.lcd_etype(left_type, right_type)
+        if lcd_type == bigframes.dtypes.INT_DTYPE:
+            return bigframes.dtypes.FLOAT_DTYPE
```
chelsea-lin, line 121: please add a comment to explain why we need a special case here?

bigframes/operations/__init__.py
```diff
@@ -392,9 +458,15 @@ class StructFieldOp(UnaryOp):
     name_or_index: str | int
 
     def output_type(self, *input_types):
-        pd_type = typing.cast(pd.ArrowDtype, input_types[0])
-        pa_struct_t = typing.cast(pa.StructType, pd_type.pyarrow_dtype)
-        pa_result_type = pa_struct_t[self.name_or_index].type
+        input_type = input_types[0]
+        if not isinstance(input_type, pd.ArrowDtype):
+            raise TypeError("field accessor input must be a struct type")
```
chelsea-lin, line 463: typo: array?

chelsea-lin (changes requested):

bigframes/operations/type.py
TrevorBergeron: The difference is whether integer inputs result in integer output for a float output. Some operations, such as logarithm, will produce float result even if input is an integer eg `log(3) ~= 0.47712`. Other operations, such as abs, preserve int type eg `abs(-2) == 2`

bigframes/operations/type.py
TrevorBergeron: Added comment

bigframes/operations/__init__.py
TrevorBergeron: this is for structs

bigframes/operations/type.py
chelsea-lin: Thanks! It would be much easier to follow then!

chelsea-lin (approved):
LGTM overall. If the change grows after fixing the presubmit tests, I can take another look.

PR #534, opened 2024-03-27
chore: experimentally double load test timeout

milkshakeiii:
While I work in parallel from other angles, I'd like to try out increasing the timeouts for the kokoro runs substantially, just as another data point.


milkshakeiii:
Never mind, this doesn't appear to be the timeout that we're hitting.

milkshakeiii:
Actually, we are hitting one of these (the 360 mins one)

PR #535, opened 2024-03-27
chore: experimentally double load test timeout

milkshakeiii:
While I work in parallel from other angles, I'd like to try out doubling the timeouts for the kokoro runs, just as another data point.


milkshakeiii:
I know that the 60 * 60 * 12 argument refers to individual tests, but I'd like to ensure we're limited by the vm lifetime so that I'm not adjusting two variables at once.

tswast (approved):

PR #537, opened 2024-03-27
fix: don't download 100gb onto local python machine in load test

milkshakeiii:

### DIFF
--- a/tests/system/load/test_large_tables.py
+++ b/tests/system/load/test_large_tables.py
@@ -74,13 +74,9 @@ def test_index_repr_large_table():
     assert actual is not None
 
 
-# FAILED
-# tests/system/load/test_large_tables.py::test_to_pandas_batches_large_table
-# google.api_core.exceptions.Forbidden: 403 Response too large to return.
-# Consider specifying a destination table in your job...
-@pytest.mark.xfail
 def test_to_pandas_batches_large_table():
-    df = bpd.read_gbq("load_testing.scalars_100gb")
+    df = bpd.read_gbq("load_testing.scalars_1gb")
+    # df will be downloaded locally
     expected_row_count, expected_column_count = df.shape
 
     row_count = 0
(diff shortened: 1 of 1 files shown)


tests/system/load/test_large_tables.py
```diff
@@ -74,13 +74,9 @@ def test_index_repr_large_table():
...
 
 
-# FAILED
-# tests/system/load/test_large_tables.py::test_to_pandas_batches_large_table
-# google.api_core.exceptions.Forbidden: 403 Response too large to return.
-# Consider specifying a destination table in your job...
-@pytest.mark.xfail
 def test_to_pandas_batches_large_table():
-    df = bpd.read_gbq("load_testing.scalars_100gb")
+    df = bpd.read_gbq("load_testing.scalars_1gb")
```
shobsi, line 78: can we take a middle value, say 20gb to still represent a large table. I suspect even pandas may be able to handle 1gb.

milkshakeiii: I think a middle value would probably work, but I don't think we're really testing hardly anything here other than the download speed of the local machine? Even if it was an error in read_gbq, we wouldn't catch that. The middle layer is the BigQuery python library and then it's Bigframes. I don't quite see what load-related errors we're catching? 1 GB is enough to test the round trip... more seems like just testing downloading, what do you think?

shobsi: my understanding is that it's really the load - more the better. I agree it depends on the size of the VM running the test, but we should test as large as we can.

milkshakeiii: Hm, alright, well the available middle ground is 10GB, so I'll switch to that. We also have other tests that do 1TB and such in this file, so we should be all set with this change, I think.

shobsi (approved):

tswast (approved):

PR #543, opened 2024-03-28
docs: Mark Gemini model as Pre-GA

GarrettWu:

### DIFF
--- a/bigframes/ml/llm.py
+++ b/bigframes/ml/llm.py
@@ -421,10 +421,11 @@ class PaLM2TextEmbeddingGenerator(base.BaseEstimator):
 class GeminiTextGenerator(base.BaseEstimator):
     """Gemini text generator LLM model.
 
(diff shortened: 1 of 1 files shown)


bigframes/ml/llm.py
```diff
@@ -421,6 +421,11 @@ def to_gbq(
 class GeminiTextGenerator(base.BaseEstimator):
     """Gemini text generator LLM model.
 
+    Note: This product or feature is subject to the "Pre-GA Offerings Terms" in the General Service Terms section of the
```
ashleyxuu, line 424: We should follow a docstring format like this for note:
```
.. note::
    This product or feature is subject to the "Pre-GA Offerings Terms" in the General Service Terms section of the
```

GarrettWu: Done

ashleyxuu (approved):
Thanks.

PR #544, opened 2024-03-29
feat: Allow DataFrame binary ops to align on either axis and with loc…

TrevorBergeron:
…al objects.


tests/system/small/test_dataframe.py
```diff
@@ -2074,6 +2075,37 @@ def test_series_binop_axis_index(
...
+    ],
+    ids=[
+        "tuple",
+        "pd_index",
+        "bf_index",
+        "pd_series",
+        "bf_series",
+    ],
+)
+def test_listlike_binop_axis_1(scalars_dfs, input):
```
milkshakeiii, line 2096: Should we add axis=1 to some existing tests for various binary ops?

bigframes/core/normalize.py
```diff
@@ -0,0 +1,49 @@
...
+# limitations under the License.
+from __future__ import annotations
+
+import pandas as pd
+
+import bigframes.core.indexes.index as index
+import bigframes.series as series
+
+
+def normalize_to_bf_series(obj, default_index: index.Index) -> series.Series:
```
milkshakeiii, line 22: I would prefer just "to_bf_series" since to_blah seems like a common name for things that basically just convert, and "normalize" has so many possible meanings.

bigframes/dataframe.py
```diff
@@ -663,22 +664,24 @@ def _apply_binop(
...
-                how=how,
-                reverse=reverse,
-            )
-        elif isinstance(other, bigframes.series.Series):
-            return self._apply_series_binop(
-                other, op, axis=axis, how=how, reverse=reverse
-            )
         elif isinstance(other, DataFrame):
             return self._apply_dataframe_binop(other, op, how=how, reverse=reverse)
+        elif isinstance(other, pandas.DataFrame):
```
milkshakeiii, line 671: might need to add pandas.DataFrame to the type hint for other?

milkshakeiii (approved):

bigframes/core/normalize.py
TrevorBergeron: Done

tests/system/small/test_dataframe.py
TrevorBergeron: axis=1 completely changes alignment so the objects would just fail to align on this axis for existing tests. I'm not too worried about the different ops breaking, the actual scalar op compilation doesn't change for axis=1, the big worry is alignment failing

bigframes/dataframe.py
TrevorBergeron: For what purpose? Mypy is able to infer based on the `isinstance` checks.

milkshakeiii: Oh, gotcha, but should we remove the other type hints then? Just for consistency/cleanliness.

tests/system/small/test_dataframe.py
milkshakeiii: Oh, right, the existing objects wouldn't wouldn't work. Okay, sounds good.

PR #553, opened 2024-03-30
feat: Add a unique session_id to Session and allow cleaning up sessions

milkshakeiii:
- temporary tables will have the session id in their names
- session.close() will delete temporary tables that were created using the session.
- add pandas.get_default_session_id()
- add manual_cleanup_by_session_id(session_id: str). this is slow but allows users to clean up if they lost the session object


bigframes/core/global_session.py
```diff
@@ -26,8 +26,8 @@
 
 def close_session() -> None:
     """Start a fresh session the next time a function requires a session.
-
-    Closes the current session if it was already started.
+    Closes the current session if it was already started, deleting any
```
tswast, line 29: This will be included as part of the first summary line. We want a new line above. (https://peps.python.org/pep-0257/#multi-line-docstrings)

```suggestion

    Closes the current session if it was already started, deleting any
```

bigframes/dataframe.py
```diff
@@ -2666,9 +2666,7 @@ def to_gbq(
                 )
             if_exists = "replace"
 
-            temp_table_ref = bigframes.session._io.bigquery.random_table(
-                self._session._anonymous_dataset
-            )
+            temp_table_ref = self._session._random_table(skip_cleanup=True)
```
tswast, line 2669: Let's add a comment here explaining that the user and/or integration tool now owns this temp table we create, so that's why it's not included in cleanup.

bigframes/pandas/__init__.py
```diff
@@ -679,6 +679,43 @@ def to_datetime(
...
+    session id of the default global session. It is appended to the
+    table id of all temporary tables created in the global session.
+
+    Returns:
+        str, the default global session id, ex. 'sessiona1b2c'
+    """
+    return get_global_session().session_id
+
+
+def manual_cleanup_by_session_id(session_id: str) -> None:
```
tswast, line 694: We don't need the word `manual` here. That's implied by the fact the user is calling it.

```suggestion
def cleanup_by_session_id(session_id: str) -> None:
```

bigframes/pandas/__init__.py
```diff
@@ -679,6 +679,43 @@ def to_datetime(
...
+    """Searches through datasets and table names in Bigquery and
+    deletes tables found matching the expected format. This is a
+    slow operation which could be useful if the python session
```
tswast, line 697: Let's keep the summary line short.

```suggestion
    deletes tables found matching the expected format.
    
    This is a slow operation which could be useful if the session
```

bigframes/pandas/__init__.py
```diff
@@ -679,6 +679,43 @@ def to_datetime(
...
+    object has been lost. Calling `session.close()` or
+    `pandas.close_session()` is preferred in most cases.
```
tswast, line 699:
```suggestion
    `bigframes.pandas.close_session()` is preferred in most cases.
```

bigframes/pandas/__init__.py
```diff
@@ -679,6 +679,43 @@ def to_datetime(
 to_datetime.__doc__ = vendored_pandas_datetimes.to_datetime.__doc__
 
 
+def get_default_session_id() -> str:
+    """Gets the default session id. This is the session id that is
+    used whenever a custom session has not been provided. It is the
+    session id of the default global session. It is appended to the
```
tswast, line 685: `append` usually means to add to the end, at least in Python.

```suggestion
    session id of the default global session. It is prepended to the
```

alternatively

```suggestion
    session id of the default global session. It is prefixed to the
```

bigframes/session/__init__.py
```diff
@@ -1970,3 +2016,21 @@ def _transform_read_gbq_configuration(configuration: Optional[dict]) -> dict:
...
+    """Searches within the dataset for tables conforming to the
+    expected session_id form, and instructs bigquery to delete them.
+    """
+
+    tables = client.list_tables(dataset, page_size=1000)
+    for table in tables:
+        split_id = table.table_id.split("_")
+        if not split_id[0].startswith("bqdf"):
+            continue
+        found_session_id = split_id[1]
```
tswast, line 2033: We might want to protect against a table named just `bqdf`.

```suggestion
        if not split_id[0].startswith("bqdf") or len(split_id) < 2:
            continue
        found_session_id = split_id[1]
```

bigframes/session/__init__.py
```diff
@@ -1970,3 +2016,21 @@ def _transform_read_gbq_configuration(configuration: Optional[dict]) -> dict:
         configuration["jobTimeoutMs"] = timeout_ms
 
     return configuration
+
+
+def _delete_tables_matching_session_id(
```
tswast, line 2021: Seeing as this doesn't require any Session object stuff, could we move this to bigframes/session/_io/bigquery.py ? This file is getting a bit long for my editor to handle comfortably sometimes.

tests/system/large/test_session.py
```diff
@@ -46,7 +50,63 @@
...
+
+    session.close()
+
+    with pytest.raises(google.cloud.exceptions.NotFound):
+        bqclient.delete_table(full_id_1)
+    with pytest.raises(google.cloud.exceptions.NotFound):
+        bqclient.delete_table(full_id_2)
+
+
+# this test is prohibitively slow on our test project
```
tswast, line 87: I'm a bit uncomfortable with this skip. I wonder if we create a session in a location our tests don't really use if it'd be faster?

tests/system/small/test_pandas_options.py
```diff
@@ -288,9 +287,7 @@ def test_close_session_after_credentials_need_reauthentication(monkeypatch):
         with pytest.raises(google.auth.exceptions.RefreshError):
             bpd.read_gbq(test_query)
 
-        # Now verify that closing the session works
-        bpd.close_session()
-        assert bigframes.core.global_session._global_session is None
```
tswast, line 293: Please find a way to restore this. If folks aren't authenticated properly I'd like to provide a way to at least restore the client-side state.

Let's raise a warning with the session ID saying that we weren't able to clean up.

third_party/bigframes_vendored/pandas/core/frame.py
```diff
@@ -329,7 +329,7 @@ def to_gbq(
             >>> df = bpd.DataFrame({'col1': [1, 2], 'col2': [3, 4]})
             >>> destination = df.to_gbq(ordering_id="ordering_id")
             >>> # The table created can be read outside of the current session.
-            >>> bpd.close_session()  # For demonstration, only.
+            >>> bpd.close_session()
```
tswast, line 332: Let's keep this comment or find a way to rephrase so it's 100% clear that the user doesn't have to close the session.

```suggestion
            >>> bpd.close_session()  # Optional, to demonstrate a new session.
```

tswast (changes requested):

tswast:
Also, please update our tests to make sure we're cleaning up after ourselved:

https://github.com/googleapis/python-bigquery-dataframes/blob/4995c0046265463bc5c502cbeb34c7632d5a255e/tests/system/conftest.py#L131

and

https://github.com/googleapis/python-bigquery-dataframes/blob/4995c0046265463bc5c502cbeb34c7632d5a255e/tests/system/conftest.py#L139

should be updated as follows:

```python
session = bigframes.Session(...)
yield session
session.close()
```

bigframes/pandas/__init__.py
milkshakeiii: Ha, that's fair.

bigframes/dataframe.py
milkshakeiii: Added, thanks

bigframes/session/__init__.py
milkshakeiii: Roger, moved

tests/system/large/test_session.py
milkshakeiii: Good idea, I tried that, and in doing so I found out the cause of the slowness. And I don't like it. for list_tables (and presumably list_datasets) on bigquery.Client, I had to pass the max_results parameter in order to get the page_size parameter to work, otherwise pages were still small despite passing page_size. :(

Anyway, test works now!

tests/system/small/test_pandas_options.py
milkshakeiii: I have changed close_session to just warn instead of error out when it hits the credentials expired error. The test now checks to make sure there's a warning.

milkshakeiii:
> Also, please update our tests to make sure we're cleaning up after ourselved:
> 
> https://github.com/googleapis/python-bigquery-dataframes/blob/4995c0046265463bc5c502cbeb34c7632d5a255e/tests/system/conftest.py#L131
> 
> and
> 
> https://github.com/googleapis/python-bigquery-dataframes/blob/4995c0046265463bc5c502cbeb34c7632d5a255e/tests/system/conftest.py#L139
> 
> should be updated as follows:
> 
> ```python
> session = bigframes.Session(...)
> yield session
> session.close()
> ```

This is a nice pytest feature. Added.

bigframes/core/global_session.py
```diff
@@ -36,7 +40,15 @@ def close_session() -> None:
...
-            _global_session.close()
+            try:
+                _global_session.close()
+            except google.auth.exceptions.RefreshError as e:
+                session_id = _global_session.session_id
+                warnings.warn(
+                    "Session cleanup failed for session with"
+                    + "id: {session_id}".format(session_id=session_id)
+                )
+                print(e)
```
tswast, line 51: Nit: This should be in stderr. Also, just printing the exception can lose valuable debugging info. Let's do the whole traceback.

https://stackoverflow.com/a/16946886/101923

```suggestion
                traceback.print_tb(err.__traceback__)
```

In future, we can do `traceback.print_exception` (https://docs.python.org/3/library/traceback.html#traceback.print_exception) but if I'm reading that right not until Python 3.10 is our minimum version.

bigframes/dataframe.py
```diff
@@ -2666,9 +2666,10 @@ def to_gbq(
                 )
             if_exists = "replace"
 
-            temp_table_ref = bigframes.session._io.bigquery.random_table(
-                self._session._anonymous_dataset
-            )
+            temp_table_ref = self._session._random_table(
+                skip_cleanup=True
+            )  # the client code owns
+            # this table reference now, so skip_cleanup=True to not clean it up when we close the session
```
tswast, line 2672: I think the formatting got a little funky maybe because of `black`

```suggestion
            temp_table_ref = self._session._random_table(
                # The client code owns this table reference now, so skip_cleanup=True
                #  to not clean it up when we close the session.
                skip_cleanup=True,
            )
```

tswast (changes requested):

bigframes/pandas/__init__.py
```diff
@@ -695,6 +695,48 @@ def to_datetime(
...
+    table id of all temporary tables created in the global session.
+
+    Returns:
+        str, the default global session id, ex. 'sessiona1b2c'
+    """
+    return get_global_session().session_id
+
+
+def clean_up_by_session_id(session_id: str) -> None:
+    """Searches through datasets and table names in Bigquery and
```
tswast, line 711: I don't think we need to loop through datasets. There is only one dataset per user + project + location combination that the temp tables would be in and that's the one we discover here:

https://github.com/googleapis/python-bigquery-dataframes/blob/0e24036fe33c60cef08ba78e7ce69319338b6b03/bigframes/session/__init__.py#L279-L288

Likely we'll want to refactor that function so we can re-use it here. Maybe put a function in https://github.com/googleapis/python-bigquery-dataframes/blob/main/bigframes/session/_io/bigquery.py that takes a bqclient, location, and project and returns the anonymous dataset?

bigframes/core/global_session.py
```diff
@@ -36,7 +41,15 @@ def close_session() -> None:
...
     with _global_session_lock:
         if _global_session is not None:
-            _global_session.close()
+            try:
+                _global_session.close()
+            except google.auth.exceptions.RefreshError as e:
+                session_id = _global_session.session_id
+                warnings.warn(
+                    "Session cleanup failed for session with"
+                    + "id: {session_id}".format(session_id=session_id)
```
tswast, line 50: Since the session is scoped to the project and location, could we include these in the output, too?

bigframes/pandas/__init__.py
```diff
@@ -695,6 +695,48 @@ def to_datetime(
...
+    session id of the default global session. It is prefixed to the
+    table id of all temporary tables created in the global session.
+
+    Returns:
+        str, the default global session id, ex. 'sessiona1b2c'
+    """
+    return get_global_session().session_id
+
+
+def clean_up_by_session_id(session_id: str) -> None:
```
tswast, line 710: Let's require project and location (or maybe make them optional parameters and use the `_anonymous_dataset` on the global session if one of those isn't passed in).

tswast (changes requested):

bigframes/core/global_session.py
milkshakeiii: Added

bigframes/pandas/__init__.py
milkshakeiii: Roger, added the suggested optional behavior and checked that neither or both are passed (since we don't want to silently ignore it if someone just passes one.)

tests/system/large/test_session.py
```diff
@@ -46,7 +50,72 @@
...
+    bqclient = session.bqclient
+
+    expiration = (
+        datetime.datetime.now(datetime.timezone.utc)
+        + bigframes.constants.DEFAULT_EXPIRATION
+    )
+    full_id_1 = bigframes.session._io.bigquery.create_temp_table(session, expiration)
+    full_id_2 = bigframes.session._io.bigquery.create_temp_table(session, expiration)
+
+    with pytest.raises(google.cloud.exceptions.Conflict):
```
tswast, line 74: Let's explain what we're doing here:

```suggestion
    # Check that the table was actually created.
    with pytest.raises(google.cloud.exceptions.Conflict):
```

Thought: Wouldn't a call to `get_table` be a little easier to tell the intention?

tests/system/large/test_session.py
```diff
@@ -46,7 +50,72 @@
...
+        bqclient.create_table(full_id_1)
+    with pytest.raises(google.cloud.exceptions.Conflict):
+        bqclient.create_table(full_id_2)
+
+    session.close()
+
+    with pytest.raises(google.cloud.exceptions.NotFound):
```
tswast, line 81: Likewise, let's explain here too.

```suggestion
   # Check that session.close() deleted the tables.
    with pytest.raises(google.cloud.exceptions.NotFound):
```

tests/system/large/test_session.py
```diff
@@ -46,7 +50,72 @@
...
+    tables_before = bqclient.list_tables(dataset, max_results=10000, page_size=10000)
+    tables_before_count = len(list(tables_before))
+    assert tables_before_count >= 2
+
+    bpd.clean_up_by_session_id(
+        session_id, location=session._location, project=session._project
+    )
+
+    tables_after = bqclient.list_tables(dataset, max_results=10000, page_size=10000)
+    assert len(list(tables_after)) <= tables_before_count - 2
```
tswast, line 118: This is going to be a pretty flaky test, I think. We run tests in parallel, so there might be other tests that create some tables in the same dataset.

In fact, this test might run at the same time across Python versions if/when we split our system tests up into multiple Kokoro jobs.

Instead, let's just check:

* In `tables_before` that there are at least some tables with the `session_id` as a prefix.
* In `tables_after` that there are no tables with the `session_id` as a prefix.

Even with this change, it could get flaky when there are more than 10000 tables, but at least it'll be a bit more robust.

tests/system/large/test_session.py
```diff
@@ -46,7 +50,72 @@
...
+
+def test_clean_up_by_session_id():
+    original_location = bigframes.options.bigquery.location
+    bpd.close_session()  # close the session to change regions
+    bigframes.options.bigquery.location = "europe-west10"
+    # we do this test in a different region in order to avoid
+    # overly large amounts of temp tables slowing the test down
+
+    session = bpd.get_global_session()
+    session_id = session.session_id
```
tswast, line 95: Using the global session for this is somewhat scary since other tests running in parallel might try to do the same thing.

Can we test what we need to test without using the global session here? If not, maybe we can create a test fixture for the global session that uses a lock so that tests that use the global session don't run in parallel?

tests/system/large/test_session.py
```diff
@@ -46,7 +50,72 @@
...
+
+    bpd.clean_up_by_session_id(
+        session_id, location=session._location, project=session._project
+    )
+
+    tables_after = bqclient.list_tables(dataset, max_results=10000, page_size=10000)
+    assert len(list(tables_after)) <= tables_before_count - 2
+
+    bpd.close_session()  # close the session to change regions
+    bigframes.options.bigquery.location = original_location
```
tswast, line 121: This can leave the tests in a weird state if the test fails. Let's use the [pytest monkeypatch fixture](https://docs.pytest.org/en/latest/reference/reference.html#pytest.MonkeyPatch.setattr) to ensure things are set to the previous value when the test finishes.

Probably should do this resetting via monkeypatch in the "global session" text fixture with the lock that I suggest above, actually.

tests/system/small/test_pandas_options.py
```diff
@@ -289,7 +290,9 @@ def test_close_session_after_credentials_need_reauthentication(monkeypatch):
             bpd.read_gbq(test_query)
 
         # Now verify that closing the session works
-        bpd.close_session()
+        with warnings.catch_warnings(record=True) as warned:
```
tswast, line 293: Please create a specific warning type for this in `bigframes.exceptions` and check for that here. See `UnknownLocationWarning`: https://github.com/search?q=repo%3Agoogleapis%2Fpython-bigquery-dataframes%20UnknownLocationWarning&type=code

tswast (changes requested):

bigframes/session/_io/bigquery.py
```diff
@@ -275,3 +259,57 @@ def pytest_log_job(query_job: bigquery.QueryJob):
...
+        if not split_id[0].startswith("bqdf") or len(split_id) < 2:
+            continue
+        found_session_id = split_id[1]
+        if found_session_id == session_id:
+            client.delete_table(table, not_found_ok=True)
+            print("Deleting temporary table '{}'.".format(table.table_id))
+
+
+def create_bq_dataset_reference(
+    bq_client: bigquery.Client, **kwargs
```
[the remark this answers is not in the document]
milkshakeiii: Okay, done.

tests/system/large/test_session.py
milkshakeiii: changed to `assert bqclient.get_table(full_id_1).created is not None`

milkshakeiii: Also added a comment.

tests/system/large/test_session.py
milkshakeiii: Roger, changed to `assert any([(session.session_id in table.full_table_id) for table in list(tables_before)])`

bigframes/session/_io/bigquery.py
```diff
@@ -275,3 +259,57 @@ def pytest_log_job(query_job: bigquery.QueryJob):
...
+        dataset (bigquery.DatasetReference):
+            The dataset to search in
+        session_id (str):
+            The session id to match on in the table name
+
+    Returns:
+        None
+    """
+
+    tables = client.list_tables(dataset, max_results=10000, page_size=10000)
```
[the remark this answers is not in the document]
milkshakeiii: Done

tests/system/large/test_session.py
milkshakeiii: Assuming just using a new session object works, we can avoid the need for this entirely.

tests/system/large/test_session.py
milkshakeiii: How about just building a session object and passing in the location in the context parameter? I've made that change. Let me know if you're aware of any problem with it. Thanks!

tswast: https://github.com/googleapis/python-bigquery-dataframes/pull/652 actually is my approach to addressing this problem. `option_context` can create a thread-local session.

tswast (commented):
Thanks! A couple minor things before we merge.

tswast (approved):

bigframes/pandas/__init__.py
```diff
@@ -699,6 +699,67 @@ def to_datetime(
...
+        None
+    """
+    client = get_global_session().bqclient
+
+    if (location is None) != (project is None):
+        raise ValueError(
+            "Only one of project or location was given. Must specify both or neither."
+        )
+    elif location is None and project is None:
+        dataset = get_global_session()._anonymous_dataset
```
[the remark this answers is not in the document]
milkshakeiii: Oh, good catch. Fixed.

tests/system/conftest.py
```diff
@@ -127,16 +127,20 @@ def resourcemanager_client(
 
 
 @pytest.fixture(scope="session")
-def session() -> bigframes.Session:
-    return bigframes.Session()
+def session() -> Generator[bigframes.Session, None, None]:
+    session = bigframes.Session()
```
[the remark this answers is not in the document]
milkshakeiii: Roger, will resolve.

PR #554, opened 2024-04-01
fix: allow `repr` for results more than 10 GB

tswast:
Revert "feat: Support max_columns in repr and make repr more efficient"

Reverts googleapis/python-bigquery-dataframes#515


tswast:
Marking as `do not merge` until I can understand what's going on. I think it might be that there's no `LIMIT` in the `repr` query any more (intentionally so).

milkshakeiii (approved):

PR #556, opened 2024-04-01
feat: (Series|DataFrame).explode

chelsea-lin:

### DIFF
--- a/bigframes/core/__init__.py
+++ b/bigframes/core/__init__.py
@@ -402,17 +402,13 @@ class ArrayValue:
         return ArrayValue(join_node)
 
     def explode(self, column_ids: typing.Sequence[str]) -> ArrayValue:
-        column_ids = [
-            column_id
-            for column_id in column_ids
-            if bigframes.dtypes.is_array_like(self.get_column_type(column_id))
-        ]
-        if len(column_ids) == 0:
-            return ArrayValue(self.node)
-        else:
-            return ArrayValue(
-                nodes.ExplodeNode(child=self.node, column_ids=tuple(column_ids))
-            )
+        assert len(column_ids) > 0
+        for column_id in column_ids:
+            assert bigframes.dtypes.is_array_like(self.get_column_type(column_id))
+
+        return ArrayValue(
+            nodes.ExplodeNode(child=self.node, column_ids=tuple(column_ids))
+        )
 
     def _uniform_sampling(self, fraction: float) -> ArrayValue:
         """Sampling the table on given fraction.
--- a/bigframes/core/__init__.py
+++ b/bigframes/core/__init__.py
@@ -401,6 +401,19 @@ class ArrayValue:
             return ArrayValue(bigframes.core.rewrite.maybe_rewrite_join(join_node))
         return ArrayValue(join_node)
 
+    def explode(self, column_ids: typing.Sequence[str]) -> ArrayValue:
+        column_ids = [
+            column_id
+            for column_id in column_ids
+            if bigframes.dtypes.is_array_like(self.get_column_type(column_id))
+        ]
+        if len(column_ids) == 0:
+            return ArrayValue(self.node)
+        else:
+            return ArrayValue(
+                nodes.ExplodeNode(child=self.node, column_ids=tuple(column_ids))
+            )
+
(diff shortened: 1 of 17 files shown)


bigframes/core/compile/compiled.py
```diff
@@ -719,6 +763,85 @@ def _uniform_sampling(self, fraction: float) -> OrderedIR:
...
+            *self._hidden_ordering_column_names,
+        )
+
+        unnest_array_id = bigframes.core.guid.generate_guid("unnest_array_")
+        unnest_offset_id = bigframes.core.guid.generate_guid("unnest_offset_")
+
+        unnest_array = table_w_zip_array[zip_array_id].unnest().name(unnest_array_id)
+        unnested_columns = [
+            unnest_array[f"f{index+2}"].name(column_id)
+            for index, column_id in zip(range(len(column_ids)), column_ids)
```
TrevorBergeron, line 806: try out `enumerate` https://docs.python.org/3/library/functions.html#enumerate

bigframes/core/compile/compiled.py
```diff
@@ -719,6 +763,85 @@ def _uniform_sampling(self, fraction: float) -> OrderedIR:
...
+            *self._column_names,
+            *self._hidden_ordering_column_names,
+        )
+
+        zip_array_id = bigframes.core.guid.generate_guid("zip_array_")
+        zip_array = (
+            table_w_offset[offset_array_id]
+            .zip(*[table_w_offset[column_id] for column_id in column_ids])
+            .name(zip_array_id)
+        )
```
TrevorBergeron, line 793: Have you tried directly indexing using the offsets rather than zipping? I'm worried compiling this creates an extra correlated JOIN.

tests/system/small/test_series.py
```diff
@@ -3416,3 +3416,99 @@ def foo(x: int, y: int, df):
...
+                {"A": {"y": 2.0}, "B": "bb"},
+                {"A": {"z": 4.0}},
+                {},
+                numpy.nan,
+            ],
+            id="struct",
+        ),
+    ],
+)
+def test_series_explode(data):
```
TrevorBergeron, line 3440: I'm not sure if any of these tests use the unordered path. Maybe use `to_pandas(ordered=False)` for a test or two? Also for unordered test cases make sure not to ignore index as resetting the index will invoke the ordered path.

bigframes/core/compile/compiled.py
chelsea-lin: Thanks!

bigframes/core/__init__.py
```diff
@@ -401,6 +401,19 @@ def join(
             return ArrayValue(bigframes.core.rewrite.maybe_rewrite_join(join_node))
         return ArrayValue(join_node)
 
+    def explode(self, column_ids: typing.Sequence[str]) -> ArrayValue:
+        column_ids = [
+            column_id
+            for column_id in column_ids
+            if bigframes.dtypes.is_array_like(self.get_column_type(column_id))
+        ]
```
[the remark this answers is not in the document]
chelsea-lin: Added a comment here.

bigframes/core/blocks.py
```diff
@@ -1162,6 +1162,26 @@ def calculate_pairwise_metric(self, op=agg_ops.CorrOp()):
...
+        column_ids: typing.Sequence[str],
+        ignore_index: Optional[bool],
+    ) -> Block:
+        expr = self.expr.explode(column_ids)
+        if ignore_index:
+            return Block(
+                expr.drop_columns(self.index_columns),
+                column_labels=self.column_labels,
+                index_columns=[],
+            )
```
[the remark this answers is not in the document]
chelsea-lin: Added a comment here.

tests/system/small/test_series.py
chelsea-lin: It looks dtype validation will trigger the unordered path. Also I added the aggregation and `to_pandas(ordered=False)` as you suggested. Thanks!

bigframes/core/compile/compiled.py
chelsea-lin: Thanks! The new logical has a better performance!

TrevorBergeron (approved):
LGTM

PR #558, opened 2024-04-01
fix: Respect hard stack size limit and swallow limit change exception.

TrevorBergeron:


bigframes/pandas/__init__.py
```diff
@@ -714,9 +714,13 @@ def to_datetime(
...
 sys.setrecursionlimit(max(10000000, sys.getrecursionlimit()))
-resource.setrlimit(
-    resource.RLIMIT_STACK, (resource.RLIM_INFINITY, resource.RLIM_INFINITY)
-)
+
+soft_limit, hard_limit = resource.getrlimit(resource.RLIMIT_STACK)
+if soft_limit < hard_limit or hard_limit == resource.RLIM_INFINITY:
+    try:
+        resource.setrlimit(resource.RLIMIT_STACK, (hard_limit, hard_limit))
+    except Exception:
```
GarrettWu, line 722: Any particular reason we want to pass all the exceptions? Generally we should never suppress all exceptions. https://g3doc.corp.google.com/eng/doc/devguide/py/style/index.md?cl=head#s2.4-exceptions

TrevorBergeron: Increasing stack limits isn't a hard requirement for bigframes to work. It merely reduces the likelihood of overflowing stack memory for deeply nested trees algorithms later on. From looking at other codebases that change resource limits, it can be rather error-prone due to each platform's inconsistency/limits/bugs and its hard to guarantee that the operation will succeed, so wanted to play it safe.

GarrettWu (approved):

PR #562, opened 2024-04-02
docs: add docs for `DataFrame` and `Series` dunder methods

shobsi:
  - `__add__`: https://screenshot.googleplex.com/8ZLiQJLZPTWEZdg
  - `__radd__`: https://screenshot.googleplex.com/5Rn7Bn8guqBrBTD
  - `radd`: https://screenshot.googleplex.com/C5KVoZCNTL9oiHq

Fixes internal issue 304747371 🦕


bigframes/dataframe.py
```diff
@@ -797,11 +798,12 @@ def _apply_dataframe_binop(
     def eq(self, other: typing.Any, axis: str | int = "columns") -> DataFrame:
         return self._apply_binop(other, ops.eq_op, axis=axis)
 
+    def __eq__(self, other) -> DataFrame:
+        return self.eq(other)
+
```
ashleyxuu, line 803: how about adding `inspect.getdoc(vendored_pandas_frame.DataFrame.__eq__)` to generate the docs?

ashleyxuu: __eq__ = eq

tswast (commented):
Were any of these docstring or examples derived from the pandas documentation? If so, they need to be moved to third_party.

shobsi:
> Were any of these docstring or examples derived from the pandas documentation? If so, they need to be moved to third_party.

`DataFrame.__array__` is one case that could be debated. There is no pandas `DataFrame.__array__` documentation, so I put BigFrames `DataFrame.__array__` directly. There is pandas `Series.__array__`, so I added BigFrames `Series.__array__` in third party. For consistency I kept wording of `DataFrame.__array__` close to `Series.__array__`. Is that acceptable?

bigframes/dataframe.py
shobsi: use this pattern at bunch of places, however for `__eq__` I created its own docstring with code samples.

tswast:
> adding docstring with code samples directly in our classes which are worded similar to their public versions 

Anything derived from third-party content needs to be in the third-party directory.

ashleyxuu (approved):

bigframes/dataframe.py
shobsi: Moved all dunder docstrings to third_party for compliance reasons (even slight similarity with an existing pandas documentation may be considered as derived) and used this pattern for explicit override.

shobsi:
> > adding docstring with code samples directly in our classes which are worded similar to their public versions
> 
> Anything derived from third-party content needs to be in the third-party directory.

Done.

PR #563, opened 2024-04-02
fix: plot.scatter s parameter cannot accept float-like column

chelsea-lin:
Fixes internal b/330574847 🦕

### DIFF
--- a/bigframes/operations/_matplotlib/core.py
+++ b/bigframes/operations/_matplotlib/core.py
@@ -115,6 +115,18 @@ class ScatterPlot(SamplingPlot):
         if self._is_column_name(c, sample) and sample[c].dtype == dtypes.STRING_DTYPE:
             sample[c] = sample[c].astype("object")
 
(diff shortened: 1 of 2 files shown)


bigframes/operations/_matplotlib/core.py
```diff
@@ -115,6 +114,18 @@ def _compute_plot_data(self):
         if self._is_column_name(c, sample) and sample[c].dtype == dtypes.STRING_DTYPE:
             sample[c] = sample[c].astype("object")
 
+        # To avoid Matplotlib's automatic conversion of `Float64` or `Int64` columns
+        # to `object` types (which breaks float-like behavior), this code proactively
+        # converts the column to a compatible format."
```
milkshakeiii, line 119:
```suggestion
        # converts the column to a compatible format.
```

Stray character

milkshakeiii (approved):

bigframes/operations/_matplotlib/core.py
chelsea-lin: Good catch. Thanks!

PR #569, opened 2024-04-02
fix: reloaded transformer .transform error

GarrettWu:

### DIFF
--- a/bigframes/ml/compose.py
+++ b/bigframes/ml/compose.py
@@ -117,15 +117,15 @@ class ColumnTransformer(
 
         output_names = []
         for transform_col in bq_model._properties["transformColumns"]:
-            transform_col = cast(dict, transform_col)
+            transform_col_dict = cast(dict, transform_col)
             # pass the columns that are not transformed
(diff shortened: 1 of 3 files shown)


bigframes/ml/compose.py
```diff
@@ -115,14 +115,17 @@ def camel_to_snake(name):
...
             # pass the columns that are not transformed
-            if "transformSql" not in transform_col:
+            if "transformSql" not in transform_col_dict:
                 continue
-            transform_sql: str = cast(dict, transform_col)["transformSql"]
+            transform_sql: str = transform_col_dict["transformSql"]
             if not transform_sql.startswith("ML."):
                 continue
 
+            output_names.append(transform_col_dict["name"])
```
shobsi, line 128: Should we also check the existence of this key in `transform_col_dict`?

GarrettWu: We can. But in all cases it should exist, and nothing we can do if it doesn't. Checking it still raises an internal error, I feel it makes no difference.

shobsi (approved):

PR #570, opened 2024-04-02
fix: error for object dtype on read_pandas

chelsea-lin:
Fixes internal bug b/329861151🦕


TrevorBergeron (changes requested):
`bpd.Series(pd.Series("hello", "world")).to_string()` actually works in bigframes. Sometimes we can infer a compatible type from "object" dtype pandas objects.

TrevorBergeron (approved):

PR #580, opened 2024-04-04
feat: support list of numerics in pandas.cut

milkshakeiii:
An internal user encountered this missing overload


bigframes/core/reshape/__init__.py
```diff
@@ -125,9 +125,22 @@ def cut(
...
             bins = tuple((bin.left.item(), bin.right.item()) for bin in bins)
-        else:
+        elif len(list(bins)) == 0:
+            raise ValueError("`bins` iterable should have at least one item")
+        elif isinstance(list(bins)[0], tuple):
             as_index = pd.IntervalIndex.from_tuples(list(bins))
             bins = tuple(bins)
+        elif pd.api.types.is_number(list(bins)[0]):
+            if len(list(bins)) < 2:
+                raise ValueError(
```
shobsi, line 135: pandas doesn't seem to have this restriction. can we not have this exception?

tests/system/small/test_pandas.py
```diff
@@ -424,6 +424,31 @@ def test_cut_default_labels(scalars_dfs):
     )
 
 
+def test_cut_numeric_breaks(scalars_dfs):
+    scalars_df, scalars_pandas_df = scalars_dfs
+
+    breaks = [0, 5, 10, 15, 20, 100, 1000]
```
shobsi, line 430: can we also add a test case (parameterize) for float values?

bigframes/core/reshape/__init__.py
milkshakeiii: I looked into this quite a bit this morning, but it looks like we need more support for the struct type in order to be able to create the column of NaN structs that this case requires.

tests/system/small/test_pandas.py
milkshakeiii: Good point, added, thanks. Also fixed mixing ints and floats.

tests/system/small/test_pandas.py
```diff
@@ -424,6 +424,38 @@ def test_cut_default_labels(scalars_dfs):
...
 
+@pytest.mark.parametrize(
+    ("breaks",),
+    [
+        ([0, 5, 10, 15, 20, 100, 1000],),  # ints
+        ([0.5, 10.5, 15.5, 20.5, 100.5, 1000.5],),  # floats
+        ([0, 5, 10.5, 15.5, 20, 100, 1000.5],),  # mixed
+    ],
+)
+def test_cut_numeric_breaks(scalars_dfs, breaks):
```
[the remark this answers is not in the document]
milkshakeiii: Added, thanks.

tests/system/small/test_pandas.py
```diff
@@ -449,6 +456,30 @@ def test_cut_numeric_breaks(scalars_dfs):
     )
 
 
+def test_cut_errors(scalars_dfs):
+    scalars_df, _ = scalars_dfs
+
+    with pytest.raises(ValueError):
+        # negative integer bins argument
+        bpd.cut(scalars_df["float64_col"], -1)
```
shobsi, line 464: [readability] could be parameterized

shobsi (approved):

tests/system/small/test_pandas.py
milkshakeiii: Added

ashleyxuu (approved):

PR #595, opened 2024-04-08
test: test most relevant dtype for aggregates

shobsi:

Fixes internal issue 329522681 🦕


bigframes/core/blocks.py
```diff
@@ -899,7 +901,22 @@ def aggregate_all_and_stack(
...
+                    [
+                        aggregate_expr.get_column_type(c)
+                        for c in aggregate_expr.column_ids
+                    ]
+                )
+                if len(aggregate_dtypes) == 1:
+                    dtype = aggregate_dtypes.pop()
+            if dtype is None:
+                dtype = pd.Float64Dtype()
+            result_expr = aggregate_expr.unpivot(
```
TrevorBergeron, line 919: Can we use some variant of this existing logic: https://github.com/googleapis/python-bigquery-dataframes/blob/main/bigframes/dtypes.py#L657 to find the common supertype?

TrevorBergeron (changes requested):

bigframes/core/blocks.py
shobsi: Done, PTAL, thanks!

TrevorBergeron (approved):

PR #599, opened 2024-04-09
feat: Series.struct.dtypes

chelsea-lin:
Fixes internal b/333563722🦕


tswast (approved):
Thanks!

Using doctest as our system test makes sense to me here.

PR #602, opened 2024-04-10
fix: infer narrowest numeric type when combining numeric columns

TrevorBergeron:


bigframes/core/__init__.py
```diff
@@ -372,15 +369,62 @@ def unpivot(
...
+        )
+        join = join_def.JoinDefinition(
+            conditions=(), mappings=(*labels_mappings, *table_mappings), type="cross"
+        )
+        if how == "left":
+            joined_array = self.join(labels_array, join_def=join)
+        else:
+            joined_array = labels_array.join(self, join_def=join)
+
+        # 3. Build output columns by switching between input columns
```
tswast, line 407: It might help to include the fact that we are going from many columns and fewer rows -> fewer columns and more rows here too.

bigframes/core/compile/scalar_op_compiler.py
```diff
@@ -1346,6 +1365,25 @@ def clip_op(
         )
 
 
+@scalar_op_compiler.register_nary_op(ops.switch_op)
+def switch_op(*cases_and_outputs: ibis_types.Value) -> ibis_types.Value:
+    # ibis can handle most type coercions, but we need to force bool -> int
```
tswast, line 1370: I wonder if it's worth leaving them an issue to allow this implicit cast?

bigframes/operations/__init__.py
```diff
@@ -148,6 +145,21 @@ def order_preserving(self) -> bool:
         return False
 
 
+@dataclasses.dataclass(frozen=True)
+class NaryOp:
```
tswast, line 149: Thoughts on making this a superclass to Unary, Binary, Ternary?

bigframes/operations/__init__.py
```diff
@@ -664,6 +676,46 @@ def output_type(self, *input_types: dtypes.ExpressionType) -> dtypes.ExpressionT
 
 clip_op = ClipOp()
 
+
+class SwitchOp(NaryOp):
```
tswast, line 680: Could we name this `CaseWhenOp` to match SQL https://cloud.google.com/bigquery/docs/reference/standard-sql/conditional_expressions ?

TIL: This is a supported op name in pandas now too: https://pandas.pydata.org/docs/reference/api/pandas.Series.case_when.html (learned via https://stackoverflow.com/a/73619829/101923)

tswast (changes requested):
Thanks! A few questions though.

Also, since this makes some user-visible changes, `refactor: ` isn't the right prefix as that will be hidden. Let's do `fix: retain integer column type when aggregating over integer columns` or similar.

bigframes/operations/__init__.py
TrevorBergeron: renamed

bigframes/operations/__init__.py
TrevorBergeron: makes sense, done

bigframes/core/__init__.py
```diff
@@ -372,15 +369,62 @@ def unpivot(
         Returns:
             ArrayValue: The unpivoted ArrayValue
         """
+        # 1. construct array_value from row_labels, with offsets
```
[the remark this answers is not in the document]
TrevorBergeron: factored out 2 private functions. and yes, we have the row labels locally here. usually, these were the former column names that are being stacked.

bigframes/core/__init__.py
```diff
@@ -354,9 +354,6 @@ def unpivot(
         *,
         passthrough_columns: typing.Sequence[str] = (),
         index_col_ids: typing.Sequence[str] = ["index"],
-        dtype: typing.Union[
```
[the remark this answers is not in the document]
TrevorBergeron: This was added so we could mix compatible numerics together in a single `df.sum()` or similar aggregation. You would manually pass the float dtype here and it would cast everything to that type. However, this isn't necessary as we can find the common supertype of the inputs and coerce to that automatically.

bigframes/core/compile/scalar_op_compiler.py
TrevorBergeron: maybe? even pandas isn't really consistent on whether bools are numeric or not.

tswast (approved):

PR #606, opened 2024-04-11
test: series explode test

chelsea-lin:

### DIFF
--- a/tests/system/small/test_series.py
+++ b/tests/system/small/test_series.py
@@ -3463,9 +3463,8 @@ def test_series_pipe(
     ],
 )
 def test_series_explode(data):
-    data = [[1, 2, 3], [], numpy.nan, [3, 4]]
     s = bigframes.pandas.Series(data)
-    pd_s = pd.Series(data)
+    pd_s = s.to_pandas()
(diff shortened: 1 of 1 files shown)


tests/system/small/test_series.py
```diff
@@ -3463,9 +3463,8 @@ def foo(x: int, y: int, df):
     ],
 )
 def test_series_explode(data):
-    data = [[1, 2, 3], [], numpy.nan, [3, 4]]
     s = bigframes.pandas.Series(data)
-    pd_s = pd.Series(data)
+    pd_s = s.to_pandas()
```
GarrettWu, line 3467: Looks like our Series constructor does differently with pandas. Is this a known issue? Otherwise we shall file a bug.

chelsea-lin: The difference is the dtype here. The `to_pandas()` would convert to `pa.StructType`, which is not the default for pandas. The `pandas.explode()` would unnest the normal struct dtype but not `pa.StructType`. In BigFrames, we want customer to run `DataFrames.struct.explode()` explicitly to explode the struct type.

GarrettWu (approved):

PR #621, opened 2024-04-18
feat: Add transpose support for small homogeneously typed DataFrames.

TrevorBergeron:


third_party/bigframes_vendored/pandas/core/frame.py
```diff
@@ -93,6 +93,84 @@ def values(self) -> np.ndarray:
...
+        """
+        raise NotImplementedError(constants.ABSTRACT_METHOD_ERROR_MESSAGE)
+
+    def transpose(self) -> DataFrame:
+        """
+        Transpose index and columns.
+
+        Reflect the DataFrame over its main diagonal by writing rows as columns
+        and vice-versa. The property :attr:`.T` is an accessor to the method
+        :meth:`transpose`.
```
milkshakeiii, line 131: Can we add something here about the dtypes requirements? (Columns must be coerceable to a common type) (I know it's below but seems important enough to be up top.)

tests/system/small/test_dataframe.py
```diff
@@ -2465,6 +2465,24 @@ def test_df_describe(scalars_dfs):
     ).all()
 
 
+def test_df_transpose():
```
milkshakeiii, line 2468: If it's a cheap test, we could add a check that we raise TypeError for mixing strs and floats.

bigframes/core/blocks.py
```diff
@@ -1570,6 +1575,52 @@ def melt(
             index_columns=[index_id],
         )
 
+    def transpose(self) -> Block:
+        """Transpose the block. Will fail if dtypes incompatible or too many rows"""
```
milkshakeiii, line 1579: Suggest we change "dtypes incompatible" to "dtypes cannot be coerced all to a common type."

milkshakeiii (approved):
LGTM, just suggested we make it more clear in the docstrings exactly what the dtype restriction is.

third_party/bigframes_vendored/pandas/core/frame.py
TrevorBergeron: added

bigframes/core/blocks.py
TrevorBergeron: done

tests/system/small/test_dataframe.py
TrevorBergeron: added

PR #624, opened 2024-04-19
docs: set `index_cols` in `read_gbq` as a best practice

tswast:

Towards internal issue 335727141
🦕

### DIFF
--- a/third_party/bigframes_vendored/pandas/io/gbq.py
+++ b/third_party/bigframes_vendored/pandas/io/gbq.py
@@ -28,12 +28,14 @@ class GBQIOMixin:
         """Loads a DataFrame from BigQuery.
 
         BigQuery tables are an unordered, unindexed data source. By default,
-        the DataFrame will have an arbitrary index and ordering.
+        the DataFrame will have an arbitrary index and ordering. Generating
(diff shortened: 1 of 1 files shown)


shobsi (approved):

third_party/bigframes_vendored/pandas/io/gbq.py
```diff
@@ -28,12 +28,14 @@ def read_gbq(
...
-
-        Set the `index_col` argument to one or more columns to choose an
-        index. The resulting DataFrame is sorted by the index columns. For the
-        best performance, ensure the index columns don't contain duplicate
-        values.
+        the DataFrame will have an arbitrary index and ordering. Generating
+        the default index uses an analytic windowed operation that prevents
+        many filtering push down operations. As a best practice, set the
+        ``index_col`` argument to one or more columns, especially on large
+        tables.
```
shobsi, line 35: Should we make it clear that setting any arbitrary column as index is not necessarily helpful, and the best practice is to use a column with unique helpful? In other words, the following sentence is actually the best practice.

tswast: Added some caveats for non-unique index. Thanks!

PR #629, opened 2024-04-22
feat: Support `axis=1` in `df.apply` for scalar outputs

shobsi:
  - Note in `remote_function` doc https://screenshot.googleplex.com/4c6ZojpEYf7Wymj
  - Overloading of the arg `input_type` https://screenshot.googleplex.com/8TZGXJLw8rewd6W
  - Note and description update for `DataFrame.apply` https://screenshot.googleplex.com/9jLWmQNUYpoVrTA
  - Code sample and arg update in `DataFrame.apply` https://screenshot.googleplex.com/7nHi5ZjsGyTruTo

Fixes internal issue 320674749 🦕


bigframes/core/blocks.py
```diff
@@ -2011,6 +2012,61 @@ def _is_monotonic(
         self._stats_cache[column_name].update({op_name: result})
         return result
 
+    def _get_rows_as_json_values(self) -> Block:
+        sql, index_column_ids, _ = self.to_sql_query(include_index=True)
+
+        # all column names, types and values
+        all_column_names = index_column_ids + [col for col in self.column_labels]
+        column_names_csv = ", ".join([f'"{col}"' for col in all_column_names])
```
TrevorBergeron, line 2020: keep in mind, column names can be any hashable python object. Might need to restrict this to types we can serialize/deserialize properly. Additionally, special characters need to be escaped properly.

bigframes/core/blocks.py
```diff
@@ -2011,6 +2012,61 @@ def _is_monotonic(
...
+{textwrap.indent(sql, "    ")}
+),
+T1 AS (
+    SELECT *,
+           JSON_OBJECT(
+               "names", [{column_names_csv}],
+               "types", [{column_types_csv}],
+               "values", [{column_references_csv}],
+               "index", [{index_column_names_csv}]
+           ) AS {row_json_column_name} FROM T0
```
TrevorBergeron, line 2044: would it be possible to do this without generating SQL ourselves here?

bigframes/core/blocks.py
```diff
@@ -2011,6 +2012,61 @@ def _is_monotonic(
...
+)
+SELECT {select_columns_csv} FROM T1
+"""
+        ibis_table = self.session.ibis_client.sql(json_sql)
+        order_for_ibis_table = ordering.ExpressionOrdering(
+            ordering_value_columns=tuple(
+                [ordering.ascending_over(column_id) for column_id in index_column_ids]
```
TrevorBergeron, line 2051: This actually changes the ordering of the table, which we do not want. The table isn't necessarily ordered by index

bigframes/functions/remote_function.py
```diff
@@ -285,38 +288,98 @@ def generate_cloud_function_main_code(self, def_, dir):
...
+    col_types = row_json["types"]
+    col_values = row_json["values"]
+    index_names = row_json["index"]
+
+    value_converters = {
+        "boolean": lambda val: val == "true",
+        "Int64": int,
+        "Float64": float,
+        "string": str,
+    }
```
TrevorBergeron, line 313: What about DateTime types? Are those supported? If not, do we provide a useful error message as soon as they call apply?

bigframes/core/blocks.py
```diff
@@ -2011,6 +2012,61 @@ def _is_monotonic(
...
+            ),
+            total_ordering_columns=frozenset(index_column_ids),
+        )
+        expr = core.ArrayValue.from_ibis(
+            self.session,
+            ibis_table,
+            [ibis_table[col] for col in select_columns],
+            hidden_ordering_columns=[],
+            ordering=order_for_ibis_table,
+        )
```
TrevorBergeron, line 2061: by creating a brand new expression object, we decouple it from the original data sources, make some optimizations more difficult, most notably an inability to use the implicit joiner. Lets try to get this working as a scalar operation

bigframes/core/blocks.py
```diff
@@ -2011,6 +2012,61 @@ def _is_monotonic(
...
 
+    def _get_rows_as_json_values(self) -> Block:
+        sql, index_column_ids, _ = self.to_sql_query(include_index=True)
+
+        # all column names, types and values
+        all_column_names = index_column_ids + [col for col in self.column_labels]
+        column_names_csv = ", ".join([f'"{col}"' for col in all_column_names])
+        column_references_csv = ", ".join(
+            [f"CAST({col} AS STRING)" for col in all_column_names]
+        )
```
TrevorBergeron, line 2023: this will error in bigquery for some types, lets make our type constraints clear

tests/system/small/test_remote_function.py
```diff
@@ -685,3 +685,54 @@ def test_read_gbq_function_enforces_explicit_types(bigquery_client, dataset_id):
...
         )
+
+
+@pytest.mark.flaky(retries=2, delay=120)
+def test_df_apply_axis_1(session, scalars_dfs):
+    columns = ["bool_col", "int64_col", "int64_too", "float64_col", "string_col"]
+    scalars_df, scalars_pandas_df = scalars_dfs
+
+    def add_ints(row):
+        return row["int64_col"] + row["int64_too"]
```
TrevorBergeron, line 696: lets also have a test that uses some pandas behavior, such as getting the dtypes of the row, or getting the len of the row, or some statistical aggregate (eg skewness)

bigframes/core/blocks.py
TrevorBergeron: maybe TO_JSON_STRING is a bit more flexible?

TrevorBergeron (changes requested):

tests/system/small/test_remote_function.py
shobsi: in the large tests I am testing serialization of the row, which requires all the info to be present as it would be in pandas, could you PTAL if that suffices?

bigframes/core/blocks.py
shobsi: I can work on a follow up change to fit thing through our ArrayValue/nodes/ops stack (I need to understand the relationships better), would like to have something working out sooner if possible.

bigframes/core/blocks.py
shobsi: I'm looking into how to find the existing ordering information, do you have a tip on that? My current thought is to scan `block.expr.node.child_nodes` for a `OrderByNode`

bigframes/functions/remote_function.py
shobsi: I am not supporting datetime right now, but I believe I can support it later with some additional testing (there is UTC restriction on BigFrames support which I will test separately). The error message will pop up for unsupported dtype, added test, PTAL.

bigframes/core/blocks.py
shobsi: I made the column type handling more generic, PTAL.

bigframes/core/blocks.py
TrevorBergeron: You should consider the tree structure to be an implementation detail. I think the best example of restoring ordering information would be the .cache() implementation, which materializes an ordering and creates a new ArrayValue using it.

tests/system/small/test_remote_function.py
shobsi: I also added a (large) test for dtype and aggregates.

bigframes/core/blocks.py
shobsi: Left a TODO for doing it in another change, tracked by internal issue 338218579.

bigframes/core/blocks.py
shobsi: Ack. Used caching to honor the order, added a dedicated test `test_df_apply_axis_1_ordering`. PTAL.

bigframes/core/blocks.py
shobsi: Created internal issue 338218579 to track

bigframes/functions/remote_function.py
```diff
@@ -285,38 +289,104 @@ def generate_cloud_function_main_code(self, def_, dir):
...
+    col_names = row_json["names"]
+    col_types = row_json["types"]
+    col_values = row_json["values"]
+    index_names = row_json["index"]
+    dtype = row_json["dtype"]
+
+    # index and column names are not necessarily strings
+    # they are serialized as repr(repr(name)) at source
+    col_names = [eval(name) for name in col_names]
+    index_names = [eval(name) for name in index_names]
```
TrevorBergeron, line 313: Ok, I see now why `repr(repr(value))` is being used. eval makes me a bit uneasy, as it will execute arbitrary code. Als, repr/eval working nicely together isn't going to be guaranteed for more advanced types later on. For the primitive types we have right now, can we use `literal_eval` instead, or just use a non-eval deserialization entirely?

bigframes/functions/remote_function.py
```diff
@@ -285,38 +289,104 @@ def generate_cloud_function_main_code(self, def_, dir):
...
+        replies = []
+        for call in calls:
+"""
+        if is_row_processor:
+            code += """\
+            reply = udf(get_pd_series(call[0]))
+            if pd.isna(reply):
+                # Pandas N/A values are not json serializable, so use a python
+                # equivalent instead
+                reply = None
```
TrevorBergeron, line 373: Do we need to differentiate NaN vs None for float dtype?

bigframes/functions/remote_function.py
```diff
@@ -285,38 +289,104 @@ def generate_cloud_function_main_code(self, def_, dir):
...
+    # index and column names are not necessarily strings
+    # they are serialized as repr(repr(name)) at source
+    col_names = [eval(name) for name in col_names]
+    index_names = [eval(name) for name in index_names]
+
+    value_converters = {
+        "boolean": lambda val: val == "true",
+        "Int64": int,
+        "Float64": float,
+        "string": str,
```
TrevorBergeron, line 319: do these converters support null?

bigframes/functions/remote_function.py
shobsi: It should be taken care by BigFrames -> BQ contract, where they both are represented the same (NULL) in BQ, and in cloud function they are deserialized as python `None`.

```python
import pandas as pd
import numpy as np

df = pd.Series([1, 2.5, pd.NA, np.nan, None])
df
```
```
0       1
1     2.5
2    <NA>
3     NaN
4    None
dtype: object
```
```python
bpd.options.bigquery.project = "bigframes-dev"

bdf = bpd.read_pandas(df)
bdf
```
```
0     1.0
1     2.5
2    <NA>
3    <NA>
4    <NA>
dtype: Float64
```

TrevorBergeron: That is a limitation of our type-widening logic as we go from non-nullable float to nullable float. In this case here, we are mapping nullable float to nullable float. In nullable float type, nan is not the null value, so I think we should be able to keep it separate.

shobsi: handled (nan, inf, -inf) as special values, PTAL

bigframes/functions/remote_function.py
```diff
@@ -24,9 +24,13 @@
 import sys
 import tempfile
 import textwrap
+
+# TODO(shobs): import typing module and use its classes through namespapace.*
```
TrevorBergeron, line 28: google style guide has an exception for typing module

bigframes/core/blocks.py
```diff
@@ -2052,6 +2053,85 @@ def _is_monotonic(
...
+        if pandas_row_dtype is None:
+            pandas_row_dtype = "object"
+        pandas_row_dtype = f'"{pandas_row_dtype}"'
+
+        # create a json column representing row through SQL manipulation
+        row_json_column_name = guid.generate_guid()
+        select_columns = (
+            [ordering_column_name] + list(self.index_columns) + [row_json_column_name]
+        )
+        select_columns_csv = ", ".join(select_columns)
```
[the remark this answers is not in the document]
TrevorBergeron: we are still joining raw strings here and hoping they don't contain any special characters. Lets put common logic in `sql` module for generating sql identifiers that will do something like `"my id" -> "`my id`"` and ideally also escape quotes

bigframes/functions/remote_function.py
```diff
@@ -800,9 +898,25 @@ def remote_function(
             function's default setting applies. For more details see
             https://cloud.google.com/functions/docs/configuring/max-instances
     """
-    if isinstance(input_types, type):
+    is_row_processor = False
+
+    import bigframes.series
+
+    if input_types == bigframes.series.Series:
```
TrevorBergeron, line 905: Should we also handle pandas.Series?

bigframes/functions/remote_function.py
TrevorBergeron: great, literal_eval makes me feel better about this. One thing to note is that the eval-repr pattern only works for types that have literals in the python language. More complex/derived types like some datetime types, etc won't work with this pattern. I think it would be best to raise Errors on client side for these cases if posisble, until we can support serializing these types.

bigframes/functions/remote_function.py
shobsi: Um, what would it mean for the user? We are position `remote_function` as a way to apply their custom code on BigFrames DataFrames. It's not meant for applying to pandas DataFram, so why bring that up?

TrevorBergeron (approved):

bigframes/dataframe.py
```diff
@@ -3308,7 +3310,59 @@ def map(self, func, na_action: Optional[str] = None) -> DataFrame:
...
+            warnings.warn(
+                "axis=1 scenario is in preview.",
+                category=bigframes.exceptions.PreviewWarning,
+            )
+
+            # Early check whether the dataframe dtypes are currently supported
+            # in the remote function
+            # NOTE: Keep in sync with the value converters used in the gcf code
+            # generated in generate_cloud_function_main_code in remote_function.py
+            remote_function_supported_dtypes = (
```
tswast, line 3324: Ideally we'd document these limitations, too, right?

bigframes/functions/remote_function.py
shobsi: added early type in .apply

bigframes/core/blocks.py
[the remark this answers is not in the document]
shobsi: common logic was moved to SQL module

bigframes/functions/remote_function.py
shobsi: I notice pandas series was supported as a type in a later change https://github.com/googleapis/python-bigquery-dataframes/pull/717/files#diff-e4315dd3ecff62becbb3e104eaf9232f49553c94e3f8fec0c5108ab98a4256aa, so resolving this.

bigframes/dataframe.py
shobsi: Adding in https://github.com/googleapis/python-bigquery-dataframes/pull/800

PR #630, opened 2024-04-22
feat: add the `bigframes.bigquery` sub-package with a `bigframes.bigquery.array_length` function

chelsea-lin:


docs/templates/toc.yml
```diff
@@ -189,5 +189,7 @@
         uid: bigframes.ml.remote.VertexAIModel
       name: remote
     name: bigframes.ml
+  - name: Bigquery Built-in Functions
```
ashleyxuu, line 192: I think the structure is wrong here? It should be parallel with `bigframes.ml`, `bigframes.pandas` and start with `- items`, and we want to name it `bigframes.bigquery` or `Bigquery Built-in Functions`?

You can run `nox -s docfx` to see if it can generate the page for you:

https://g3doc.corp.google.com/devrel/cloud/infra/devrel_services/g3doc/dev/content_automation/cloud_rad/python.md?cl=head#staging-local-changes-from-github-to-devsite to stage docs manually.

tests/system/small/bigquery/test_array.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2023 Google LLC
```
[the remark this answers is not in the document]
chelsea-lin: Fixed. Thanks!

docs/templates/toc.yml
chelsea-lin: Good catch! It should be fixed. Also check the generated docs from this change: https://bigframes-swe.teams.x20web.corp.google.com/chelsealin/bigframes-docs-2024-04-22/reference/bigframes.bigquery/index.html

tests/system/small/bigquery/test_array.py
```diff
@@ -0,0 +1,30 @@
...
+import numpy as np
+import pandas as pd
+
+import bigframes.bigquery as bbq
+import bigframes.pandas as bpd
+
+
+def test_array_length():
+    series = bpd.Series([["A", "AA", "AAA"], ["BB", "B"], np.nan, [], ["C"]])
+    expected = pd.Series([3, 2, 0, 0, 1])
```
TrevorBergeron, line 24: Seems NULL -> NULL would be more consistent with other APIs than NULL -> 0?

tests/system/small/bigquery/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2024 Google LLC
```
TrevorBergeron, line 1: Do we need this file?

TrevorBergeron (commented):
Lets fix the spelling in the PR title as well

tests/system/small/bigquery/__init__.py
tswast: We'll need it if we ever want to import a file like https://github.com/googleapis/python-bigquery-dataframes/blob/main/tests/system/utils.py from this directory, I believe. Though some of that behavior has been simplified in newer versions of Python if I remember, correctly.

tests/system/small/bigquery/test_array.py
tswast: A possible problem here is I think the only way to express a NULL array in SQL is through inline SQL and yet I think ARRAY might be one of the cases where we always fallback to load jobs. Storing / returning null arrays are not possible.

bigframes/bigquery/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+        1    0
+        2    2
+        dtype: Int64
+
+    Returns:
+        bigframes.series.Series: A Series of integer values indicating
+            the length of each element in the Series.
+
+    """
+    return series._apply_unary_op(ops.len_op)
```
tswast, line 52: We might want to use the `ibis.udf.scalar.builtin` trick here to make sure we aren't introducing any ibis-isms. Ibis can make subtle changes to try and make the various backends more consistent, which isn't really desired in the `bigframes.bigquery` sub-package.

See:

https://github.com/googleapis/python-bigquery-dataframes/blob/2b84c4f173e956ba2c7fcc0ad92785ae95161d8e/bigframes/core/compile/scalar_op_compiler.py#L1403-L1406

docs/templates/toc.yml
```diff
@@ -189,5 +189,9 @@
         uid: bigframes.ml.remote.VertexAIModel
       name: remote
     name: bigframes.ml
+  - items:
+    - name: Bigquery Built-in Functions
```
tswast, line 193: Nit: BigQuery

```suggestion
    - name: BigQuery Built-in Functions
```

tests/system/small/bigquery/__init__.py
chelsea-lin: Thank you! The resource is helpful for learning more about the details of Python packages. I'll keep this empty file to imply the `bigquery` is a regular package.

tests/system/small/bigquery/test_array.py
chelsea-lin: Thanks for pointing out the issue. I will follow it up as a separate issue and track it in a internal bug 336880368

bigframes/bigquery/__init__.py
chelsea-lin: As discussed offline, this comment relates to the null array issue. Will follow up in the internal bug 336880368

tswast:
e2e failure: `FAILED tests/system/small/ml/test_llm.py::test_create_text_generator_32k_model` appears to be unrelated.

docs/templates/toc.yml
tswast, line 193: Nit: BigQuery

tests/system/small/bigquery/test_array.py
tswast: Thanks, let's add a TODO.

```suggestion
    # TODO(b/336880368): Allow for NULL values to be input for ARRAY columns.
    # Once we actually store NULL values, this will be NULL where the input is NULL.
    expected = pd.Series([3, 2, 0, 0, 1])
```

tswast (approved):
Thanks so much, love it! This is a big milestone and will allow for future expansion of BigQuery SQL APIs, including those without a direct analogue to pandas methods.

ashleyxuu (approved):

PR #631, opened 2024-04-22
feat: raise `NoDefaultIndexError` from `read_gbq` on clustered/partitioned tables with no `index_col` or `filters` set

tswast:
Please review https://github.com/googleapis/python-bigquery-dataframes/pull/636 first, as this PR builds on that.

Fixes internal issue 335727141
🦕


third_party/bigframes_vendored/pandas/io/gbq.py
```diff
@@ -107,11 +117,18 @@ def read_gbq(
...
 
+                If an empty iterable, such as ``()``, a default index is
+                generated. Do not depend on specific index values in this case.
+
                 **New in bigframes version 1.3.0**: If ``index_cols`` is not
                 set, the primary key(s) of the table are used as the index.
+
+                **New in bigframes version 1.4.0**: Support
+                :class:`bigframes.enums.TypeKind` to override default index
+                behavior.
```
TrevorBergeron, line 131: TypeKind or DefaultIndexKind?

third_party/bigframes_vendored/pandas/io/gbq.py
```diff
@@ -28,16 +29,25 @@ def read_gbq(
...
+          analytic windowed operation that prevents filtering push down. Avoid
+          using on large clustered or partitioned tables.
         * (Recommended) Set the ``index_col`` argument to one or more columns.
           Unique values for the row labels are recommended. Duplicate labels
           are possible, but note that joins on a non-unique index can duplicate
-          rows and operations like ``cumsum()`` that window across a non-unique
-          index can have some non-deternimism.
+          rows via pandas-like outer join behavior. Operations like
+          ``cumsum()`` that window across a non-unique index can have some
+          unpredictability due to ambiguous ordering.
```
TrevorBergeron, line 50: The part about non-determinism I don't think is correct? My understanding is if the index is non-unique, we fall back to hidden hashes to ensure total ordering.

bigframes/__init__.py
```diff
@@ -25,6 +27,8 @@
     "BigQueryOptions",
     "get_global_session",
     "close_session",
+    "enums",
```
TrevorBergeron, line 30: Is enums an intuitive module, or would a domain-related term be better, eg `indexing.IndexType` or directly putting the enum in the main module, `bigframes.pandas.IndexType`?

third_party/bigframes_vendored/pandas/io/gbq.py
```diff
@@ -107,11 +117,18 @@ def read_gbq(
                 `project.dataset.tablename` or `dataset.tablename`.
                 Can also take wildcard table name, such as `project.dataset.table_prefix*`.
                 In tha case, will read all the matched table as one DataFrame.
-            index_col (Iterable[str] or str):
+            index_col (Iterable[str], str, bigframes.enums.IndexKind):
```
TrevorBergeron, line 120: DefaultIndexKind?

bigframes/__init__.py
tswast: I tried to find some guidance on this, but Python community doesn't seem particularly prescriptive about module names.

PEP-8 has this to say:

> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability.

https://peps.python.org/pep-0008/#package-and-module-names

Google Python style guide has a bit more to say:

> Place related classes and top-level functions together in a module. Unlike Java, there is no need to limit yourself to one class per module.
>
> Use CapWords for class names, but lower_with_under.py for module names.

https://google.github.io/styleguide/pyguide.html#3162-naming-conventions

I tried a few of these options out locally (bigframes.indexes.DefaultIndexKind and bigframes.pandas.DefaultIndexKind), but it feels strange to have something not really mimicking pandas in the pandas sub-package and bigframes.indexes.DefaultIndexKind would imply that we should move the Index and MultiIndex classes there, which is kinda the opposite of what we want to do.

The other option we could try is `bigframes.pandas.core.indexes`, but in pandas "core" is how they signify that an API is private an not to be relied on.

IMO, determining if classes are "related" by type for the basic types (e.g. exceptions, enums, ...) will be less effort for us long-term than having to figure out which public package to place these things if it doesn't fit in an existing API.

third_party/bigframes_vendored/pandas/io/gbq.py
tswast: Dropped this.

third_party/bigframes_vendored/pandas/io/gbq.py
tswast: Good catch. Done.

third_party/bigframes_vendored/pandas/io/gbq.py
tswast: DefaultIndexKind. Good catch. I had tried a few names for this but the refactor option didn't catch docstrings I think.

TrevorBergeron (approved):

tswast:
Looks like `test_read_csv_bq_engine_throws_not_implemented_error` is a real failure. Will address.

tswast:
> Looks like test_read_csv_bq_engine_throws_not_implemented_error is a real failure. Will address.

Done! Looks like from https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html `index_col=False` is equivalent to our `DefaultIndexKind.SEQUENTIAL_INT64`. I see no reason to raise NotImplementedError in that case, so I've updated those tests and made some slight tweaks so that they are treated the same.

PR #632, opened 2024-04-22
fix: Extend row hash to 128 bits to guarantee unique row id

TrevorBergeron:

### DIFF
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -835,11 +835,11 @@ class Session(
                 )
             else:
                 array_value = self._create_total_ordering(
-                    table_expression, table_size=table.num_rows
+                    table_expression, table_rows=table.num_rows
                 )
         else:
             array_value = self._create_total_ordering(
-                table_expression, table_size=table.num_rows
+                table_expression, table_rows=table.num_rows
             )
 
         value_columns = [col for col in array_value.column_ids if col not in index_cols]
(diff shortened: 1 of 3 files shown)


tswast:
> ERROR tests/system/small/ml/test_llm.py::test_create_embedding_generator_multilingual_model

Looks like a test flake. Will raise an issue with our on-call this week.

bigframes/session/__init__.py
```diff
@@ -832,9 +834,13 @@ def _read_gbq_table(
                     ordering=ordering,
                 )
             else:
-                array_value = self._create_total_ordering(table_expression)
+                array_value = self._create_total_ordering(
+                    table_expression, table_size=table.num_rows
```
tswast, line 838: Could you please try this with a VIEW type table? I don't think `num_rows` is always populated.

bigframes/session/__init__.py
```diff
@@ -1455,10 +1461,18 @@ def _create_empty_temp_table(
     def _create_total_ordering(
         self,
         table: ibis_types.Table,
+        table_size: int,
```
tswast, line 1464: Likewise, we might need to handle missing size here.

tswast (changes requested):

bigframes/session/__init__.py
TrevorBergeron: added.

bigframes/session/__init__.py
TrevorBergeron: yes, this won't be populated for views or potentially with streaming inserts. defaulting to assume table is large and need 128 bits.

tswast (approved):

PR #636, opened 2024-04-24
refactor: cache table metadata alongside snapshot time

tswast:
This ensures the cached `primary_keys` is more likely to be correct, in case the user called ALTER TABLE after we originally cached the snapshot time.

Towards internal issue 335727141 and https://github.com/googleapis/python-bigquery-dataframes/pull/631
🦕

### DIFF
--- a/bigframes/session/_io/bigquery.py
+++ b/bigframes/session/_io/bigquery.py
@@ -23,6 +23,7 @@ import textwrap
 import types
 from typing import Dict, Iterable, Optional, Sequence, Tuple, Union
 import uuid
+import warnings
 
 import google.api_core.exceptions
 import google.cloud.bigquery as bigquery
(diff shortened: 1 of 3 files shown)


bigframes/session/_io/bigquery.py
```diff
@@ -121,6 +122,59 @@ def table_ref_to_sql(table: bigquery.TableReference) -> str:
...
 
+def get_snapshot_datetime_and_table_metadata(
+    bqclient: bigquery.Client,
+    table_ref: bigquery.TableReference,
+    *,
+    api_name: str,
+    cache: Dict[bigquery.TableReference, Tuple[datetime.datetime, bigquery.Table]],
+    use_cache: bool = True,
+) -> Tuple[datetime.datetime, bigquery.Table]:
+    cached_table = cache.get(table_ref)
```
chelsea-lin, line 133: nit: if `use_cache=False`, we can avoid looking through the cache, though the cache won't be too large and the looking is not too expensive.

chelsea-lin (approved):
LGTM

bigframes/session/_io/bigquery.py
tswast: Good observation!

Python does dictionary lookups all the time, even for accessing attributes of objects, so I'm not too worried about it. Will leave as-is to avoid the extra nesting.

PR #643, opened 2024-04-29
fix: llm palm score tests

ashleyxuu:

Fixes internal # 336527025🦕

### DIFF
--- a/tests/system/load/test_llm.py
+++ b/tests/system/load/test_llm.py
@@ -49,6 +49,7 @@ def llm_remote_text_df(session, llm_remote_text_pandas_df):
     return session.read_pandas(llm_remote_text_pandas_df)
 
 
+@pytest.mark.flaky(retries=2)
 def test_llm_palm_configure_fit(llm_fine_tune_df_default_index, llm_remote_text_df):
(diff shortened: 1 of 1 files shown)


tests/system/load/test_llm.py
```diff
@@ -54,7 +54,7 @@ def test_llm_palm_configure_fit(llm_fine_tune_df_default_index, llm_remote_text_
         model_name="text-bison", max_iterations=1
     )
 
-    df = llm_fine_tune_df_default_index.dropna()
+    df = llm_fine_tune_df_default_index.dropna().sample(n=100)
```
GarrettWu, line 57: Let's add retry for experimental features.

GarrettWu, line 57: curious how much run time reduction with sampling.

GarrettWu (approved):

tests/system/load/test_llm.py
ashleyxuu: Done.

ashleyxuu: It didn't improve much. It still takes the same amount of time around 30-40 mins, and the durations has some variations.

PR #650, opened 2024-05-01
test: explicitly use US location for session in tests

tswast:
This avoids some warnings we see and ignore in our tests.

It might also address some flakiness in
`tests/system/small/ml/test_llm.py::test_create_text_generator_model` and
`tests/system/small/ml/test_llm.py::test_create_text_generator_32k_model`, but the root cause of that flakiness is still TBD.

### DIFF
--- a/tests/system/conftest.py
+++ b/tests/system/conftest.py
@@ -128,7 +128,10 @@ def resourcemanager_client(
 
 @pytest.fixture(scope="session")
 def session() -> bigframes.Session:
-    return bigframes.Session()
+    context = bigframes.BigQueryOptions(
+        location="US",
+    )
+    return bigframes.Session(context=context)
 
 
 @pytest.fixture(scope="session")


GarrettWu (approved):

ashleyxuu (approved):

tests/system/conftest.py
```diff
@@ -128,7 +128,10 @@ def resourcemanager_client(
 
 @pytest.fixture(scope="session")
 def session() -> bigframes.Session:
-    return bigframes.Session()
+    context = bigframes.BigQueryOptions(
+        location="US",
```
shobsi, line 132: This should already be the default https://github.com/googleapis/python-bigquery-dataframes/blob/c67e501a4958ac097216cc1c0a9d5c1530c87ae5/bigframes/session/__init__.py#L182

tswast: Note that the line below that raises a warning. So this avoids that noise at least, even if it doesn't fix the issue.

PR #651, opened 2024-05-01
fix: use explicit session in `PaLM2TextGenerator`

shobsi:
This is to avoid running into conflict with the global session which may have different options (e.g. location, connection etc.) set.

### DIFF
--- a/bigframes/ml/llm.py
+++ b/bigframes/ml/llm.py
@@ -144,7 +144,7 @@ class PaLM2TextGenerator(base.BaseEstimator):
         kwargs: dict = {}
         last_fitting = model.training_runs[-1]["trainingOptions"]
 
-        dummy_text_generator = cls()
+        dummy_text_generator = cls(session=session)
         for bf_param, _ in dummy_text_generator.__dict__.items():
             bqml_param = _BQML_PARAMS_MAPPING.get(bf_param)
             if bqml_param in last_fitting:


tswast (approved):

ashleyxuu (approved):

bigframes/ml/llm.py
```diff
@@ -144,7 +144,7 @@ def _from_bq(
         kwargs: dict = {}
         last_fitting = model.training_runs[-1]["trainingOptions"]
 
-        dummy_text_generator = cls()
+        dummy_text_generator = cls(session=session)
```
GarrettWu, line 147: I don't think this is the issue. It is only a dummy object to retrieve param types. The actual creation is in line 155.

GarrettWu (changes requested):

bigframes/ml/llm.py
shobsi: This does call the `.ctor` -> `_create_bqml_model` -> `create_bq_connection`?

GarrettWu: I see. Thanks for figuring out.

GarrettWu (approved):

bigframes/ml/llm.py
GarrettWu: Actually this whole block is unnecessary and does more than what we want. I'll send another PR to refactor some.

PR #652, opened 2024-05-01
feat: `bigframes.options` and  `bigframes.option_context` now uses thread-local variables to prevent context managers in separate threads from affecting each other

tswast:
In our tests, this allows us to actually test things like `bf.option_context("display.repr_mode", "deferred"):` without always having some other test change the display mode and break the test.

Fixes internal issue 308657813

### DIFF
--- a/bigframes/_config/__init__.py
+++ b/bigframes/_config/__init__.py
@@ -48,7 +48,8 @@ class Options:
         """Initialize thread-local options, based on current global options."""
 
         # Already thread-local, so don't reset any options that have been set
-        # already.
+        # already. No locks needed since this only modifies thread-local
+        # variables.
         if self._local.is_bigquery_thread_local:
             return
 
--- a/bigframes/_config/__init__.py
+++ b/bigframes/_config/__init__.py
@@ -44,7 +44,6 @@ class Options:
         # BigQuery options are special because they can only be set once per
         # session, so we need an indicator as to whether we are using the
         # thread-local session or the global session.
-        self._local.is_bigquery_thread_local = False
         self._bigquery_options = bigquery_options.BigQueryOptions()
 
     def _init_bigquery_thread_local(self):
@@ -53,20 +52,18 @@ class Options:
         # Already thread-local, so don't reset any options that have been set
         # already. No locks needed since this only modifies thread-local
         # variables.
-        if self._local.is_bigquery_thread_local:
+        if self._local.bigquery_options is not None:
             return
 
-        self._local.is_bigquery_thread_local = True
         self._local.bigquery_options = copy.deepcopy(self._bigquery_options)
         self._local.bigquery_options._session_started = False
 
     @property
     def bigquery(self) -> bigquery_options.BigQueryOptions:
         """Options to use with the BigQuery engine."""
-        if self._local.is_bigquery_thread_local:
+        if self._local.bigquery_options is not None:
             # The only way we can get here is if someone called
-            # _init_bigquery_thread_local, which initializes
-            # self._local.bigquery_options.
+            # _init_bigquery_thread_local.
             return self._local.bigquery_options
 
         return self._bigquery_options
@@ -106,9 +103,10 @@ class Options:
     def is_bigquery_thread_local(self) -> bool:
         """Indicator that we're using a thread-local session.
 
-        This is set to True by using `option_context` with a "bigquery" option.
+        A thread-local session can be started by using
+        `with bigframes.option_context("bigquery.some_option", "some-value"):`.
         """
-        return self._local.is_bigquery_thread_local
+        return self._local.bigquery_options is not None
 
 
 options = Options()
--- a/bigframes/_config/__init__.py
+++ b/bigframes/_config/__init__.py
@@ -44,7 +44,7 @@ class Options:
         self._bigquery_options = bigquery_options.BigQueryOptions()
         self._local.bigquery_options = None
 
-    def _init_biquery_thread_local(self):
+    def _init_bigquery_thread_local(self):
         """Initialize thread-local options, based on current global options."""
 
         # Already thread-local, so don't reset any options that have been set
--- a/bigframes/_config/__init__.py
+++ b/bigframes/_config/__init__.py
@@ -33,16 +33,19 @@ class Options:
 
     def __init__(self):
         self._local = threading.local()
-        self._local.display_options = display_options.DisplayOptions()
-        self._local.sampling_options = sampling_options.SamplingOptions()
-        self._local.compute_options = compute_options.ComputeOptions()
+
+        # Initialize these in the property getters to make sure we do have a
+        # separate instance per thread.
+        self._local.bigquery_options = None
+        self._local.display_options = None
+        self._local.sampling_options = None
+        self._local.compute_options = None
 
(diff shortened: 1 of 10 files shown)


bigframes/_config/__init__.py
```diff
@@ -29,44 +32,77 @@ class Options:
...
+        # session, so we need an indicator as to whether we are using the
+        # thread-local session or the global session.
+        self._local.is_bigquery_thread_local = False
         self._bigquery_options = bigquery_options.BigQueryOptions()
-        self._display_options = display_options.DisplayOptions()
-        self._sampling_options = sampling_options.SamplingOptions()
-        self._compute_options = compute_options.ComputeOptions()
+        self._local.bigquery_options = None
+
+    def _init_biquery_thread_local(self):
```
GarrettWu, line 47: typo: bigquery

tswast: Thanks. Fixed in [4fa1a92](https://github.com/googleapis/python-bigquery-dataframes/pull/652/commits/4fa1a924782947aa343642e6f31600d969e62392)

tswast:
e2e failure of `test_model_register` appears to be unrelated.

tswast:
Cover failed.

`tests/system/small/test_progress_bar.py                                72      1     28      1    98%   144`

Looks like I now have some dead code in that test. Will investigate.

milkshakeiii:
Messaged offline. @tswast will fix the failing test, and then I will review.

tswast:
I'm able to reproduce locally. The thread local variable is working how I expect, but the test is making an assumption that the DataFrame doesn't have a `query_job` populated. This assumption is violated when the DataFrame is used across multiple tests.

I'll update the test.

```
________________________________________________________________________________ test_query_job_dry_run ________________________________________________________________________________
[gw2] linux -- Python 3.11.1 /usr/local/google/home/swast/envs/bigframes/bin/python3.11

penguins_df_default_index =                                       species     island  culmen_length_mm  \
0           Gentoo penguin (Pygoscelis p...  <NA>         <NA>    <NA>  
24             20.7              191.0       3900.0  FEMALE  
...

[344 rows x 7 columns]

    def test_query_job_dry_run(penguins_df_default_index: bf.dataframe.DataFrame):
        expected = "Computation deferred. Computation will process"
    
        with bf.option_context("display.repr_mode", "deferred"):
            df_result = repr(penguins_df_default_index)
>           assert expected in df_result
E           AssertionError: assert 'Computation deferred. Computation will process' in 'Query Job Info\nJob url: https://console.cloud.google.com/bigquery?project=swast-scratch&j=bq:US:9a761424-b164-4c31-a...240503_session444e7b_fbabcb4962f841a8bb92ec2f27d17bb2\nSlot Time: a moment\nBytes Processed: 28.9 kB\nCache hit: False'

tests/system/small/test_progress_bar.py:142: AssertionError
--------------------------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------------------------
DisplayOptions(max_columns=20, max_rows=25, progress_bar='auto', repr_mode='deferred', max_info_columns=100, max_info_rows=200000, memory_usage=True)
deferred
```

bigframes/_config/__init__.py
```diff
@@ -29,44 +32,93 @@ class Options:
...
+        # separate instance per thread.
+        self._local.bigquery_options = None
+        self._local.display_options = None
+        self._local.sampling_options = None
+        self._local.compute_options = None
+
+        # BigQuery options are special because they can only be set once per
+        # session, so we need an indicator as to whether we are using the
+        # thread-local session or the global session.
+        self._local.is_bigquery_thread_local = False
```
milkshakeiii, line 47: If I'm reading this correctly, would checking for self._local.bigquery_options is None be able to replace this boolean state? It seems a little more DRY to do it that way, if so.

bigframes/_config/__init__.py
```diff
@@ -29,44 +32,93 @@ class Options:
...
-        self._sampling_options = sampling_options.SamplingOptions()
-        self._compute_options = compute_options.ComputeOptions()
+
+    def _init_bigquery_thread_local(self):
+        """Initialize thread-local options, based on current global options."""
+
+        # Already thread-local, so don't reset any options that have been set
+        # already. No locks needed since this only modifies thread-local
+        # variables.
+        if self._local.is_bigquery_thread_local:
```
milkshakeiii, line 56: i.e., this could be "if self._local.bigquery_options is not None"

bigframes/_config/__init__.py
```diff
@@ -29,44 +32,93 @@ class Options:
...
+            return
+
+        self._local.is_bigquery_thread_local = True
+        self._local.bigquery_options = copy.deepcopy(self._bigquery_options)
+        self._local.bigquery_options._session_started = False
 
     @property
     def bigquery(self) -> bigquery_options.BigQueryOptions:
         """Options to use with the BigQuery engine."""
+        if self._local.is_bigquery_thread_local:
```
milkshakeiii, line 66: Here too

bigframes/_config/__init__.py
```diff
@@ -29,44 +32,93 @@ class Options:
...
+
+        return self._local.compute_options
+
+    @property
+    def is_bigquery_thread_local(self) -> bool:
+        """Indicator that we're using a thread-local session.
+
+        This is set to True by using `option_context` with a "bigquery" option.
+        """
+        return self._local.is_bigquery_thread_local
```
milkshakeiii, line 111: This could also check none (and the comment would need to be tweaked.)

third_party/bigframes_vendored/pandas/_config/config.py
```diff
@@ -40,7 +51,19 @@ def __exit__(self, *args) -> None:
             for pat, val in self.undo:
                 self._set_option(pat, val)
 
+        # TODO(tswast): What to do if someone nests several context managers
+        # with separate "bigquery" options? We might need a "stack" of
+        # sessions if we allow that.
+        if bigframes.options.is_bigquery_thread_local:
+            bigframes.close_session()
+            bigframes.options._local.is_bigquery_thread_local = False
```
milkshakeiii, line 59: close_session could set _local.bigquery_options to None

milkshakeiii (changes requested):
Just had one simplification suggestion, feel free to just close and re-request if you prefer the way it is now. Thanks!

third_party/bigframes_vendored/pandas/_config/config.py
tswast: Someone could call `close_session()` inside of a context manager. We actually have some tests that start and stop multiple sessions in this thread-local context. By using the options as the thread locality indicator, we need to make sure that object stays around.

bigframes/_config/__init__.py
tswast: I like it! Done in https://github.com/googleapis/python-bigquery-dataframes/pull/652/commits/a7f3c967e0e1be92189c1123d33651901f4561b9

third_party/bigframes_vendored/pandas/_config/config.py
milkshakeiii: Oh, right, makes sense.

milkshakeiii (approved):
LGTM, thanks!

PR #656, opened 2024-05-02
refactor: ml model load read from class type hints

GarrettWu:


bigframes/ml/cluster.py
```diff
@@ -34,6 +34,7 @@
     "distance_type": "distanceType",
     "max_iter": "maxIterations",
     "tol": "minRelativeProgress",
+    "warm_start": "warmStart",
```
ashleyxuu, line 37: We should add the unit test for all these newly-added params. Same for all other BQML models, but feel free to address in a follow-up PR.

bigframes/ml/decomposition.py
```diff
@@ -27,6 +27,8 @@
 from bigframes.ml import base, core, globals, utils
 import bigframes.pandas as bpd
 
+_BQML_PARAMS_MAPPING = {"svd_solver": "pcaSolver"}
```
ashleyxuu, line 30: Same here.

GarrettWu: This one isn't a new adding, but moved piece.

bigframes/ml/cluster.py
GarrettWu: yes, will do a follow up.

ashleyxuu (approved):

PR #662, opened 2024-05-06
feat: add `DefaultIndexKind.NULL` to use as `index_col` in `read_gbq*`, creating an indexless DataFrame/Series

TrevorBergeron:


bigframes/core/__init__.py
```diff
@@ -453,6 +453,19 @@ def join(
             return ArrayValue(bigframes.core.rewrite.maybe_rewrite_join(join_node))
         return ArrayValue(join_node)
 
+    def try_align_as_projection(
+        self,
+        other: ArrayValue,
+        join_type: join_def.JoinType,
+        mappings: typing.Tuple[join_def.JoinColumnMapping, ...],
+    ) -> typing.Optional[ArrayValue]:
```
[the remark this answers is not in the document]
tswast: Aside: This is slightly safer to return None now, since we have type checking. But, it's not universally applied (see all the `note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]` warnings in our mypy session.

I created Draft PR https://github.com/googleapis/python-bigquery-dataframes/pull/693 to fix that, but there's `151 errors in 32 files (checked 187 source files)` to cleanup.

bigframes/core/blocks.py
```diff
@@ -2062,6 +2062,12 @@ def _is_monotonic(
         self._stats_cache[column_name].update({op_name: result})
         return result
 
+    def _null_index_guard(self):
+        if len(self.index_columns) == 0:
+            raise bigframes.exceptions.NullIndexError(
+                "Cannot perform this operation without an index. Set an index using set_index."
```
tswast, line 2068: Let's add an `operation` argument to the parameters so users have a bit more context about what caused this in the message itself. We could default it to `operation="this operation"` if there are cases where it doesn't make sense to pass one in.

That'd allow us to be more specific too if there are certain combinations of parameters that aren't possible. It could be hard to determine from the stack trace in those cases.

bigframes/core/rewrite.py
```diff
@@ -113,10 +111,17 @@ def maybe_join(
         if (self.root != right.root) or any(
             l_expr != r_expr for l_expr, r_expr in zip(l_join_exprs, r_join_exprs)
         ):
+            return False
+        return True
+
+    def merge(
```
tswast, line 117: maybe_merge? I'd like to telegraph that this can return None.

bigframes/dataframe.py
```diff
@@ -610,7 +629,14 @@ def __repr__(self) -> str:
         column_count = len(pandas_df.columns)
 
         with display_options.pandas_repr(opts):
-            repr_string = repr(pandas_df)
+            import pandas.io.formats
+
+            options = (
```
tswast, line 634: Can we rename to `to_string_kwargs`? I got confused and thought we were modifying a global dictionary.

bigframes/dataframe.py
```diff
@@ -610,7 +629,14 @@ def __repr__(self) -> str:
...
+                pandas.io.formats.format.get_dataframe_repr_params()  # type: ignore
```
tswast, line 635: We're sure this returns a copy, not the global instance?

bigframes/dataframe.py
```diff
@@ -2034,16 +2063,15 @@ def quantile(
             frame._block, frame._block.value_columns, qs=tuple(q) if multi_q else (q,)  # type: ignore
         )
         if multi_q:
-            return DataFrame(result.stack()).droplevel(0)
+            return DataFrame(result.stack())  # .droplevel(0)
```
tswast, line 2066: Is commenting out droplevel intended?

bigframes/dataframe.py
```diff
@@ -3518,3 +3546,9 @@ def __matmul__(self, other) -> DataFrame:
         return self.dot(other)
 
     __matmul__.__doc__ = inspect.getdoc(vendored_pandas_frame.DataFrame.__matmul__)
+
+    def _null_index_guard(self):
```
tswast, line 3550: Same here, let's make this a verb.

bigframes/enums.py
```diff
@@ -27,3 +27,5 @@ class DefaultIndexKind(enum.Enum):
     #: ``n - 3``, ``n - 2``, ``n - 1``, where ``n`` is the number of items in
     #: the index.
     SEQUENTIAL_INT64 = enum.auto()
+    # A completely null index incapable of indexing or alignment.
+    EMPTY = enum.auto()
```
tswast, line 31: Let's align this with the exception type, either `NULL` here or `EmptyIndexError` there.

bigframes/series.py
```diff
@@ -288,7 +289,17 @@ def __repr__(self) -> str:
         pandas_df, _, query_job = self._block.retrieve_repr_request_results(max_results)
         self._set_internal_query_job(query_job)
 
-        return repr(pandas_df.iloc[:, 0])
+        pd_series = pandas_df.iloc[:, 0]
+
+        with display_options.pandas_repr(opts):
+            import pandas.io.formats
+
+            options = pandas.io.formats.format.get_series_repr_params()  # type: ignore
```
tswast, line 297: Ditto, let's use `kwargs` in the variable name and confirm we are working with a copy, not modifying global state. Please add a comment with what you discover regarding global state or not.

bigframes/session/__init__.py
```diff
@@ -755,15 +757,25 @@ def _read_gbq_table(
             )
 
         # ----------------------------------------------------
-        # Create Block & default index if len(index_cols) == 0
+        # Create Default Sequential Index if still have no index
         # ----------------------------------------------------
 
+        if not index_col and len(index_cols) == 0:
```
tswast, line 763: This reads odd to me: `len(index_cols) == 0` implies `not index_col`. Could you rephrase what you're trying to check here? I assume you're trying to exclude the DefaultIndexKind enum from this check. Let's be explicit about that.

tests/system/small/test_empty_index.py
```diff
@@ -0,0 +1,168 @@
+import pandas as pd
```
tswast, line 1: Needs license header.

tests/system/small/test_empty_index.py
```diff
@@ -0,0 +1,168 @@
...
+
+import bigframes.exceptions
+import bigframes.pandas as bpd
+
+
+def test_empty_index_materialize(
+    scalars_df_empty_index, scalars_pandas_df_default_index
+):
+    bf_result = scalars_df_empty_index.to_pandas()
+    pd.testing.assert_frame_equal(
```
tswast, line 12: presumably we'll need to make this robust to non-total ordering at some point, right?

tests/system/small/test_empty_index.py
tswast, line 1: What are your overall thoughts on testing? How can we be confident that empty/null index works? As we add operations that should support null/empty index, we add tests here and in the usual location?

bigframes/core/blocks.py
```diff
@@ -117,19 +117,6 @@ def __init__(
...
 
-        # If no index columns are set, create one.
-        #
-        # Note: get_index_cols_and_uniqueness in
-        # bigframes/session/_io/bigquery/read_gbq_table.py depends on this
-        # being as sequential integer index column. If this default behavior
-        # ever changes, please also update get_index_cols_and_uniqueness so
-        # that users who explicitly request a sequential integer index can
-        # still get one.
-        if len(index_columns) == 0:
```
tswast, line 128: Would this be a good opportunity to raise some subclass of PreviewWarning? I'm not 100% confident that we've covered all the cases in tests well enough to call this feature "GA".

tswast (changes requested):

bigframes/dataframe.py
TrevorBergeron: renamed

bigframes/dataframe.py
TrevorBergeron: yes, its a brand new dict, safe to mutate. adding comments to code to communicate this

bigframes/core/blocks.py
TrevorBergeron: Added preview warning

tests/system/small/test_empty_index.py
TrevorBergeron: Hmm, I think it will be tricky, as null index tests will require manually configured expectations for many tests, as we cannot always compare against pandas (which doesn't have null index). So yeah, I think we will be stuck with a parallel test suite, which will be a bit burdensome to maintain. Being pandas-equivalent has been a huge boon for testing thus far.

tests/system/small/test_empty_index.py
TrevorBergeron: We do have a variant that ignore ordering completely. Once we bring in unordered mode, we will want to validate partial orderings, which we don't have a utility for yet.

tests/system/small/test_empty_index.py
TrevorBergeron: added

bigframes/session/__init__.py
TrevorBergeron: yeah, basically, trying to fall back to sequential index if don't have null index, user provided index columns, or metadata-derived index columns. Rewrote the condition though with the code structure, its still a bit weird.

bigframes/series.py
TrevorBergeron: done, added comment assuring its not global state

bigframes/enums.py
TrevorBergeron: renamed to null

bigframes/dataframe.py
TrevorBergeron: done

bigframes/dataframe.py
TrevorBergeron: should just be removed, removed

bigframes/core/rewrite.py
TrevorBergeron: done

bigframes/core/blocks.py
TrevorBergeron: added

bigframes/core/blocks.py
```diff
@@ -2062,6 +2062,12 @@ def _is_monotonic(
         self._stats_cache[column_name].update({op_name: result})
         return result
 
+    def _null_index_guard(self):
```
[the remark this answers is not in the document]
TrevorBergeron: done

bigframes/core/__init__.py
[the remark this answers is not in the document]
TrevorBergeron: I tend to prefer capturing expected failures in the return value rather than as an exception, for the type safety, and easier composition. The calling code can provide a more informative exception anyways and errors like this don't need a stack trace.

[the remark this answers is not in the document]
tswast: Resolving. I feel better about this now that https://github.com/googleapis/python-bigquery-dataframes/pull/693 is merged.

bigframes/core/blocks.py
```diff
@@ -1923,6 +1929,16 @@ def join(
             raise NotImplementedError(
                 f"Only how='outer','left','right','inner' currently supported. {constants.FEEDBACK_LINK}"
             )
+        # Special case for null index,
+        if (
+            (self.index.nlevels == other.index.nlevels == 0)
+            and (sort is False)
+            and (block_identity_join is False)
```
tswast, line 1936: Since these are only bool per type checking, we don't have to worry about any false-y values to mess up "not"

```suggestion
            and not sort
            and not block_identity_join
```

tswast: Question: Doesn't `block_identity_join` imply indexless join _is_ allowed? What's this checking for?

Edit: found it

https://github.com/googleapis/python-bigquery-dataframes/blob/ca284cc21d5386b4817ecab85b68235456361815/bigframes/core/blocks.py#L2292

Apparently `block_identity_join` is the opposite of what I thought it is. Please add a docstring to this function explaining what these args mean.

bigframes/core/rewrite.py
```diff
@@ -201,20 +206,25 @@ def maybe_squash_projection(node: nodes.BigFrameNode) -> nodes.BigFrameNode:
...
     left_side = SquashedSelect.from_node(join_node.left_child)
     right_side = SquashedSelect.from_node(join_node.right_child)
-    joined = left_side.maybe_join(right_side, join_node.join)
-    if joined is not None:
-        return joined.expand()
+    if left_side.can_join(right_side, join_node.join):
+        merged = left_side.maybe_merge(
+            right_side, join_node.join.type, join_node.join.mappings
+        )
+        assert merged is not None
```
tswast, line 213: If we do raise an assertion error somehow, let's include the feedback link.

```suggestion
        import bigframes.constants
        assert (merged is not None), "Couldn't merge nodes. This shouldn't happen. Please share full stacktrace with the BigQuery DataFrames team at bigframes-feedback@google.com." 
```

bigframes/session/__init__.py
```diff
@@ -800,15 +802,28 @@ def _read_gbq_table(
             )
 
         # ----------------------------------------------------
-        # Create Block & default index if len(index_cols) == 0
+        # Create Default Sequential Index if still have no index
         # ----------------------------------------------------
 
+        # If no idnex columns provided or found, fall back to sequential index
```
tswast, line 808:
```suggestion
        # If no index columns provided or found, fall back to sequential index
```

bigframes/core/blocks.py
TrevorBergeron: added docstring

bigframes/core/blocks.py
```diff
@@ -1910,6 +1915,7 @@ def join(
         sort=False,
```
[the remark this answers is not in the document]
TrevorBergeron: added

tswast (approved):
Thanks!

tswast:
Test failure `FAILED tests/system/large/test_remote_function.py::test_remote_function_via_session_vpc_invalid` appears to be unrelated.

PR #663, opened 2024-05-06
feat: bigframes.bigquery.array_agg(SeriesGroupBy|DataFrameGroupby)

chelsea-lin:
This change introduces the `bigframes.bigquery.array_agg` method for `SeriesGroupBy` and `DataFrameGroupby`. By default, aggregated arrays are ordered by the underlying sorting columns. Additionally, `array_agg` is the inverse operation of `(Series|Dataframe).explode()`.

Fixes internal bug: 338232748🦕

### DIFF
--- a/bigframes/core/compile/aggregate_compiler.py
+++ b/bigframes/core/compile/aggregate_compiler.py
@@ -38,7 +38,9 @@ def compile_aggregate(
 ) -> ibis_types.Value:
     if isinstance(aggregate, ex.UnaryAggregation):
         input = scalar_compiler.compile_expression(aggregate.arg, bindings=bindings)
-        return compile_unary_agg(aggregate.op, input, order_by=order_by)
+        return compile_unary_agg(
+            aggregate.op, input, order_by=order_by if aggregate.op.can_order_by else []
+        )
     elif isinstance(aggregate, ex.BinaryAggregation):
         left = scalar_compiler.compile_expression(aggregate.left, bindings=bindings)
         right = scalar_compiler.compile_expression(aggregate.right, bindings=bindings)
--- a/bigframes/core/compile/aggregate_compiler.py
+++ b/bigframes/core/compile/aggregate_compiler.py
@@ -34,13 +34,11 @@ scalar_compiler = scalar_compilers.scalar_op_compiler
 def compile_aggregate(
     aggregate: ex.Aggregation,
     bindings: typing.Dict[str, ibis_types.Value],
+    agg_order_by: typing.Sequence[ibis_types.Value] = [],
 ) -> ibis_types.Value:
     if isinstance(aggregate, ex.UnaryAggregation):
         input = scalar_compiler.compile_expression(aggregate.arg, bindings=bindings)
-        return compile_unary_agg(
-            aggregate.op,
-            input,
-        )
+        return compile_unary_agg(aggregate.op, input, agg_order_by=agg_order_by)
     elif isinstance(aggregate, ex.BinaryAggregation):
         left = scalar_compiler.compile_expression(aggregate.left, bindings=bindings)
         right = scalar_compiler.compile_expression(aggregate.right, bindings=bindings)
@@ -54,6 +52,7 @@ def compile_analytic(
     window: window_spec.WindowSpec,
     bindings: typing.Dict[str, ibis_types.Value],
 ) -> ibis_types.Value:
+    # TODO: check how to trigger here!
     if isinstance(aggregate, ex.UnaryAggregation):
         input = scalar_compiler.compile_expression(aggregate.arg, bindings=bindings)
         return compile_unary_agg(aggregate.op, input, window)
@@ -77,13 +76,19 @@ def compile_unary_agg(
     op: agg_ops.WindowOp,
     input: ibis_types.Column,
     window: Optional[window_spec.WindowSpec] = None,
+    agg_order_by: typing.Sequence[ibis_types.Value] = [],
 ) -> ibis_types.Value:
     raise ValueError(f"Can't compile unrecognized operation: {op}")
 
 
 def numeric_op(operation):
     @functools.wraps(operation)
-    def constrained_op(op, column: ibis_types.Column, window=None):
+    def constrained_op(
+        op,
+        column: ibis_types.Column,
+        window=None,
+        agg_order_by: typing.Sequence[ibis_types.Value] = [],
+    ):
         if column.type().is_boolean():
             column = typing.cast(
                 ibis_types.NumericColumn, column.cast(ibis_dtypes.int64)
@@ -104,7 +109,10 @@ def numeric_op(operation):
 @compile_unary_agg.register
 @numeric_op
 def _(
-    op: agg_ops.SumOp, column: ibis_types.NumericColumn, window=None
+    op: agg_ops.SumOp,
+    column: ibis_types.NumericColumn,
+    window=None,
+    agg_order_by: typing.Sequence[ibis_types.Value] = [],
 ) -> ibis_types.NumericValue:
     # Will be null if all inputs are null. Pandas defaults to zero sum though.
     bq_sum = _apply_window_if_present(column.sum(), window)
@@ -116,7 +124,10 @@ def _(
 @compile_unary_agg.register
 @numeric_op
 def _(
-    op: agg_ops.MedianOp, column: ibis_types.NumericColumn, window=None
+    op: agg_ops.MedianOp,
+    column: ibis_types.NumericColumn,
+    window=None,
+    agg_order_by: typing.Sequence[ibis_types.Value] = [],
 ) -> ibis_types.NumericValue:
     # PERCENTILE_CONT has very few allowed windows. For example, "window
     # framing clause is not allowed for analytic function percentile_cont".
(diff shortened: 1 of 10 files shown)


bigframes/core/compile/compiler.py
```diff
@@ -155,10 +155,14 @@ def compile_rowcount(node: nodes.RowCountNode, ordered: bool = True):
 
 @_compile_node.register
 def compile_aggregate(node: nodes.AggregateNode, ordered: bool = True):
-    result = compile_unordered_ir(node.child).aggregate(
-        node.aggregations, node.by_column_ids, node.dropna
-    )
-    return result if ordered else result.to_unordered()
+    if ordered:
```
TrevorBergeron, line 158: child should be compiled in ordered mode if and only if we need the ordering information for the purpose of ordered aggregations.

bigframes/core/compile/compiled.py
```diff
@@ -331,18 +335,18 @@ def aggregate(
...
-            # Ordering is irrelevant for single-row output, but set ordering id regardless as other ops(join etc.) expect it.
+            # Ordering is irrelevant for single-row output, but set ordering id regardless
+            # as other ops(join etc.) expect it.
             # TODO: Maybe can make completely empty
             ordering = ExpressionOrdering(
                 ordering_value_columns=tuple([]),
                 total_ordering_columns=frozenset([]),
             )
-            return OrderedIR(
+            return UnorderedIR(
```
TrevorBergeron: This should produce ordered IR , compiler can drop immediately if ordering not needed.

bigframes/core/compile/aggregate_compiler.py
```diff
@@ -416,7 +537,10 @@ def _(
 
 @compile_unary_agg.register
 def _(
-    op: agg_ops.AnyOp, column: ibis_types.Column, window=None
+    op: agg_ops.AnyOp,
+    column: ibis_types.Column,
+    window=None,
+    order_by: typing.Sequence[ibis_types.Value] = [],
```
TrevorBergeron, line 543: Hmm, this does clutter up the signature for just a single op. Not sure a better way though, maybe we will think of something later.

third_party/bigframes_vendored/ibis/expr/operations/reductions.py
```diff
@@ -18,6 +20,18 @@ class ApproximateMultiQuantile(Filterable, Reduction):
     dtype = dt.Array(dt.float64)
 
 
-__all__ = [
-    "ApproximateMultiQuantile",
-]
+class ArrayAggregate(Filterable, Reduction):
```
TrevorBergeron, line 23: Worried that our expansion of Ibis reduction semantics to add order_by might make this fragile to current or future ibis tree manipulations, since they will not account for our ordering clause.

bigframes/operations/aggregations.py
```diff
@@ -222,6 +226,24 @@ def output_type(self, *input_types: dtypes.ExpressionType) -> dtypes.ExpressionT
         ).output_type(input_types[0])
 
 
+@dataclasses.dataclass(frozen=True)
+class ArrayAggOp(UnaryAggregateOp):
+    name: ClassVar[str] = "arrayagg"
+
+    @property
+    def can_order_by(self):
+        return True
```
TrevorBergeron, line 235: its a more of a "must" than "can" right? At least for the purposes of non-determinism

bigframes/core/compile/compiled.py
chelsea-lin: Fixed. Done

bigframes/operations/aggregations.py
chelsea-lin: It's "must" for ordered mode but "can" for unordered mode. The unordered mode will introduce non-determinism.

third_party/bigframes_vendored/ibis/expr/operations/reductions.py
chelsea-lin: @tswast Do you have any thoughts about it?

bigframes/core/compile/aggregate_compiler.py
chelsea-lin: Yeah, `order_by` is overwhelm to apply for all aggregate operations, while only `array_agg`, `string_agg` and `array_concat_agg` supports `order_by`.

bigframes/core/compile/compiler.py
chelsea-lin: Fixed! Thanks!

bigframes/core/compile/compiled.py
```diff
@@ -604,6 +610,60 @@ def reversed(self) -> OrderedIR:
...
+        self,
+        aggregations: typing.Sequence[typing.Tuple[ex.Aggregation, str]],
+        by_column_ids: typing.Sequence[str] = (),
+        dropna: bool = True,
+    ) -> OrderedIR:
+        """
+        Apply aggregations to the expression.
+        Arguments:
+            aggregations: input_column_id, operation, output_column_id tuples
+            by_column_id: column id of the aggregation key, this is preserved through
```
tswast, line 623:
```suggestion
            by_column_ids: column ids of the aggregation key, this is preserved through
```

bigframes/core/compile/compiled.py
```diff
@@ -604,6 +610,60 @@ def reversed(self) -> OrderedIR:
         expr_builder.ordering = self._ordering.with_reverse()
         return expr_builder.build()
 
+    def aggregate(
```
tswast, line 613: @chelsea-lin Can you help me understand more about where this came from? Presumably we already support aggregates, right, so I'm wondering:

1. what this does that the existing code paths can't handle and
2. whether those existing code paths are now redundant and should be removed / migrated to this

bigframes/core/compile/compiler.py
```diff
@@ -155,10 +155,18 @@ def compile_rowcount(node: nodes.RowCountNode, ordered: bool = True):
 
 @_compile_node.register
 def compile_aggregate(node: nodes.AggregateNode, ordered: bool = True):
-    result = compile_unordered_ir(node.child).aggregate(
-        node.aggregations, node.by_column_ids, node.dropna
+    has_orderred_aggregation_ops = any(
```
tswast, line 158: Typo: should be `ordered`

third_party/bigframes_vendored/ibis/backends/bigquery/registry.py
```diff
@@ -38,13 +38,25 @@ def _quantile(translator, op: ibis_reductions.Quantile):
     return f"PERCENTILE_CONT({arg}, {quantile})"
 
 
+def _array_aggregate(translator, op: vendored_ibis_ops.ArrayAggregate):
```
tswast, line 41: Do we have an issue upstream for this? If so, please add a comment with a link so that we can remove this once its resolved. If not, please file one.

third_party/bigframes_vendored/ibis/expr/operations/reductions.py
tswast: Since we're using a totally new node, this seems safe enough to me, but we should make sure we're coordinating with upstream.

bigframes/core/compile/compiled.py
chelsea-lin: The `array_agg` is the first aggregated method, which requires `OrderedIR` so that a new `aggregate()` method is added to `OrderedIR`. Good catch! I just remove the redundant codes in the new commit.

bigframes/core/compile/compiled.py
```diff
@@ -163,6 +169,54 @@ def get_column_type(self, key: str) -> bigframes.dtypes.Dtype:
             bigframes.dtypes.ibis_dtype_to_bigframes_dtype(ibis_type),
         )
 
+    def _aggregate_helper(
```
tswast, line 172: Could we get a better name for this method, please. "helper" is very generic, so it's hard to understand what this method is doing.

tswast (commented):
A nit regarding the helper function naming, but otherwise looks good!

bigframes/core/compile/compiled.py
chelsea-lin: Renamed it into `_aggergate_base` to match the object name `BaseIbisIR`.

bigframes/core/compile/aggregate_compiler.py
```diff
@@ -202,44 +226,102 @@ def _(
...
 
 
 @compile_unary_agg.register
-def _(op: agg_ops.CutOp, x: ibis_types.Column, window=None):
+def _(
+    op: agg_ops.ArrayAggOp,
+    column: ibis_types.Column,
+    window=None,
+    order_by: typing.Sequence[ibis_types.Value] = [],
+) -> ibis_types.ArrayValue:
```
TrevorBergeron, line 299: Instead of registering as a UnaryAggOp, and having to modify every other operation, should ArrayAgg be its own aggregate expression class, maybe OrderedUnaryAggOp?

TrevorBergeron (approved):

bigframes/core/compile/aggregate_compiler.py
chelsea-lin: To avoid modifying other operation, creating a new `compile_ordered_unary_agg` dispatcher and let `can_order_by` property of the `UnaryAggOp` decides the dispatcher.

chelsea-lin:
The end-to-end tests that failed are not caused by this particular change.

PR #679, opened 2024-05-10
docs: add code snippets for imported tensorflow model

ashleyxuu:

Fixes internal #338878382 🦕

### DIFF
--- a/samples/snippets/imported_tensorflow_model_test.py
+++ b/samples/snippets/imported_tensorflow_model_test.py
@@ -1,0 +1,42 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (t
+# you may not use this file except in compliance wi
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in
+# distributed under the License is distributed on a
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, eit
+# See the License for the specific language governi
+# limitations under the License.
+
+
+def test_imported_tensorflow_model() -> None:
+    # Determine project id, in this case prefer the one set in the environment
+    # variable GOOGLE_CLOUD_PROJECT (if any)
+    import os
+
+    PROJECT_ID = os.getenv("GOOGLE_CLOUD_PROJECT", "bigframes-dev")
+    LOCATION = "US"
+    MODEL_PATH = "gs://cloud-training-demos/txtclass/export/exporter/1549825580/*"
+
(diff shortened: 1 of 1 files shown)


samples/snippets/imported_tensorflow_model_test.py
```diff
@@ -0,0 +1,42 @@
...
+    PROJECT_ID = os.getenv("GOOGLE_CLOUD_PROJECT", "bigframes-dev")
+    LOCATION = "US"
+    MODEL_PATH = "gs://cloud-training-demos/txtclass/export/exporter/1549825580/*"
+
+    # [START bigquery_dataframes_imported_tensorflow_tutorial_import_tensorflow_models]
+    import bigframes
+    from bigframes.ml.imported import TensorFlowModel
+
+    bigframes.options.bigquery.project = PROJECT_ID
+    bigframes.options.bigquery.location = LOCATION
```
tswast, line 30: No need for this to be a constant. Our goal is to be as copy-pasteable as possible and this adds a barrier to that.

Just say `"US"` and add a comment saying they can change it to one of the valid locations (link to BQ locations).

samples/snippets/imported_tensorflow_model_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+    imported_tensorflow_model = TensorFlowModel(model_path=MODEL_PATH)
```
tswast, line 32: Please add a comment showing an example string. I want them to see that they need the important information, like `gs://`, the bucket name, and the `*` wildcard. `model_path=MODEL_PATH,  # e.g. "gs://bucket-name/path/to/model/*"``

samples/snippets/imported_tensorflow_model_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def test_imported_tensorflow_model() -> None:
+    # Determine project id, in this case prefer the one set in the environment
+    # variable GOOGLE_CLOUD_PROJECT (if any)
+    import os
+
+    PROJECT_ID = os.getenv("GOOGLE_CLOUD_PROJECT", "bigframes-dev")
+    LOCATION = "US"
+    MODEL_PATH = "gs://cloud-training-demos/txtclass/export/exporter/1549825580/*"
```
tswast, line 23: Is this a public dataset? If so, let's just include it in the code sample itself.

tswast (changes requested):

samples/snippets/imported_tensorflow_model_test.py
ashleyxuu: I don't think it is a public dataset, I put `"gs://bucket/path/to/saved_model/*"` in the comment

tswast (approved):

PR #680, opened 2024-05-10
feat: Add Series.combine

TrevorBergeron:

### DIFF
--- a/third_party/bigframes_vendored/pandas/core/series.py
+++ b/third_party/bigframes_vendored/pandas/core/series.py
@@ -1279,6 +1279,62 @@ class Series(NDFrame):  # type: ignore[misc]
         """
         raise NotImplementedError(constants.ABSTRACT_METHOD_ERROR_MESSAGE)
 
+    def combine(
+        self,
+        other: Series | Hashable,
+        func,
+    ) -> Series:
+        """
+        Combine the Series with a Series or scalar according to `func`.
+
+        Combine the Series and `other` using `func` to perform elementwise
+        selection for combined Series.
+        `fill_value` is assumed when value is missing at some index
+        from one of the two objects being combined.
+
+        **Examples:**
+
+            >>> import bigframes.pandas as bpd
+            >>> import numpy as np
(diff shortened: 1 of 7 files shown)


third_party/bigframes_vendored/pandas/core/series.py
```diff
@@ -1279,6 +1279,62 @@ def apply(
...
+            >>> s2
+            falcon    345.0
+            eagle     200.0
+            duck       30.0
+            dtype: Float64
+
+            Now, to combine the two datasets and view the highest speeds
+            of the birds across the two datasets
+
+            >>> s1.combine(s2, np.maximum)
```
chelsea-lin, line 1319: The order of the outputs are different as pandas. How we decide the ordering of the output?

third_party/bigframes_vendored/pandas/core/series.py
```diff
@@ -1279,6 +1279,62 @@ def apply(
...
+            falcon    345.0
+            eagle     200.0
+            duck       <NA>
+            dtype: Float64
+
+        Args:
+            other (Series or scalar):
+                The value(s) to be combined with the `Series`.
+            func (function):
+                BigFrames DataFrames ``remote_function`` to apply.
```
chelsea-lin, line 1329: Is the `np.maxinum` a remote function in the example?

chelsea-lin (approved):
LTGM

third_party/bigframes_vendored/pandas/core/series.py
TrevorBergeron: Pandas will check the index of both sides, if they are not exactly the same, it will sort the index. If we wanted to copy this behavior, we would need to download the index from both sides of the operation, which I don't think is worth it. Users can always call `sort_index()` after combining if needed.

third_party/bigframes_vendored/pandas/core/series.py
TrevorBergeron: Its a numpy ufunc. Having remote functions in the doctests has been problematic so far.

PR #681, opened 2024-05-10
fix: remove a usage of the `resource` package when not available, such as on Windows

tswast:
Towards internal issue 339769193.

Note: I'm still working on getting a Windows Python dev environment setup, but this should address at least one known problem. I aim to follow-up with automated tests on Windows, too.

🦕


tswast:
I verified manually that `import bigframes.pandas` now works on Windows. I'm working on automated testing to make sure all functionality works and we don't regress.

TrevorBergeron (approved):

ashleyxuu (approved):

PR #689, opened 2024-05-14
fix: `read_gbq_table` respects primary keys even when `filters` are set

tswast:
Closes internal issues 338039517 (primary key inconsistency), 338037499 (LIMIT for max_results), 340540991 (avoid running query immediately if time travel is supported), 337925142 (push down column filters to when we create the time travel subquery).

feat: `read_gbq_query` supports `filters`
perf: use a `LIMIT` clause when `max_results` is set
perf: don't run query immediately from `read_gbq_table` if `filters` is set

🦕

### DIFF
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -339,19 +339,6 @@ class Session(
         elif col_order:
             columns = col_order
 
-        filters = list(filters)
-        if len(filters) != 0 or bf_io_bigquery.is_table_with_wildcard_suffix(
-            query_or_table
-        ):
-            # TODO(b/338111344): This appears to be missing index_cols, which
-            # are necessary to be selected.
-            # TODO(b/338039517): Refactor this to be called inside both
-            # _read_gbq_query and _read_gbq_table (after detecting primary keys)
-            # so we can make sure index_col/index_cols reflects primary keys.
-            query_or_table = bf_io_bigquery.to_query(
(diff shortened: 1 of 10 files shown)


TrevorBergeron (commented):
mostly looks good, seems we've lost some validation on unknown column names though?

bigframes/session/__init__.py
```diff
@@ -690,16 +686,67 @@ def _read_gbq_table(
...
-            time_travel_timestamp,
+            ibis_client=self.ibis_client,
+            table_ref=table_ref,
+            index_cols=index_cols,
+            columns=columns,
+            filters=filters,
+            time_travel_timestamp=time_travel_timestamp,
         )
 
         for key in columns:
```
tswast, line 752: @TrevorBergeron Which "validation for unknown column names'" are we missing? I still see this here.

tswast: Would you like me to move it to before we create the ibis expression? It's possible ibis will fail with a dry run error when we try to access the columns in it now.

tswast: I see `tests/system/small/test_session.py::test_read_gbq_w_columns[unknown_col]` failure now. Likely this is the issue. I'll move this validation to after we tech the table metadata.

tswast: Fixed in 87365db9d494cdc6cfd18ba930dea2b788c591de

TrevorBergeron (approved):

PR #700, opened 2024-05-17
chore: type error in test_ensemble

ashleyxuu:

Fixes internal #340888645🦕


tests/system/large/ml/test_ensemble.py
```diff
@@ -46,10 +46,10 @@ def test_xgbregressor_default_params(penguins_df_default_index, dataset_id):
     reloaded_model = model.to_gbq(
         f"{dataset_id}.temp_configured_xgbregressor_model", replace=True
     )
-    # TODO(b/340888645): fix type error
+    assert reloaded_model._bqml_model is not None
```
tswast, line 49: Might need to save this to a temp variable, since it's theoretically possible `reloaded_model._bqml_model` changes between here and the assertion, but that does not look possible in this case.

tswast (approved):
Thanks!

PR #703, opened 2024-05-17
feat: bigframes.streaming module for continuous queries

milkshakeiii:
This PR adds to_bigtable and a test. to_pubsub will be a next step

Also adds a script for creating the bigtables used by the test.


bigframes/streaming/__init__.py
```diff
@@ -0,0 +1,150 @@
...
+    # get default client if not passed
+    if bq_client is None:
+        bq_client = bigframes.get_global_session().bqclient
+
+    # build export string from parameters
+    app_profile_url_string = ""
+    if app_profile is not None:
+        app_profile_url_string = f"appProfiles/{app_profile}/"
+
+    truncate_string = "FALSE"
```
GarrettWu, line 102: You can just pass in "truncate" and other boolean values as SQL doesn't distinguish upper case and lower case. Which also improves readability.

tests/system/large/test_streaming.py
```diff
@@ -0,0 +1,45 @@
...
+import bigframes.streaming
+
+
+def test_streaming_to_bigtable():
+    # launch a continuous query
+    sql = """SELECT
+        body_mass_g, island as rowkey
+        FROM birds.penguins"""
+    query_job = bigframes.streaming.to_bigtable(
+        sql,
```
GarrettWu, line 26: Is there a way that we can check these parameters are kept? If possible may worth adding tests to check.

bigframes/streaming/__init__.py
```diff
@@ -0,0 +1,150 @@
+# Copyright 2023 Google LLC
```
tswast, line 1: This can safely be 2024, since we haven't published any of this code in 2023.

bigframes/streaming/__init__.py
milkshakeiii: Good point, thanks, fixed.

tests/system/large/test_streaming.py
milkshakeiii: For job_id, yes, added it. For the other parameters, they pass through to bigtable or the client library and aren't accessible on the output QueryJob object, so there isn't much we can do to check them, unfortunately. But fortunately they just pass through, and our documentation points to the parameters they map to, so we wouldn't necessarily need to test that anyway.

bigframes/streaming/__init__.py
GarrettWu: The same for "auto_create_column_families_string" :)

milkshakeiii: Oops, fixed, thanks.

GarrettWu (approved):

PR #709, opened 2024-05-20
refactor: New read node that defers ibis table instantiation

TrevorBergeron:

### DIFF
--- a/bigframes/core/compile/compiler.py
+++ b/bigframes/core/compile/compiler.py
@@ -29,6 +29,7 @@ import bigframes.core.compile.schema_translator
 import bigframes.core.compile.single_column
 import bigframes.core.nodes as nodes
 import bigframes.core.ordering as bf_ordering
+import bigframes.dtypes
 
 if typing.TYPE_CHECKING:
     import bigframes.core
@@ -99,8 +100,9 @@ def compile_readtable(node: nodes.ReadTableNode, ordered: bool = True):
     full_table_name = f"{node.project_id}.{node.dataset_id}.{node.table_id}"
     import bigframes.core.compile.schema_translator
 
-    ibis_schema = bigframes.core.compile.schema_translator.convert_bf_schema(
-        node.schema
+    # Physical schema might include unused columns, unsupported datatypes like JSON
+    physical_schema = ibis.backends.bigquery.BigQuerySchema.to_ibis(
+        list(node.physical_schema)
     )
     if node.snapshot_time is not None or node.sql_predicate is not None:
         import bigframes.session._io.bigquery
@@ -114,9 +116,11 @@ def compile_readtable(node: nodes.ReadTableNode, ordered: bool = True):
             index_cols=(),
             max_results=None,
         )
-        ibis_table = ibis.backends.bigquery.Backend().sql(schema=ibis_schema, query=sql)
+        ibis_table = ibis.backends.bigquery.Backend().sql(
+            schema=physical_schema, query=sql
+        )
     else:
-        ibis_table = ibis.table(ibis_schema, full_table_name)
+        ibis_table = ibis.table(physical_schema, full_table_name)
 
     if ordered:
         if node.primary_key:
@@ -146,14 +150,20 @@ def compile_readtable(node: nodes.ReadTableNode, ordered: bool = True):
             )
         return compiled.OrderedIR(
             ibis_table,
-            columns=tuple(ibis_table[col] for col in node.schema.names),
+            columns=tuple(
+                bigframes.dtypes.ibis_value_to_canonical_type(ibis_table[col])
+                for col in node.schema.names
+            ),
             ordering=ordering,
             hidden_ordering_columns=hidden_columns,
         )
     else:
         return compiled.UnorderedIR(
             ibis_table,
-            tuple(ibis_table[col] for col in node.schema.names),
+            tuple(
+                bigframes.dtypes.ibis_value_to_canonical_type(ibis_table[col])
+                for col in node.schema.names
+            ),
         )
 
 
--- a/bigframes/core/compile/compiler.py
+++ b/bigframes/core/compile/compiler.py
@@ -17,12 +17,18 @@ import functools
 import io
 import typing
 
+import ibis
+import ibis.backends
+import ibis.backends.bigquery
 import pandas as pd
 
 import bigframes.core.compile.compiled as compiled
 import bigframes.core.compile.concat as concat_impl
+import bigframes.core.compile.default_ordering as default_ordering
+import bigframes.core.compile.schema_translator
 import bigframes.core.compile.single_column
 import bigframes.core.nodes as nodes
+import bigframes.core.ordering as bf_ordering
 
 if typing.TYPE_CHECKING:
     import bigframes.core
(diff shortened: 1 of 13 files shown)


bigframes/core/compile/compiler.py
```diff
@@ -88,6 +95,80 @@ def compile_readlocal(node: nodes.ReadLocalNode, ordered: bool = True):
...
+            max_results=None,
+        )
+        ibis_table = ibis.backends.bigquery.Backend().sql(
+            schema=physical_schema, query=sql
+        )
+    else:
+        ibis_table = ibis.table(physical_schema, full_table_name)
+
+    if ordered:
+        if node.primary_key:
```
tswast, line 128: Seems a bit odd to me to put ordering generation here, but I guess this is just for total ordering, right? We still generate separate order by when we add the index, right?

bigframes/core/compile/compiler.py
```diff
@@ -88,6 +95,80 @@ def compile_readlocal(node: nodes.ReadLocalNode, ordered: bool = True):
...
+            ordering_value_columns = tuple(
+                bf_ordering.ascending_over(col) for col in node.primary_key
+            )
+            if node.primary_key_sequential:
```
tswast, line 132: Where do we have primary keys that we know are sequential integers?

bigframes/core/compile/compiler.py
```diff
@@ -88,6 +95,80 @@ def compile_readlocal(node: nodes.ReadLocalNode, ordered: bool = True):
...
+                )
+            else:
+                integer_encoding = bf_ordering.IntegerEncoding()
+            ordering = bf_ordering.ExpressionOrdering(
+                ordering_value_columns,
+                integer_encoding=integer_encoding,
+                total_ordering_columns=frozenset(node.primary_key),
+            )
+            hidden_columns = ()
+        else:
```
tswast, line 144: This is getting a bit long for me to track indentation-wise. Could we make some helper functions, please?

tswast (commented):
Nowhere near finished reviewing, but sending some early comments so it doesn't get stuck for too long.

bigframes/core/nodes.py
```diff
@@ -345,6 +348,67 @@ def transform_children(
...
+    project_id: str = field()
+    dataset_id: str = field()
+    table_id: str = field()
+
+    physical_schema: Tuple[bq.SchemaField, ...] = field()
+    # Subset of physical schema columns, with chosen BQ types
+    columns: schemata.ArraySchema = field()
+
+    table_session: bigframes.session.Session = field()
+    # Should this even be stored here?
```
tswast, line 363: Would "native ordering column" or something be more appropriate? Such a name might allow us to use a row ID pseudocolumn as a fallback if one becomes available.

bigframes/core/nodes.py
```diff
@@ -345,6 +348,67 @@ def transform_children(
...
+    # Empty tuple if no primary key (primary key can be any set of columns that together form a unique key)
+    # Empty if no known unique key
+    primary_key: Tuple[str, ...] = field()  # subset of schema
+    # indicates a primary key that is exactly offsets 0, 1, 2, ..., N-2, N-1
```
tswast, line 367: I'm still curious when we know this will be the case. Maybe when we cache an ordered object?

bigframes/core/nodes.py
```diff
@@ -345,6 +348,67 @@ def transform_children(
...
+    primary_key_sequential: bool = False
+    snapshot_time: typing.Optional[datetime.datetime] = None
```
tswast, line 369: Technically "time travel" which is different from snapshot in BQ. https://cloud.google.com/bigquery/docs/access-historical-data

Although looking at that, even our the backend messages conflate the two.

bigframes/core/nodes.py
```diff
@@ -345,6 +348,67 @@ def transform_children(
...
+    # Added for backwards compatibility, not validated
+    sql_predicate: typing.Optional[str] = None
```
tswast, line 371: Fascinating. This implies some level of SQL compilation outside of this node. Should this be a structured "filters" object, instead?

bigframes/core/nodes.py
```diff
@@ -345,6 +348,67 @@ def transform_children(
...
+
+    def __post_init__(self):
+        # enforce invariants
+        physical_names = set(map(lambda i: i.name, self.physical_schema))
+        assert len(self.columns.names) > 0
```
tswast, line 376: Why this assertion? It is possible to create a completely empty table in BQ. Why one would want to do so, I'm not certain, but it is possible.

bigframes/core/nodes.py
```diff
@@ -345,6 +348,67 @@ def transform_children(
...
+        assert set(self.primary_key).issubset(physical_names)
```
tswast, line 377: This assertion might be false in future if "primary key" contains psuedo columns.

bigframes/core/nodes.py
```diff
@@ -345,6 +348,67 @@ def transform_children(
...
+        assert set(self.columns.names).issubset(physical_names)
```
tswast, line 378: If we ever reach this line of code, it would likely be an indication we should have a ValueError further up the call stack. Would at least be helpful to have a custom error message here as in other assertions so the user know to file a bug that we missed a validation check somewhere.

bigframes/dtypes.py
```diff
@@ -280,6 +281,10 @@ def ibis_dtype_to_bigframes_dtype(
     if isinstance(ibis_dtype, ibis_dtypes.Integer):
         return pd.Int64Dtype()
 
+    # Temporary: Will eventually support an explicit json type instead of casting to string.
```
tswast, line 284: We should probably raise a warning (PreviewWarning?) in this case to make sure folks know that depending on any JSON functionality may break in future.

bigframes/session/_io/bigquery/__init__.py
```diff
@@ -210,6 +210,7 @@ def start_query_with_client(
         )
 
     try:
+        print(sql)
```
tswast, line 213: Remove this `print`.

bigframes/session/__init__.py
```diff
@@ -793,19 +797,26 @@ def _read_gbq_table(
...
             columns=columns,
-            filters=filters,
+            sql_predicate=bf_io_bigquery.compile_filters(filters) if filters else None,
             time_travel_timestamp=time_travel_timestamp,
+            # If we've made it this far, we know we don't have any
+            # max_results to worry about, because in that case we will
+            # have executed a query with a LIMIT clause.
+            max_results=None,
         )
+        bf_read_gbq_table.validate_sql_through_ibis(sql, self.ibis_client)
```
tswast, line 819: We'll need Henry's logic to dryrun with and without the `time_travel_timestamp` so we can continue to support tables that don't support time travel.

tswast (changes requested):

bigframes/core/nodes.py
TrevorBergeron: Removed this constraint, though we might need a bit more work to support pseudo columns anyways.

bigframes/core/nodes.py
TrevorBergeron: Added some error messages

bigframes/core/nodes.py
TrevorBergeron: Yeah, I guess we should allow empty tables, removed this constraint

bigframes/core/nodes.py
TrevorBergeron: renamed to `total_order_cols`

bigframes/dtypes.py
TrevorBergeron: added a preview warning

bigframes/core/nodes.py
tswast: Hmm... Forcing compilation to a string doesn't seem like the right choice to me. Some namedtuple or frozen dataclass would make the most sense to me.

tswast: > If eq and frozen are both true, by default @dataclass will generate a __hash__() method for you.

https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass

tswast (approved):

PR #712, opened 2024-05-21
fix: warn and disable time travel for linked datasets

milkshakeiii:


bigframes/session/__init__.py
```diff
@@ -758,9 +759,31 @@ def _read_gbq_table(
...
+            time_travel_timestamp=time_travel_timestamp,
+            max_results=None,
+        )
+        dry_run_config = bigquery.QueryJobConfig()
+        dry_run_config.dry_run = True
+        try:
+            self._start_query(sql, job_config=dry_run_config)
+        except google.api_core.exceptions.NotFound:
+            time_travel_timestamp = None
+            warnings.warn(
```
tswast, line 780: Let's use a custom category here from `bigframes.exceptions` so folks can filter this out in pipelines where they know they are fine with a query on a table that could be changing from under them.

tswast (approved):

PR #723, opened 2024-05-24
fix: correct index labels in multiple aggregations for DataFrameGroupBy

chelsea-lin:
- [X] Make sure to open an issue as a bug

Fixes internal issue 341157901 🦕

### DIFF
--- a/bigframes/core/groupby/__init__.py
+++ b/bigframes/core/groupby/__init__.py
@@ -340,19 +340,29 @@ class DataFrameGroupBy(vendored_pandas_groupby.DataFrameGroupBy):
             for f in func
         ]
 
-        aggregated_columns = pd.MultiIndex.from_tuples(
-            [
-                self._block.col_id_to_label[col_id]
-                for col_id in self._aggregated_columns()
-            ],
-            names=[*self._block.column_labels.names],
-        ).to_frame(index=False)
+        if self._block.column_labels.nlevels > 1:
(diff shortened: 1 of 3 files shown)


bigframes/core/groupby/__init__.py
```diff
@@ -326,7 +326,9 @@ def _agg_list(self, func: typing.Sequence) -> df.DataFrame:
             for f in func
         ]
         column_labels = [
-            (col_id, f) for col_id in self._aggregated_columns() for f in func
+            (self._block.col_id_to_label[col_id], f)
+            for col_id in self._aggregated_columns()
+            for f in func
         ]
```
TrevorBergeron, line 332: Can you also test with column multi-index? Not sure this will work for that case

chelsea-lin: It works: https://screenshot.googleplex.com/BPBX6WpPFuXTuHs

TrevorBergeron: Not quite, we should have a case where the original columns have multi-part names

TrevorBergeron (approved):

bigframes/core/groupby/__init__.py
chelsea-lin: Good catch! I've corrected the handling of multi-indexes when applying multiple aggregation functions.

PR #727, opened 2024-05-28
feat: Add Series.peek to preview data efficiently

TrevorBergeron:


bigframes/core/pruning.py
```diff
@@ -0,0 +1,66 @@
...
+        ops.ge_op,
+        ops.lt_op,
+        ops.le_op,
+    )
+)
+
+
+def cluster_cols_for_predicate(predicate: ex.Expression) -> Sequence[str]:
+    """Try to determine cluster col candidates that work with given predicates."""
+    # TODO: Prioritize equality predicates over ranges
```
tswast, line 36: Could you rephrase? It took me a while to understand that you meant.

Perhaps add that since equality is narrower filter, it's more likely to reduce the data read if it's the first clustering filter.

Maybe this TODO should be a sort by how selective the predicates are?

bigframes/core/pruning.py
```diff
@@ -0,0 +1,66 @@
...
+
+from typing import Sequence
+
+import bigframes.core.expression as ex
+import bigframes.operations as ops
+
+COMPARISON_OP_TYPES = tuple(
+    type(i)
+    for i in (
+        ops.eq_op,
```
tswast, line 23: I'm curious why we're enumerating the instances of these classes instead of listing out the classes directly?

bigframes/core/pruning.py
```diff
@@ -0,0 +1,66 @@
...
+    else:
+        # Constant
+        return []
+
+
+def cluster_cols_for_comparison(
+    left_ex: ex.Expression, right_ex: ex.Expression
+) -> Sequence[str]:
+    if left_ex.is_const:
+        # There are some invertible ops that would also be ok
```
tswast, line 60: Should this be a TODO?

I'm not quite understanding this comment either. Why would the be OK? I assume "OK" means, "can extract a column to use as a clustering column"?

bigframes/core/tree_properties.py
```diff
@@ -91,6 +91,30 @@ def _node_counts_inner(
     )
 
 
+def count_nodes(forest: Sequence[nodes.BigFrameNode]) -> dict[nodes.BigFrameNode, int]:
```
tswast, line 94: Would love a docstring here.

bigframes/operations/__init__.py
```diff
@@ -60,6 +60,11 @@ def order_preserving(self) -> bool:
         """Whether the row operation preserves total ordering. Can be pruned from ordering expressions."""
         return False
 
+    @property
+    def pruning_compatible(self) -> bool:
+        """Whether the operation preserves locality o"""
```
tswast, line 65: I see a hanging "o". Was that meant to be `"or ..."`?

tswast: I'd also like some more information for help in determining when an operation would be pruning compatible.

bigframes/series.py
```diff
@@ -617,6 +617,39 @@ def head(self, n: int = 5) -> Series:
     def tail(self, n: int = 5) -> Series:
         return typing.cast(Series, self.iloc[-n:])
 
+    def peek(self, n: int = 5, *, force: bool = True) -> pandas.DataFrame:
+        """
+        Preview n arbitrary elements from the series. No guarantees about row selection or ordering.
+        ``Series.peek(force=False)`` will always be very fast, but will not succeed if data requires
```
tswast, line 623: Let's try to keep the first line summary short.

```suggestion
        Preview n arbitrary elements from the series without guarantees about row selection or ordering.

        ``Series.peek(force=False)`` will always be very fast, but will not succeed if data requires
```

bigframes/session/__init__.py
```diff
@@ -332,13 +333,15 @@ def session_id(self):
     @property
     def objects(
         self,
-    ) -> collections.abc.Set[
+    ) -> Tuple[
```
tswast, line 336: Technically a breaking change. Maybe OK since we didn't actually document this property, but might be better to change from `Set` to a broader type like `Iterable`.

tswast (changes requested):
Could we split out the session-aware caching to a separate PR or is this pretty tightly coupled to `Series.peek()`?

bigframes/session/__init__.py
TrevorBergeron: Hmm, yeah, actually, should we just make this private? Added this very recently and wasn't really intending this for user consumption.

bigframes/series.py
TrevorBergeron: done

bigframes/operations/__init__.py
TrevorBergeron: Actually didn't end up using this. Removed from new revision. Later on will add some concept of "inverse" operation to help normalize predicates. For now though, only considering range and equality predicates between column and a constant.

bigframes/core/tree_properties.py
TrevorBergeron: added

bigframes/core/pruning.py
TrevorBergeron: Updated to a todo, with some examples of how conditions can be normalized.

bigframes/core/pruning.py
TrevorBergeron: Unfortunately, I don't actually have a direct handle on the actual classes, as these ops are created programmatically.

bigframes/core/pruning.py
TrevorBergeron: yeah, the idea is to cluster on the filter predicted to be most selective. update the todo

TrevorBergeron:
> Could we split out the session-aware caching to a separate PR or is this pretty tightly coupled to `Series.peek()`?

I worry that implementing `Series.peek()` without session-aware caching will result in a lot of redundant re-computation if users peek at multiple columns of the same dataframe. I could start with session-aware caching and apply it to dataframe.peek instead I guess.

bigframes/series.py
```diff
@@ -617,6 +617,40 @@ def head(self, n: int = 5) -> Series:
...
+            if force:
+                self._cached()
+                maybe_result = self._block.try_peek(n, force=True)
+                assert maybe_result is not None
+            else:
+                raise ValueError(
+                    "Cannot peek efficiently when data has aggregates, joins or window functions applied. Use force=True to fully compute dataframe."
+                )
+        as_series = maybe_result.squeeze(axis=1)
+        as_series.name = self.name
```
tswast, line 651: Do we need to do anything different here to support multi-index?

bigframes/series.py
```diff
@@ -617,6 +617,40 @@ def head(self, n: int = 5) -> Series:
     def tail(self, n: int = 5) -> Series:
         return typing.cast(Series, self.iloc[-n:])
 
+    def peek(self, n: int = 5, *, force: bool = True) -> pandas.DataFrame:
+        """
+        Preview n arbitrary elements from the series without guarantees about row selection or ordering.
+
+        ``Series.peek(force=False)`` will always be very fast, but will not succeed if data requires
+        full data scanning. Using ``force=True`` will always succeed, but may be perform queries.
+        Query results will be cached so that future steps will benefit from these queries.
```
tswast, line 626: Do we need a caveat here that caching is session-aware and will attempt to cache the optimal subtree? (Not sure exactly how to phrase that in a friendlier way.)

bigframes/session/__init__.py
```diff
@@ -1918,6 +1921,18 @@ def _cache_with_offsets(self, array_value: core.ArrayValue):
         ).node
         self._cached_executions[array_value.node] = cached_replacement
 
+    def _session_aware_caching(self, array_value: core.ArrayValue) -> None:
```
tswast, line 1924: Let's verbify this.

```suggestion
    def _cache_with_session_awareness(self, array_value: core.ArrayValue) -> None:
```

tests/system/small/test_series.py
```diff
@@ -1970,6 +1970,42 @@ def test_head_then_series_operation(scalars_dfs):
     )
 
 
+def test_series_peek(scalars_dfs):
```
tswast, line 1973: Let's add a multi-index test.

tests/unit/test_planner.py
```diff
@@ -0,0 +1,78 @@
...
+LEAF: core.ArrayValue = core.ArrayValue.from_ibis(
+    session=None,  # type: ignore
+    table=ibis_table,
+    columns=cols,
+    hidden_ordering_columns=[],
+    ordering=bf_order.ExpressionOrdering((bf_order.ascending_over("col_a"),)),
+)
+
+
+def test_session_aware_caching_project_filter():
```
tswast, line 36: Could we rephrase and/or add a docstring explaining the expected behavior the test is trying to verify?

For example:

```
"""
Test that if a node is filtered by a column, the node is cached pre-filter and clustered by the filter column.
"""
```

tests/unit/test_planner.py
```diff
@@ -0,0 +1,78 @@
...
+        ops.gt_op.as_expr("col_a", ex.const(3))
+    )
+    result, cluster_col = planner.session_aware_cache_plan(
+        target.node, [obj.node for obj in session_objects]
+    )
+    assert result == LEAF.node
+    assert cluster_col == "col_a"
+
+
+def test_session_aware_caching_unusable_filter():
```
tswast, line 48: Likewise

"""
Test that if a node is filtered by multiple columns in the same comparison, the node is cached pre-filter and not clustered by either column.

(Also explain a bit more why such a filter is unusable for clustering)
"""

tests/unit/test_planner.py
```diff
@@ -0,0 +1,78 @@
...
+        ops.gt_op.as_expr("col_a", "col_b")
+    )
+    result, cluster_col = planner.session_aware_cache_plan(
+        target.node, [obj.node for obj in session_objects]
+    )
+    assert result == LEAF.node
+    assert cluster_col is None
+
+
+def test_session_aware_caching_fork_after_window_op():
```
tswast, line 60: Likewise

"""
Test that caching happens *after* an windowed operation.

(Also, explain a bit more why this is important to do)
"""

I see that a filter and clustering is also present in this test. Is that necessary for what you want to test? If so, please also describe in the docstring.

bigframes/session/planner.py
```diff
@@ -0,0 +1,62 @@
...
+import bigframes.core.tree_properties as traversals
+
+
+def session_aware_cache_plan(
+    root: nodes.BigFrameNode, session_forest: Sequence[nodes.BigFrameNode]
+) -> Tuple[nodes.BigFrameNode, Optional[str]]:
+    """
+    Determines the best node to cache given a target and a list of object roots for objects in a session.
+
+    Returns the node to cache, and optionally a clustering column.
```
tswast, line 31: Would there be benefit in returning more than just one clustering columns? We can support up to 4: https://cloud.google.com/bigquery/docs/creating-clustered-tables#create_an_empty_clustered_table_with_a_schema_definition

Also, do we need to add a filter based on clusterable types? Or is the set of types that can be used in a filter the same as those that are clusterable?

bigframes/core/pruning.py
```diff
@@ -0,0 +1,68 @@
...
+def cluster_cols_for_predicate(predicate: ex.Expression) -> Sequence[str]:
+    """Try to determine cluster col candidates that work with given predicates."""
+    # TODO: Prioritize based on predicted selectivity (eg. equality conditions are probably very selective)
+    if isinstance(predicate, ex.UnboundVariableExpression):
+        return [predicate.id]
+    if isinstance(predicate, ex.OpExpression):
+        op = predicate.op
+        if isinstance(op, COMPARISON_OP_TYPES):
+            return cluster_cols_for_comparison(predicate.inputs[0], predicate.inputs[1])
+        if isinstance(op, (type(ops.invert_op))):
```
tswast, line 43: Let's add a TODO for geo, too. Looks like functions like `st_dwithin` can take advantage of clustering on geo columns. https://cloud.google.com/blog/products/data-analytics/best-practices-for-spatial-clustering-in-bigquery?e=48754805

bigframes/session/planner.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    node_counts = traversals.count_nodes(session_forest)
+    # These node types are cheap to re-compute
```
tswast, line 34: Let's complete the thought in this comment for clarity.

```suggestion
    # These node types are cheap to re-compute, so it makes more sense to cache their children.
```

bigframes/session/planner.py
```diff
@@ -0,0 +1,62 @@
...
+    filters: list[
+        ex.Expression
+    ] = []  # accumulate filters into this as traverse downwards
+    cluster_col: Optional[str] = None
+    while isinstance(cur_node, de_cachable_types):
+        if isinstance(cur_node, nodes.FilterNode):
+            filters.append(cur_node.predicate)
+        elif isinstance(cur_node, nodes.ProjectionNode):
+            bindings = {name: expr for expr, name in cur_node.assignments}
+            filters = [i.bind_all_variables(bindings) for i in filters]
```
tswast, line 48: I had a little trouble understanding what this is doing / why it is needed. Could we add comments explaining that filters contains unbound variables and that a projection node is the only node in which variables could be defined?

> the only node in which variables could be defined?

I'm not sure that's true or not. Maybe they could also be defined in table nodes, but those wouldn't make sense to recompute?

Maybe "projection node is the only de-cachable node"?

bigframes/session/planner.py
```diff
@@ -0,0 +1,62 @@
...
+
+        cur_node = cur_node.child
+        cur_node_refs = node_counts.get(cur_node, 0)
+        if cur_node_refs > caching_target_refs:
+            caching_target, caching_target_refs = cur_node, cur_node_refs
+            cluster_col = None
+            # Just pick the first cluster-compatible predicate
```
tswast, line 55: TODO to sort by a selectivity heuristic? Seems like this layer might make more sense than `cluster_cols_for_predicate` to do that sort.

tswast (changes requested):

bigframes/session/planner.py
TrevorBergeron: added todo

bigframes/session/planner.py
TrevorBergeron: Added some more comments, we only de-cache projection and filter nodes, and filter nodes do not define any variables, so no need to substitute variables with their definition.

bigframes/session/planner.py
TrevorBergeron: done

bigframes/session/planner.py
TrevorBergeron: I was worried that the extra cluster columns would be redundant, but actually the fact that the users is specifying multiple filters means that each additional column probably has value. Changed to select 4 cluster columns

tests/unit/test_planner.py
TrevorBergeron: added

tests/unit/test_planner.py
TrevorBergeron: added

tests/unit/test_planner.py
TrevorBergeron: added

tests/system/small/test_series.py
TrevorBergeron: added

bigframes/session/__init__.py
TrevorBergeron: done

bigframes/core/pruning.py
TrevorBergeron: added todo

bigframes/series.py
TrevorBergeron: Yeah, not sure how/if we should communicate this to users. I also don't want to lock in any specific execution strategy other than "we might cache if force=True, but we will make that cache as useful as possible using some unspecified approach".

bigframes/dtypes.py
```diff
@@ -214,6 +214,16 @@ def is_orderable(type: ExpressionType) -> bool:
     return not is_array_like(type) and not is_struct_like(type) and (type != GEO_DTYPE)
 
 
+def is_clusterable(type: ExpressionType) -> bool:
+    # https://cloud.google.com/bigquery/docs/clustered-tables#cluster_column_types
+    # This is based on default database type mapping, could in theory represent in non-default bq type to cluster.
+    return (
+        not is_array_like(type)
+        and not is_struct_like(type)
+        and (type not in (GEO_DTYPE, TIME_DTYPE, FLOAT_DTYPE))
```
tswast, line 223: Geo is clusterable but not orderable.

tswast: Should we make this an allowlist, instead? I suspect as new types are added they aren't likely to be clusterable.

TrevorBergeron: added clusterable property to dtype metadata struct, defaulting as `False`

bigframes/dtypes.py
```diff
@@ -209,9 +252,21 @@ def is_comparable(type: ExpressionType) -> bool:
...
+_ORDERABLE_SIMPLE_TYPES = set(mapping.dtype for mapping in SIMPLE_TYPES)
+
+
 def is_orderable(type: ExpressionType) -> bool:
     # On BQ side, ARRAY, STRUCT, GEOGRAPHY, JSON are not orderable
-    return not is_array_like(type) and not is_struct_like(type) and (type != GEO_DTYPE)
+    return type in _ORDERABLE_SIMPLE_TYPES
+
+
+_CLUSTERABLE_SIMPLE_TYPES = set(mapping.dtype for mapping in SIMPLE_TYPES)
```
tswast, line 263: Are we missing an `if clusterable` here?

```suggestion
_CLUSTERABLE_SIMPLE_TYPES = set(mapping.dtype for mapping in SIMPLE_TYPES if mapping.clusterable)
```

bigframes/session/planner.py
```diff
@@ -0,0 +1,75 @@
...
+        else:
+            raise ValueError(f"Unexpected de-cached node: {cur_node}")
+
+        cur_node = cur_node.child
+        cur_node_refs = node_counts.get(cur_node, 0)
+        if cur_node_refs > caching_target_refs:
+            caching_target, caching_target_refs = cur_node, cur_node_refs
+            schema = cur_node.schema
+            # Cluster cols only consider the target object and not other sesssion objects
+            # Note, this
```
tswast, line 62: Looks like this comment ended mid-sentence.

bigframes/session/planner.py
```diff
@@ -0,0 +1,75 @@
...
+            # that instead reference variables in the child node.
+            bindings = {name: expr for expr, name in cur_node.assignments}
+            filters = [i.bind_all_variables(bindings) for i in filters]
+        else:
+            raise ValueError(f"Unexpected de-cached node: {cur_node}")
+
+        cur_node = cur_node.child
+        cur_node_refs = node_counts.get(cur_node, 0)
+        if cur_node_refs > caching_target_refs:
+            caching_target, caching_target_refs = cur_node, cur_node_refs
```
tswast, line 59: Do we need to do anything to make sure we aren't selecting more columns than needed? I have some worries that column selection wouldn't have the desired affect.

Though, I suppose that'll only matter with unordered + unindexed DataFrames due to our hashing of the row. Maybe worth a TODO to be resolved with that project?

That said, I'd be curious to see if unordered/unindexed would benefit from caching at all due to the difficulties of using the cache in row identity joins.

tswast (approved):
LGTM with a couple of comments that would be good to resolve before merging.

bigframes/session/planner.py
TrevorBergeron: removed

bigframes/session/planner.py
TrevorBergeron: Row hashing shouldn't matter, as that only happens for initial table scan, which shouldn't need to be cached. However, yes, we could try to prune columns unused by the session before caching. Would need to be careful not to invalidate existing caching or join->projection rewriter, but should be possible. This could be done in a few ways, such as a partial cache (containing only some columns), or by rewriting all the session BFETs with a column pruning pass before caching.

PR #731, opened 2024-05-29
feat: adds bigframes.bigquery.array_to_string to convert array elements to delimited strings

chelsea-lin:
- [X] Make sure to open an issue as a bug/issue (internal issue: 342668759 )

Fixes internal issue: 342668759 🦕

### DIFF
--- a/bigframes/bigquery/__init__.py
+++ b/bigframes/bigquery/__init__.py
@@ -57,8 +57,7 @@ def array_length(series: series.Series) -> series.Series:
         dtype: Int64
 
     Args:
-        series (bigframes.series.Series):
-                A Series with array columns.
+        series (bigframes.series.Series): A Series with array columns.
 
     Returns:
         bigframes.series.Series: A Series of integer values indicating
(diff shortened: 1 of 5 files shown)


bigframes/bigquery/__init__.py
```diff
@@ -119,3 +118,33 @@ def array_agg(
         raise ValueError(
             f"Unsupported type {type(obj)} to apply `array_agg` function. {constants.FEEDBACK_LINK}"
         )
+
+
+def array_to_string(series: series.Series, delimiter: str) -> series.Series:
```
tswast, line 123: Should we add `null_text` as an optional parameter, too?

https://cloud.google.com/bigquery/docs/reference/standard-sql/array_functions#array_to_string

I recall it's a bit difficult to test arrays containing nulls since there are many paths that don't support them, so I'm okay omitting for now.

tswast (approved):

bigframes/bigquery/__init__.py
chelsea-lin: I initially experimented with using `null_text` but determined it wasn't beneficial at this stage. The `null_text` parameter is designed to handle null values within arrays, whereas BF arrays cannot contain nulls. This stems from BigQuery's behavior of permitting nulls in intermediate query results but not in materialized views or tables.

PR #733, opened 2024-05-29
feat: `merge` only generates a default index if both inputs already have an index

tswast:


bigframes/_config/__init__.py
```diff
@@ -61,10 +61,12 @@ def _init_bigquery_thread_local(self):
     @property
     def bigquery(self) -> bigquery_options.BigQueryOptions:
         """Options to use with the BigQuery engine."""
-        if self._local.bigquery_options is not None:
+        if (
+            bigquery_options := getattr(self._local, "bigquery_options", None)
```
tswast, line 65: Not 100% sure how this was failing, but for some reason `self._local.bigquery_options` wasn't being initialized when I ran from a notebook in VS Code.

I think it's because the original `Options.__init__` from the import ran in a different thread from the one where the actual cell code was running.

bigframes/core/blocks.py
```diff
@@ -1930,10 +1930,22 @@ def merge(
...
+        # indexes. In other words, joining anything to a NULL index object
+        # keeps everything as a NULL index.
+        #
+        # This keeps us from generating an index if the user joins a large
+        # BigQuery table against small local data, for example.
+        if len(self._index_columns) > 0 and len(other._index_columns) > 0:
+            offset_index_id = guid.generate_guid()
+            expr = joined_expr.promote_offsets(offset_index_id)
+            index_columns = [offset_index_id]
+        else:
```
tswast, line 1944: TODO: Add a test for this branch (other than the `index_col_null.ipynb` notebook)

tests/system/small/test_null_index.py
```diff
@@ -0,0 +1,288 @@
+# Copyright 2024 Google LLC
```
tswast, line 1: Moved from tests/system/small/test_empty_index.py and added `test_null_index_merge_*` tests.

TrevorBergeron (approved):

bigframes/_config/__init__.py
tswast: Fixed by https://github.com/googleapis/python-bigquery-dataframes/pull/741

PR #734, opened 2024-05-30
tests: add benchmark and split nox sessions into multiple kokoro jobs

milkshakeiii:
1. add benchmark nightly kokoro job with placeholder benchmark script
2. also split notebook and doctests tests to separate kokoro jobs to ensure more issues surface when flakes occur
3. also fix a bug where reporting variable was getting overwritten


.kokoro/continuous/doctest.cfg
```diff
@@ -0,0 +1,17 @@
+# Format: //devtools/kokoro/config/proto/build.proto
```
tswast, line 1: Could you add the google3 one first? I want to see this job trigger before we merge this PR.

noxfile.py
```diff
@@ -658,25 +658,16 @@ def system_prerelease(session: nox.sessions.Session):
...
-        "pytest-retry",
-        "nbmake",
-        "google-cloud-aiplatform",
-        "matplotlib",
-        "seaborn",
-    )
-
-    notebooks_list = list(Path("notebooks/").glob("*/*.ipynb"))
-
+    folders = [
```
tswast, line 661: I worry about making this an allowlist. Seems like the default should be to run notebooks in the notebooks session

milkshakeiii: Okay, we can do a folders_denylist then if that's OK (since we will want to filter out experimental and benchmark notebooks by folder).

notebooks/benchmark/simple_benchmark.ipynb
```diff
@@ -0,0 +1,64 @@
+{
```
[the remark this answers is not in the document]
milkshakeiii: Well, it's mostly just because the performance monitoring feature I added first was specifically for the notebooks. I also thought tests would be too granular if we had a separate output file for each test (or the opposite problem if it was one output per pytest command). I didn't even think about just using regular python files, though. That would seem to make sense, do you think it's preferred to notebooks?

.kokoro/continuous/doctest.cfg
milkshakeiii: For sure, I'm working through some presubmit issues on that one right now.

notebooks/benchmark/simple_benchmark.ipynb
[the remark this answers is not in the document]
tswast: It would certainly be simpler to setup. :-)

I don't think there's anything notebook-specific that we need to benchmark, is there?

noxfile.py
tswast: Yes, I'd prefer a denylist. Alternatively, I'd be curious to see exploring just using regular pytest / python files for our benchmark runs.

notebooks/benchmark/simple_benchmark.ipynb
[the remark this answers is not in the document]
milkshakeiii: Makes sense, we'll switch to regular python files for the benchmark tests then. Thanks!

noxfile.py
milkshakeiii: Per below, we're switching to regular python files for benchmarks.

milkshakeiii:
nightly job local test:

---BIGQUERY USAGE REPORT---
simple_benchmark_2.py - query count: 15, bytes processed sum: 16000000, slot millis sum: 280025
simple_benchmark.py - query count: 15, bytes processed sum: 16000000, slot millis sum: 271229
---total queries: 30, total bytes: 32000000, total slot millis: 551254---

.kokoro/presubmit/doctest.cfg
```diff
@@ -0,0 +1,17 @@
...
+
+# Only run this nox session.
+env_vars: {
+    key: "NOX_SESSION"
+    value: "doctest"
+}
+
+env_vars: {
+    key: "GOOGLE_CLOUD_PROJECT"
+    value: "bigframes-load-testing"
```
tswast, line 11: I suspect as the number of user journeys we want to test grows, it might make sense to test in a separate project. I'm especially worried about some APIs with limited per-project quota like Cloud Functions, BigQuery Remote Functions, and some Vertex + BQML flows. What do you think?

noxfile.py
```diff
@@ -803,24 +807,59 @@ def notebook(session: nox.Session):
...
+                    LOGGING_NAME_ENV_VAR: "scripts/benchmark/"
+                    + os.path.basename(benchmark)
+                }
+            },
+        )
+        process.start()
+        processes.append(process)
+
+    for process in processes:
+        process.join()
```
tswast, line 837: Fancy! I think it's smart to run these in parallel, though it might increase the variance a bit due to resource contention.

tswast (approved):

.kokoro/presubmit/doctest.cfg
milkshakeiii: I believe Cloud Support can increase almost any quota and some of them are meant to be increased (although not sure about Functions) but I agree it would make sense to share the burden bigframes-load-testing is doing right now for the sake of cleanliness if nothing else, since it has several different functions currently which could get confusing.

PR #682, opened 2024-05-10
fix: Improve escaping of literals and identifiers

TrevorBergeron:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery-dataframes/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #<issue_number_goes_here> 🦕

PR #736, opened 2024-05-30
feat: support `bigquery.vector_search()`

ashleyxuu:

Fixes internal #344019437 🦕


bigframes/bigquery/__init__.py
```diff
@@ -20,19 +20,17 @@
...
 
-import typing
+from typing import Literal, Optional, Union
 
+import bigframes.bigquery.utils as utils
 import bigframes.constants as constants
 import bigframes.core.groupby as groupby
 import bigframes.operations as ops
 import bigframes.operations.aggregations as agg_ops
+import bigframes.pandas as bpd
```
tswast, line 30: Cloud RAD is really bad at finding classes when we do things this way. Let's use `bigframes.series.Series` and `bigframes.dataframe.DataFrame` for our type annotations, please. See: b/296390934

bigframes/bigquery/utils.py
```diff
@@ -0,0 +1,85 @@
+# Copyright 2024 Google LLC
```
tswast, line 1: No `utils` files, please (go/no-utils). We should consolidate our SQL generation following Trevor's recent refactor. See: https://github.com/googleapis/python-bigquery-dataframes/pull/682

tswast (changes requested):

bigframes/bigquery/__init__.py
ashleyxuu: Done.

bigframes/bigquery/utils.py
ashleyxuu: Done.

bigframes/core/sql.py
```diff
@@ -169,3 +169,47 @@ def ordering_clause(
...
+
+    if query_column_to_search is not None:
+        query_str = f"""
+    SELECT
+        query.*,
+        base.*,
+        distance,
+    FROM VECTOR_SEARCH(
+        TABLE `{base_table}`,
+        "{column_to_search}",
```
tswast, line 194: I think it's possible for column names to have quote characters.

Let's use `simple_literal()` to make sure we do the right escaping for all of the options values https://github.com/googleapis/python-bigquery-dataframes/blob/ea5d367d5ecc6826d30082e75c957af8362c9e61/bigframes/core/sql.py#L36

tswast: Even if column names don't allow such characters, we don't want `vector_search()` to accidentally expose our users to SQL injection attacks.

bigframes/bigquery/__init__.py
```diff
@@ -148,3 +152,150 @@ def array_to_string(series: series.Series, delimiter: str) -> series.Series:
...
+    Conduct vector search which searches embeddings to find semantically similar entities.
+
+    **Examples:**
+
+
+        >>> import bigframes.pandas as bpd
+        >>> import bigframes.bigquery as bbq
+        >>> bpd.options.display.progress_bar = None
+
+    DataFrame embeddings for which to find nearest neighbors:
```
tswast, line 178: Let's clarify here that the `ARRAY<FLOAT>` column is used as the search query.

bigframes/bigquery/__init__.py
```diff
@@ -148,3 +152,150 @@ def array_to_string(series: series.Series, delimiter: str) -> series.Series:
...
+        ...             top_k=2)
+             embedding  id my_embedding  distance
+        dog    [1. 2.]   1      [1. 2.]       0.0
+        cat  [3.  5.2]   5    [5.  5.4]  2.009975
+        dog    [1. 2.]   4    [1.  3.2]       1.2
+        cat  [3.  5.2]   2      [2. 4.]   1.56205
+        <BLANKLINE>
+        [4 rows x 4 columns]
+
+    You can specify the name of the column in the query DataFrame embeddings and distance type:
```
tswast, line 213: Presumably `query_column_to_search` is required in this case, right? Let's tell the user that if so. Or if not, what happens if they don't specify?

tswast (changes requested):

bigframes/bigquery/__init__.py
ashleyxuu: done

bigframes/bigquery/__init__.py
ashleyxuu: Done

bigframes/bigquery/__init__.py
```diff
@@ -148,3 +152,153 @@ def array_to_string(series: series.Series, delimiter: str) -> series.Series:
...
+    Conduct vector search which searches embeddings to find semantically similar entities.
+
+    **Examples:**
+
+
+        >>> import bigframes.pandas as bpd
+        >>> import bigframes.bigquery as bbq
+        >>> bpd.options.display.progress_bar = None
+
+    DataFrame embeddings for which to find nearest neighbors, and ARRAY<FLOAT> column
```
tswast, line 178: Let's use code font for `ARRAY<FLOAT>`. Also the BQ type is called `FLOAT64` in standard SQL https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#floating_point_types

```suggestion
    DataFrame embeddings for which to find nearest neighbors. The ``ARRAY<FLOAT64>`` column
```

tswast (approved):

PR #738, opened 2024-05-30
refactor: New node type for cached subtrees

TrevorBergeron:


bigframes/core/compile/compiler.py
```diff
@@ -96,6 +96,42 @@ def compile_readlocal(node: nodes.ReadLocalNode, ordered: bool = True):
...
+@_compile_node.register
+def compile_cached_table(node: nodes.CachedTableNode, ordered: bool = True):
+    full_table_name = f"{node.project_id}.{node.dataset_id}.{node.table_id}"
+    hidden_col_names = [
+        col for col in node.ordering.referenced_columns if col not in node.schema.names
+    ]
+    used_columns = (
+        *node.schema.names,
+        *hidden_col_names,
+    )
```
tswast, line 108: I wonder if it would be better to move this "used columns" logic to a more central place such as CachedTableNode itself? Seems like something that could potentially get out of sync if we add other kinds of hidden columns.

bigframes/core/nodes.py
```diff
@@ -409,6 +409,49 @@ def transform_children(
...
+    physical_schema: Tuple[bq.SchemaField, ...] = field()
+
+    ordering: orderings.ExpressionOrdering = field()
+
+    @property
+    def session(self):
+        return self.original_node.session
+
+    def __hash__(self):
+        return self._node_hash
```
tswast, line 430: Where is this initialized?

Thoughts on calling `hash(self.original_node)`, instead?

bigframes/core/nodes.py
```diff
@@ -409,6 +409,49 @@ def transform_children(
...
+    def roots(self) -> typing.Set[BigFrameNode]:
+        return {self}
+
+    @property
+    def schema(self) -> schemata.ArraySchema:
+        return self.original_node.schema
+
+    @property
+    def relation_ops_created(self) -> int:
+        # Assume worst case, where readgbq actually has baked in analytic operation to generate index
```
tswast, line 442: I'm curious when this is used? If the cost is calculated after we already materialized the table, seems like it wouldn't apply.

bigframes/core/ordering.py
```diff
@@ -215,6 +215,15 @@ def is_sequential(self) -> bool:
     def all_ordering_columns(self) -> Sequence[OrderingExpression]:
         return list(self.ordering_value_columns)
 
+    @property
+    def referenced_columns(self) -> Sequence[str]:
+        cols = []
+        for part in self.ordering_value_columns:
+            for col in part.scalar_expression.unbound_variables:
+                if col not in cols:
```
tswast, line 223: Does the order of `cols` matter? If not a `set()` may be more appropriate than a list.

bigframes/session/__init__.py
```diff
@@ -1154,35 +1153,14 @@ def _read_pandas_load_job(
...
-            self,
-            table_expression,
-            columns=column_values,
-            hidden_ordering_columns=[table_expression[ordering_col]],
-            ordering=ordering,
-        )
+        destination_table = self.bqclient.get_table(load_table_destination)
+        array_value = core.ArrayValue.from_table(
+            table=destination_table,
+            # TODO: Generate this directly from original pandas df.
```
tswast, line 1159: Would we actually want to do that, though? Presumably the type in BQ is the source of truth for us.

TrevorBergeron: I was thinking about if we virtualize some datatypes such as microsecond intervals, which would be represented by integer in bq

bigframes/core/nodes.py
TrevorBergeron: that is true, this node doesn't need to generate ordering since it will already be provided by the originating query. changed to default single relational op

bigframes/core/ordering.py
TrevorBergeron: Changed to set, caller can sort for determinism

bigframes/core/nodes.py
TrevorBergeron: The _node_hash is a property defined on the base class that introspects all the dataclass fields to generate a hash. Generating an equivalent hash to `original_node` is an interesting idea, but I think for the most part, we want hash to represent structural equivalence rather than going for some sort of imperfect semantic equivalence.

bigframes/core/__init__.py
```diff
@@ -95,6 +95,23 @@ def from_pyarrow(cls, arrow_table: pa.Table, session: Session):
         )
         return cls(node)
 
+    @classmethod
+    def from_cached(
+        cls,
+        original: ArrayValue,
+        table: google.cloud.bigquery.Table,
```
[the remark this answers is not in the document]
TrevorBergeron: Yea, is the cache destination. Right now, it must already been materialized.

bigframes/core/compile/compiler.py
TrevorBergeron: Moved logic to define "hidden columns" to node definitions.

tswast (approved):

PR #745, opened 2024-06-03
refactor: Move ibis type mappings to compile module

TrevorBergeron:

### DIFF
--- a/bigframes/core/compile/scalar_op_compiler.py
+++ b/bigframes/core/compile/scalar_op_compiler.py
@@ -27,6 +27,7 @@ import numpy as np
 import pandas as pd
 
 import bigframes.constants as constants
+import bigframes.core.compile.ibis_types
 import bigframes.core.expression as ex
 import bigframes.dtypes
 import bigframes.operations as ops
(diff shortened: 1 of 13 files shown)


bigframes/core/compile/scalar_op_compiler.py
```diff
@@ -1013,15 +1016,7 @@ def add_op(
...
     if isinstance(x, ibis_types.NullScalar) or isinstance(x, ibis_types.NullScalar):
         return ibis.null()
-    try:
-        # Could be string concatenation or numeric addition.
-        return x + y  # type: ignore
-    except ibis.common.annotations.SignatureValidationError as exc:
-        left_type = bigframes.dtypes.ibis_dtype_to_bigframes_dtype(x.type())
-        right_type = bigframes.dtypes.ibis_dtype_to_bigframes_dtype(y.type())
-        raise TypeError(
-            f"Cannot add {repr(left_type)} and {repr(right_type)}. {constants.FEEDBACK_LINK}"
```
junyazhang, line 1023: Since we are removing the try except block here, shall we add the FEEDBACK_LINK to https://github.com/googleapis/python-bigquery-dataframes/blob/93353c368f6af5ae 186493137616bf562d890118/bigframes/operations/__init__.py#L629?

junyazhang (approved):

bigframes/core/compile/scalar_op_compiler.py
TrevorBergeron: So far the pattern has been to give feedback link when we have missing features or bugs, not for user error. I would consider this to be a user error, as it would be caused by users trying to add two incompatible types. Maybe we should just always provide feedback error though, just in case the users think its a feature gap?

PR #746, opened 2024-06-04
fix: Improve to_pandas_batches for large results

milkshakeiii:


tests/system/load/test_large_tables.py
```diff
@@ -76,19 +76,21 @@ def test_index_repr_large_table():
...
+    for column in df:
+        df[column + "_2"] = df[column]
     # df will be downloaded locally
     expected_row_count, expected_column_count = df.shape
 
     row_count = 0
-    # TODO(b/340890167): fix type error
-    for df in df.to_pandas_batches():  # type: ignore
-        batch_row_count, batch_column_count = df.shape
+    for pdf in df.to_pandas_batches(page_size=500000):
```
tswast, line 86: I think `page_size` can counter-intuitively result in much lower throughput, as it requires the use of the BigQuery REST API, not the BigQuery Storage API (which has some tunables but not exactly based on number of rows per page). I'd prefer we wait to add this parameter until we get an explicit request for it.

milkshakeiii: I will change it to only download 3 batches, so that we don't need to use large pages just to get the test to finish in a reasonable time. Then we can delay putting in the page_size parameter as you suggest.

bigframes/core/blocks.py
```diff
@@ -512,9 +512,17 @@ def to_pandas_batches(self):
...
-        results_iterator, _ = self.session._execute(self.expr, sorted=True)
+        _, query_job = self.session._query_to_destination(
+            self.session._to_sql(self.expr, sorted=True),
+            list(self.index_columns),
+            api_name="cached",
+            do_clustering=False,
+        )
+        results_iterator = query_job.result()
         for arrow_table in results_iterator.to_arrow_iterable(
-            bqstorage_client=self.session.bqstoragereadclient
```
milkshakeiii, line 517: @tswast Does this mesh with what you know of the python client? It doesn't appear to be streaming one page at a time when the storage feature is being used. https://github.com/googleapis/python-bigquery/blob/cc7b3995eef4f357f031e06389db07f949987eeb/google/cloud/bigquery/_pandas_helpers.py#L816

milkshakeiii: In other words based on what I'm seeing we should either do this or go back to having the page_size parameter, because the current behavior doesn't really work (isn't fully "batched") for large tables. This way seems perhaps better because, as you mentioned, it's counterintuitive for passing page_size to change behavior so drastically. But adding page_size would also make sense to me since it's something people would probably try if they needed to (for example I knew to try it when I encountered slowness this time, at least.)

tswast: It should be sending one message at a time back via the stream, but I think you're right that there might not be a mechanism to shut it down if we stop iterating over it.

I'd rather we allow `page_size` as a workaround to users than never use the BQ Storage API at all. Let's go back to the way you had it. :-)

tswast (approved):
Thanks!

PR #762, opened 2024-06-06
chore: cleanup type errors in bigframes/core/indexes/base.py

Genesis929:

### DIFF
--- a/bigframes/core/indexes/base.py
+++ b/bigframes/core/indexes/base.py
@@ -90,9 +90,12 @@ class Index(vendored_pandas_index.Index):
         # TODO: Support more index subtypes
         from bigframes.core.indexes.multi import MultiIndex
 
-        klass = MultiIndex if len(block._index_columns) > 1 else cls
-        # TODO(b/340893286): fix type error
-        result = typing.cast(Index, object.__new__(klass))  # type: ignore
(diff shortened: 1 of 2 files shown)


bigframes/core/indexes/base.py
```diff
@@ -90,9 +90,12 @@ def __new__(
         # TODO: Support more index subtypes
         from bigframes.core.indexes.multi import MultiIndex
 
-        klass = MultiIndex if len(block._index_columns) > 1 else cls
-        # TODO(b/340893286): fix type error
-        result = typing.cast(Index, object.__new__(klass))  # type: ignore
+        if len(block._index_columns) <= 1:
```
GarrettWu, line 93: hmmm. I'm surprised that mypy distinguishes those two syntax. Do you know why is it?

GarrettWu (approved):

bigframes/core/indexes/base.py
Genesis929: From what I find, first, mypy can't tell the type from things like x if y else z, it will assume it's object.

Then, if we use if else, multiindex is a subclass of index, so we need to handle the index class first, otherwise it will return another type error. 

So this is the final result.

PR #772, opened 2024-06-10
refactor: create internal session ordering mode flag

TrevorBergeron:


bigframes/_config/bigquery_options.py
```diff
@@ -235,3 +239,8 @@ def kms_key_name(self, value: str):
             raise ValueError(SESSION_STARTED_MESSAGE.format(attribute="kms_key_name"))
 
         self._kms_key_name = value
+
+    @property
+    def _strictly_ordered(self) -> bool:
```
tswast, line 244: Presumably we'll have a public version of this at some future date?

bigframes/core/blocks.py
```diff
@@ -2474,7 +2474,7 @@ def column_ids(self) -> Sequence[str]:
         """Column(s) to use as row labels."""
         return self._block._index_columns
 
-    def to_pandas(self) -> pd.Index:
+    def to_pandas(self, *, sorted: Optional[bool] = None) -> pd.Index:
```
tswast, line 2477: I'm curious how this relates to the existing `ordered` bool on the public `to_pandas` method?

https://github.com/googleapis/python-bigquery-dataframes/blob/d5ae6807022aa556e80c76a254c2a72aef82f54b/bigframes/series.py#L326

and

https://github.com/googleapis/python-bigquery-dataframes/blob/d5ae6807022aa556e80c76a254c2a72aef82f54b/bigframes/dataframe.py#L1191

Should we align these names?

bigframes/dataframe.py
```diff
@@ -1195,7 +1195,7 @@ def to_pandas(
                 The seed for the uniform downsampling algorithm. If provided, the uniform method may
                 take longer to execute and require more computation. If set to a value other than
                 None, this will supersede the global config.
-            ordered (bool, default True):
+            ordered (bool, default None):
```
tswast, line 1198: Let's say what `None` does. (I assume it respects any `sort_values` calls but may be non-deterministic in the case of not a total orderign?)

bigframes/_config/bigquery_options.py
TrevorBergeron: Yes, this is just a private placeholder until the feature is complete enough to expose through a public API. We will also need to settle on the naming.

bigframes/core/blocks.py
TrevorBergeron: Yes, the names mean the same thing. Renamed to `ordered`

bigframes/dataframe.py
TrevorBergeron: `None` just uses the session default, there is still only 2 modes (True/False). Extended arg description in new revision

bigframes/core/blocks.py
```diff
@@ -2484,7 +2484,10 @@ def to_pandas(self) -> pd.Index:
         index_columns = list(self._block.index_columns)
         dtypes = dict(zip(index_columns, self.dtypes))
         expr = self._expr.select_columns(index_columns)
-        results, _ = self.session._execute(expr)
+        results, _ = self.session._execute(
+            expr,
+            sorted=ordered if (ordered is not None) else self.session._strictly_ordered,
```
tswast, line 2489: Is it too late to rename this one too?

TrevorBergeron: renamed remaining `sorted` to `ordered` in new revision

bigframes/series.py
```diff
@@ -323,7 +323,7 @@ def to_pandas(
         sampling_method: Optional[str] = None,
         random_state: Optional[int] = None,
         *,
-        ordered: bool = True,
+        ordered: Optional[bool] = None,
```
chelsea-lin, line 326: I might not fully follow the conversation. How to handle the breaking changes usually?

tests/system/small/test_unordered.py
```diff
@@ -26,3 +27,34 @@ def test_unordered_mode_cache_aggregate(unordered_session):
     pd_result = pd_df - pd_df.mean()
 
     assert_pandas_df_equal(bf_result, pd_result, ignore_order=True)
+
+
+@skip_legacy_pandas
+def test_unordered_mode_read_gbq(unordered_session):
+    df = unordered_session.read_gbq(
```
chelsea-lin, line 34: can the test be flaky if the `unordered_session` may return data in a different order?

TrevorBergeron: yes, should have `ignore_order=True` like the other test. added

bigframes/series.py
TrevorBergeron: I guess I don't see it as a breaking change necessarily, since all existing code should work the same `to_pandas()` without ordered parameter will still be ordered by default for all existing code.

chelsea-lin (approved):

PR #775, opened 2024-06-11
refactor: update core/sql.py selects to use SQL AST nodes and nodes updates.

Genesis929:


bigframes/core/compile/googlesql/query.py
```diff
@@ -104,45 +112,44 @@ def sql(self) -> str:
...
+                f"Unsupported expression type {type(self.expression).__name__};"
+                "expected one of TableExpression, QueryExpr, str, or CTEExpression."
+            )
 
         if self.as_alias is None:
             return text
         else:
             return f"{text} {self.as_alias.sql()}"
 
+    @classmethod
```
chelsea-lin, line 138: nit: put the `from_table_ref` before the `sql` method?

bigframes/core/compile/googlesql/query.py
```diff
@@ -51,10 +55,14 @@ class Select(abc.SQLSyntax):
 
     select_list: typing.Sequence[typing.Union[SelectExpression, SelectAll]]
     from_clause_list: typing.Sequence[FromClause] = ()
+    distinct: bool = False
 
     def sql(self) -> str:
         text = ["SELECT"]
 
+        if self.distinct:
```
chelsea-lin, line 63: Could you please write unit tests for the new functions in this file? I'm aiming for 100% unit-test coverage for AST compilation.

bigframes/core/compile/googlesql/query.py
Genesis929: Updated.

bigframes/core/compile/googlesql/query.py
Genesis929: Updated.

chelsea-lin (approved):
LGTM! Thanks!

TrevorBergeron (changes requested):
Are we preserving identifier escaping (for column names and table expression names)?

chelsea-lin:
> Are we preserving identifier escaping (for column names and table expression names)?
I expect more refactor PRs coming for `core/sql.py`. But you're right, the table reference should escape the identifier according to the [table naming rules](https://cloud.google.com/bigquery/docs/tables#table_naming), while the column names should be find according to the [column naming rules](https://cloud.google.com/bigquery/docs/schemas#column_names). But we can include all of them for simpler

bigframes/core/compile/googlesql/expression.py
```diff
@@ -72,10 +73,10 @@ def __post_init__(self):
...
         text = []
         if self.project_id is not None:
-            text.append(f"`{self.project_id}`")
+            text.append(f"`{utils._escape_special_characters(self.project_id)}`")
         if self.dataset_id is not None:
-            text.append(f"`{self.dataset_id}`")
-        text.append(f"`{self.table_id}`")
+            text.append(f"`{utils._escape_special_characters(self.dataset_id)}`")
+        text.append(f"`{utils._escape_special_characters(self.table_id)}`")
         return ".".join(text)
```
chelsea-lin, line 80: nit: `utils._escape_special_characters(".".join(text))` only?

TrevorBergeron (approved):

chelsea-lin (approved):

bigframes/core/compile/googlesql/expression.py
Genesis929: This may not work.

PR #782, opened 2024-06-12
feat: add `bigframes.bigquery.json_set`

chelsea-lin:
BEGIN_COMMIT_OVERRIDE
feat: add `bigframes.bigquery.json_set` (#782)
END_COMMIT_OVERRIDE

This change is creating a `bigframes.bigquery.json_set` API, which can produces a new JSON value within a Series by inserting or replacing values at specified paths.
- [X] Make sure to open an issue as a internal bug 346813016

Fixes internal issue 346813016 🦕


bigframes/bigquery/__init__.py
```diff
@@ -154,6 +158,64 @@ def array_to_string(series: series.Series, delimiter: str) -> series.Series:
...
+    """
+    # SQLGlot parser does not support the "create_if_missing => true" syntax, so
+    # create_if_missing is not currently implemented.
+
+    # Currently limited to single JSON path/value pairs (binary operations only).
+    if len(json_path_value_pairs) != 1:
+        raise ValueError(
+            "Expected exactly one JSON path and value pair but found "
+            + f"{len(json_path_value_pairs)} pairs."
+        )
```
TrevorBergeron, line 205: Why not just loop through the pairs and repeatedly apply?

bigframes/operations/__init__.py
```diff
@@ -707,6 +707,19 @@ def output_type(self, *input_types):
...
+@dataclasses.dataclass(frozen=True)
+class JSONSet(BinaryOp):
+    name: typing.ClassVar[str] = "json_set"
+    json_path: str
+
+    def output_type(self, *input_types):
+        input_type = input_types[0]
+        if not dtypes.is_json_like(input_type):
+            raise TypeError("Input type must be an JSON or JSON-formatted string type.")
+        return input_type
```
TrevorBergeron, line 720: Are there any restrictions on the right argument? Or does it support every single datatype?

bigframes/core/compile/scalar_op_compiler.py
```diff
@@ -894,6 +894,16 @@ def array_to_string_op_impl(x: ibis_types.Value, op: ops.ArrayToStringOp):
...
 
 
+# JSON Ops
+@scalar_op_compiler.register_binary_op(ops.JSONSet, pass_op=True)
+def json_set_op_impl(x: ibis_types.Value, y: ibis_types.Value, op: ops.JSONSet):
+    return vendored_ibis_ops.JSONSet(
+        x,
+        json_value=y,
+        json_path=op.json_path,
+    ).to_expr()
```
TrevorBergeron, line 904: is it possible to implement through the ibis udf interface, or is the syntax too non-standard? The new custom op will add extra work when upgrading to ibis 9

tswast: See: https://github.com/googleapis/python-bigquery-dataframes/blob/ec5b06856762e67060642cc106f43f0d202a08bf/bigframes/core/compile/scalar_op_compiler.py#L1462 for an example.

bigframes/bigquery/__init__.py
chelsea-lin: Good points. It should be solved!

bigframes/operations/__init__.py
chelsea-lin: Good points. Added additional type checking here.

bigframes/core/compile/scalar_op_compiler.py
chelsea-lin: yes, it can be implemented through ibis udf interface.

TrevorBergeron (approved):
LGTM

PR #786, opened 2024-06-12
refactor: Reduce compile interface to returning just SQL text

TrevorBergeron:


bigframes/core/compile/api.py
```diff
@@ -0,0 +1,72 @@
...
+    return compiler.compile_ordered_ir(node).to_sql(
+        col_id_overrides=col_id_overrides, sorted=True
+    )
+
+
+def compile_raw(
+    node: bigframes.core.nodes.BigFrameNode,
+) -> Tuple[str, bigframes.core.ordering.ExpressionOrdering]:
+    """Compile node into sql that exposes all columns, including hidden ordering-only columns."""
+    ir = compiler.compile_ordered_ir(node)
```
chelsea-lin, line 51: Can `compiled_raw` return an unordered version of SQL?

chelsea-lin (approved):

bigframes/core/compile/api.py
TrevorBergeron: No, `compile_raw` exits to return the hidden ordering columns. This is needed primarily to cache as bq table without generating row numbers. `compile_unordered` should be used if ordering information isn't needed.

PR #789, opened 2024-06-13
fix: df.loc with the 2nd input as bigframes boolean Series

GarrettWu:

Fixes #347032818 🦕

### DIFF
--- a/tests/system/small/test_dataframe.py
+++ b/tests/system/small/test_dataframe.py
@@ -3023,11 +3023,19 @@ def test_loc_select_with_column_condition(scalars_df_index, scalars_pandas_df_in
 def test_loc_select_with_column_condition_bf_series(
     scalars_df_index, scalars_pandas_df_index
 ):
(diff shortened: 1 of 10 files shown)


tests/system/small/test_dataframe.py
```diff
@@ -3020,6 +3020,21 @@ def test_loc_select_with_column_condition(scalars_df_index, scalars_pandas_df_in
     )
 
 
+def test_loc_select_with_column_condition_bf_series(
+    scalars_df_index, scalars_pandas_df_index
+):
+    bf_result = scalars_df_index.loc[
+        :, series.Series(scalars_df_index.dtypes == "Int64")
```
tswast, line 3027: Could we do a test that derives from `scalars_df_index` instead? I'm a little worried about index alignment.

GarrettWu: Done. But it identifies another problem b/347072677. May need SQL to fix.

Genesis929 (approved):

PR #797, opened 2024-06-17
fix: Self-join optimization doesn't needlessly invalidate caching

TrevorBergeron:


bigframes/core/rewrite.py
```diff
@@ -311,3 +326,25 @@ def get_node_column_ids(node: nodes.BigFrameNode) -> Tuple[str, ...]:
     import bigframes.core
 
     return tuple(bigframes.core.ArrayValue(node).column_ids)
+
+
+def common_subtree(
+    l_tree: nodes.BigFrameNode, r_tree: nodes.BigFrameNode
+) -> Optional[nodes.BigFrameNode]:
+    """Find common subtree between join subtrees"""
```
chelsea-lin, line 334: To find common substree between two trees, the substree should have same node type and as well all children nodes should be same as well. IIUC, the below code can return a substree when its root node are same while its child nodes can be different. Another question here, is the `REWRITABLE_NODE_TYPES` unary nodes only? Can be binary node also?

bigframes/core/rewrite.py
```diff
@@ -311,3 +326,25 @@ def get_node_column_ids(node: nodes.BigFrameNode) -> Tuple[str, ...]:
...
+        l_node = l_node.child
+    l_nodes.add(l_node)
+
+    r_node = r_tree
+    while isinstance(r_node, REWRITABLE_NODE_TYPES):
+        if r_node in l_nodes:
+            return r_node
+        r_node = r_node.child
+
+    if r_node in l_nodes:
```
chelsea-lin, line 348: With the above codes, this condition will never be triggered? Can you double check the test coverage?

chelsea-lin (changes requested):

bigframes/core/rewrite.py
TrevorBergeron: equality check is deep by default with dataclasses
try this for example. this returns `True`
```import dataclasses

@dataclasses.dataclass(frozen=True)
class Bar:
  a: int
  b: int

@dataclasses.dataclass(frozen=True)
class Foo:
  a: Bar


Foo(Bar(1,2)) == Foo(Bar(1,2))
```

Also, `REWRITABLE_NODE_TYPES` are the set of nodes that can be rewritten from a join into a projection.

bigframes/core/rewrite.py
TrevorBergeron: the last line in the loop is `r_node = r_node.child`. this node isn't checked on the last iteration, so must check one last time. Not the cleanest code, but this line is needed. Probably a cleaner way of organizing this

chelsea-lin: Got it! Thanks!

bigframes/core/rewrite.py
```diff
@@ -196,28 +203,36 @@ def expand(self) -> nodes.BigFrameNode:
...
-
 def maybe_rewrite_join(join_node: nodes.JoinNode) -> nodes.BigFrameNode:
-    left_side = SquashedSelect.from_node(join_node.left_child)
-    right_side = SquashedSelect.from_node(join_node.right_child)
+    rewrite_common_node = common_subtree(join_node.left_child, join_node.right_child)
+    if rewrite_common_node is None:
+        return join_node
+    left_side = SquashedSelect.from_node(join_node.left_child, rewrite_common_node)
+    right_side = SquashedSelect.from_node(join_node.right_child, rewrite_common_node)
     if left_side.can_join(right_side, join_node.join):
```
chelsea-lin, line 212: maybe rename it as `can_merge`, I thought it was checking if two trees are joinable or not.

chelsea-lin (approved):
LGTM but left comments to rename or add more comments to describe the optimization strategy.

bigframes/core/rewrite.py
```diff
@@ -116,14 +123,14 @@ def can_join(
...
+    def merge(
         self,
         right: SquashedSelect,
         join_type: join_defs.JoinType,
         mappings: Tuple[join_defs.JoinColumnMapping, ...],
-    ) -> Optional[SquashedSelect]:
+    ) -> SquashedSelect:
         if self.root != right.root:
-            return None
+            raise ValueError("Cannot merge expressions with different roots")
```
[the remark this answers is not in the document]
TrevorBergeron: So far we don't do this for other similar errors - couldn't find `InternalError` anywhere in the code base. Maybe we should automatically wrap all exceptions with feedback link? Seems out of scope

bigframes/core/rewrite.py
TrevorBergeron: done

bigframes/core/rewrite.py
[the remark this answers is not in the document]
chelsea-lin: Could be `SystemError` in the python exceptions?

PR #799, opened 2024-06-18
fix: include internally required packages in `remote_function` hash

shobsi:
We compute cloud function and BQ remote function id based on a hash computed from the udf code and external package requirements. But there are internal package requirements as well, which don't play a part in the hash computation. Because of this, if internal package version changes, it does not trigger redeployment of the cloud function. We are fixing that behavior in this change by including the internal package requirements in the hash computation. 

Fixes internal issue 345549684 🦕


tswast (approved):
Thanks!

PR #801, opened 2024-06-20
feat: Add `bigframes.streaming.to_pubsub` method to create continuous query that writes to Pub/Sub

milkshakeiii:
Also add script that simply creates a pubsub topic named "penguins"


bigframes/streaming/__init__.py
```diff
@@ -123,7 +130,108 @@ def to_bigtable(
...
+    )
+
+    # return the query job to the user for lifetime management
+    return query_job
+
+
+def to_pubsub(
+    query: str,
+    topic: str,
+    service_account: str,
```
milkshakeiii, line 159: Make these required keyword arguments.

bigframes/streaming/__init__.py
```diff
@@ -123,7 +130,108 @@ def to_bigtable(
...
+        # (should accept abstract job_config, only takes concrete)
+        job_id=job_id,
+        job_id_prefix=job_id_prefix,
+    )
+
+    # return the query job to the user for lifetime management
+    return query_job
+
+
+def to_pubsub(
```
milkshakeiii, line 156: Raise a preview warning.

bigframes/streaming/__init__.py
```diff
@@ -123,7 +130,108 @@ def to_bigtable(
...
+                "connectionProperties": {
+                    "key": "service_account",
+                    "value": service_account,
+                },
+            }
+        }
+    )
 
     # begin the query job
     query_job = bq_client.query(
```
milkshakeiii, line 237: add BigFrames API label to jobconfig object.

milkshakeiii (commented):
Add to sphinx documentation (toc.yaml and something.rst) Thanks

bigframes/streaming/__init__.py
milkshakeiii: added

bigframes/streaming/__init__.py
milkshakeiii: done

bigframes/streaming/__init__.py
milkshakeiii: done

milkshakeiii:
> Add to sphinx documentation (toc.yaml and something.rst) Thanks

This will be added in a follow-up.

tests/system/large/test_streaming.py
```diff
@@ -46,3 +47,29 @@ def test_streaming_to_bigtable():
...
+    )
+
+    try:
+        # wait 100 seconds in order to ensure the query doesn't stop
+        # (i.e. it is continuous)
+        time.sleep(100)
+        assert query_job.error_result is None
+        assert query_job.errors is None
+        assert query_job.running()
+        assert str(query_job.job_id).startswith(job_id_prefix)
```
tswast, line 73: Thoughts on checking that we wrote at least one message to the pub/sub topic?

bigframes/streaming/__init__.py
```diff
@@ -24,9 +25,11 @@
 
 def to_bigtable(
     query: str,
+    *,
     instance: str,
     table: str,
-    bq_client: Optional[bigquery.Client] = None,
+    service_account: Optional[str] = None,
```
tswast, line 31: Maybe `service_account_email`? I'm worried `service_account` is too ambiguous. I don't want someone to think they need a key file, JSON, for example.

tswast (commented):
Left some suggestions. Let me know what you think about the name change and if it's feasible to test for pub/sub message.

bigframes/streaming/__init__.py
tswast: FWIW: `service_account_email` would be consistent with google-auth. See: https://googleapis.dev/python/google-auth/latest/reference/google.auth.compute_engine.html

tests/system/large/test_streaming.py
milkshakeiii: By default, we wouldn't write any message to the pubsub topic, if I understand correctly. We would have to write to the BigQuery table in order for subscribers to receive anything. We would have to create a subscription externally and listen in this test using the client library. That would protect against any situation where the query keeps running but fails silently at actually transmitting data, but I don't think there is any situation like that, especially not one that would be a BigFrames bug rather than a continuous queries bug. Does that make sense?

bigframes/streaming/__init__.py
milkshakeiii: Makes sense, changed, thanks.

bigframes/streaming/__init__.py
```diff
@@ -123,7 +138,117 @@ def to_bigtable(
...
+
+    # return the query job to the user for lifetime management
+    return query_job
+
+
+def to_pubsub(
+    query: str,
+    *,
+    topic: str,
+    service_account: str,
```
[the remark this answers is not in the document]
milkshakeiii: Roger, changed.

tests/system/large/test_streaming.py
milkshakeiii: I'm still not sure we need it, though. We're really just launching a continuous query and as long as the query was created without errors and is indeed continuous, to me what data gets exported is the realm of continuous queries behavior and the user's SQL. (And possibly destination configuration which we are also hands-off about)

tswast (approved):

PR #803, opened 2024-06-20
feat: expose gcf memory param in `remote_function`

shobsi:

Fixes internal issue 346857621 🦕


tests/system/large/test_remote_function.py
```diff
@@ -1800,3 +1800,63 @@ def float_parser(row):
         cleanup_remote_function_assets(
             session.bqclient, session.cloudfunctionsclient, float_parser_remote
         )
+
+
+@pytest.mark.parametrize(
+    ("memory_mib_args", "expected_memory"),
+    [
+        pytest.param({}, "1024Mi", id="no-set"),
+        pytest.param({"cloud_function_memory_mib": None}, "256M", id="set-None"),
```
TrevorBergeron, line 1809: Why different values for unset vs None?

tests/system/large/test_remote_function.py
```diff
@@ -1800,3 +1800,63 @@ def float_parser(row):
...
+        pytest.param({"cloud_function_memory_mib": 128}, "128Mi", id="set-128"),
+        pytest.param({"cloud_function_memory_mib": 1024}, "1024Mi", id="set-1024"),
+        pytest.param({"cloud_function_memory_mib": 4096}, "4096Mi", id="set-4096"),
+        pytest.param({"cloud_function_memory_mib": 32768}, "32768Mi", id="set-32768"),
```
TrevorBergeron, line 1813: What happens if users provides a value that is within min-max bounds but is not a power of two?

tests/system/large/test_remote_function.py
shobsi: unset should be backward compatible, meaning we should continue to set bigframes default of 1024M. `None` is a way to indicate that any explicit memory from the bigframes side should not be set, as added in the documentation "If you want to let the default memory of cloud functions be allocated, pass `None`". There is another param `cloud_function_timeout` with similar behavior. I acknowledge the confusion though, how about I send a separate PR to make unset and None behavior same in both param for consistency?

tests/system/large/test_remote_function.py
shobsi: there is some server side logic to mod/ceil it the nearest well known number.

TrevorBergeron (approved):

PR #805, opened 2024-06-24
feat: add PolynomialFeatures to_gbq and pipeline support

GarrettWu:
BEGIN_COMMIT_OVERRIDE
feat: add `PolynomialFeatures` support to `to_gbq` and pipelines (#805)
END_COMMIT_OVERRIDE


bigframes/ml/compose.py
```diff
@@ -107,13 +111,13 @@ def _extract_from_bq_model(
...
 
-        transformers: List[
+        transformers_set: Set[
             Tuple[
                 str,
                 Union[preprocessing.PreprocessingType, impute.SimpleImputer],
                 Union[str, List[str]],
             ]
-        ] = []
+        ] = set()
```
shobsi, line 120: what's the reason for going from list to set?

GarrettWu: The processor is different from others. Where previous are 1 -> 1 transform, poly_features is n -> m. So other transformers we get 1 transformed_column, but poly_features we get m transformed_column. Which need to be deduped. And the order of the transformers doesn't matter.

shobsi (approved):

Issue #735, opened 2024-05-30
Polars Support


firmai:
It would be great to offer Polars support, it is currently half as popular as Pandas, and generally work better for large datasets. Polars is bound to replace most data-scientist day to day operations within the next five years. 

Thanks for developing bigframes, it very useful.

TrevorBergeron:
What kind of polars support would you find useful? Would you want BigQuery DataFrames to have an polars-like DataFrame API (as an alternative to the current pandas-like one) or simply interop with polars objects more easily?

lmmx:
I would like automatic schema supply, this is currently the limiting step in automatically uploading Polars DataFrames: `write_ndjson` seems to be the only way I can upload list dtypes (Parquet seems to not be viable, see [this issue](https://github.com/pola-rs/polars/issues/16938)), but NDJSON requires the schema to be passed. I'm really looking for something that will just let me put my Polars DataFrame in a BQ table without fiddling with schemas: there should be enough info already here to do that for me.

tswast:
For uploading to BigQuery, I have updated the polars docs to indicate how to get BigQuery to correctly handle list types https://github.com/pola-rs/polars/pull/20292

I think that `read_polars` and `to_polars` methods would be reasonable requests for bigframes. I have done some refactoring recently to our I/O that might make it a bit easier, but would probably require a little more refactoring to have pyarrow tables/recordbatches as the intermediate format instead of pandas dataframes. The other thing to be careful about is that polars would be an optional "extra" dependency in `setup.py` to avoid a hard dependency on the polars package.

lmmx:
Amazing! Should this issue be closed now?

tswast:
I just mailed https://github.com/googleapis/python-bigquery-dataframes/pull/1855 with `bpd.read_arrow(pyarrow.Table)` to round out the other side of this conversion.

Technically I think this was possible before by going through the DataFrame constructor, but that ended up translating to pandas as an in-between layer. Now we can just go from polars -> Arrow -> BigFrames without pandas in the middle.

PR #807, opened 2024-06-26
feat: add `DataFrame.to_arrow` to create Arrow Table from DataFrame

tswast:

Towards #735  🦕


samples/polars/noxfile.py
```diff
@@ -0,0 +1,292 @@
+# Copyright 2019 Google LLC
```
chelsea-lin, line 1: Is this file auto generated from a template? If not, can we update the copyright year to `2024`? Also check other new files too?

samples/polars/create_polars_df_with_to_arrow_test.py
```diff
@@ -0,0 +1,39 @@
...
+    # ...
+
+    # Run the query and download the results as an Arrow table to convert into
+    # a Polars DataFrame. Use ordered=False if your polars analysis is OK with
+    # non-deterministic ordering.
+    arrow_table = bf_df.to_arrow(ordered=False)
+    polars_df = polars.from_arrow(arrow_table)
+    # [END bigquery_dataframes_to_polars]
+
+    assert polars_df is not None
```
chelsea-lin, line 39: Can we assert `polars_df` is equal to `bf_df`? Or maybe there are some diverse?

chelsea-lin (approved):
LGTM

samples/polars/noxfile.py
tswast: This is templated, so not much we can do.

samples/polars/create_polars_df_with_to_arrow_test.py
tswast: We could possibly compare shapes or some aggregate values. I'll give it a try.

PR #815, opened 2024-07-02
feat: add stratify param support to ml.model_selection.train_test_split method

GarrettWu:

### DIFF
--- a/bigframes/ml/model_selection.py
+++ b/bigframes/ml/model_selection.py
@@ -29,6 +29,7 @@ def train_test_split(
     test_size: Union[float, None] = None,
     train_size: Union[float, None] = None,
     random_state: Union[int, None] = None,
+    stratify: Union[bpd.Series, None] = None,
 ) -> List[Union[bpd.DataFrame, bpd.Series]]:
     """Splits dataframes or series into random train and test subsets.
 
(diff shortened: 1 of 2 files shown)


bigframes/ml/model_selection.py
```diff
@@ -76,7 +80,38 @@ def train_test_split(
...
+                cur,
+                test_size=test_size,
+                train_size=train_size,
+                random_state=random_state,
+            )
+            train_dfs.append(train)
+            test_dfs.append(test)
+
+        train_df = cast(
+            bpd.DataFrame, bpd.concat(train_dfs).drop(columns="bigframes_stratify_col")
```
shobsi, line 102: concat complexity was much discussed in various threads (like RAG notebook, google next demo), e.g. https://screenshot.googleplex.com/7jehfHCAVJrmc3p.  The number of unique values in the stratify col could be large to run into that. It would be a good idea to test and document where the limit lies.

shobsi (approved):

bigframes/ml/model_selection.py
GarrettWu: Yes. Concat will grow BFET and SQL size. It may encounter SQL size or OOM errors when unique values size is too large. Added a note.

PR #818, opened 2024-07-03
feat: support remote function cleanup with `session.close`

shobsi:
  - `remote_function`'s `name` param: http://screenshot/Bw8JPsmK8PvUdhA
  -  `Session.close`: http://screenshot/4X7skByF4FUFe4h
  - `bpd.clean_up_by_session_id`:  http://screenshot/BqNUF2yPWUzbCSX

Fixes internal issue 294925044 🦕


bigframes/functions/remote_function.py
```diff
@@ -67,10 +70,118 @@
...
+# Module level mapping of session-id to remote function artifacts
+_temp_session_artifacts: Dict[str, Dict[str, str]] = {}
+_session_artifacts_lock = threading.Lock()
+
+
+def _update_alive_session_artifacts(session_id: str, bqrf_routine: str, gcf_path: str):
+    """Update remote function artifacts for a session id in the current runtime."""
+    global _temp_session_artifacts, _session_artifacts_lock
+
+    with _session_artifacts_lock:
```
TrevorBergeron, line 91: why do we need this lock? why do the different sessions need to share a single dict?

shobsi: Think of this as some extra protection if the user is running multiple sessions in parallel threads (since this is a module level dict). Our own testing is a good example.

TrevorBergeron: Putting everything in one dictionary just leaves more room for errors, no? Why doesn't each sesion own its own set of artifacts, instead of a shared global super-dict?

shobsi: The reason is because the session doesn't know the artifacts created. The remote function module is kind of the manager/keeper of all things remote function. It returns a wrapper to the session and the actual artifacts are created inside the wrapper. What kind of errors do we have in mind?

TrevorBergeron: Generally, global variables and locks make code harder to test, refactor, and reason about. It seems that in this case, each session is responsible for keeping track of the resources it creates. By registering the resources directly to the session object, we can avoid unnecessary global state and locking logic.

shobsi: Agree in general, for resources directly created by the session should stay with the session and be directly managed by session. The remote function is implemented in a delegation pattern really, where the low level work stays with the remote function module, to which the session _delegates_ the creation and removal of resources. If you notice, there is no `func` argument to the `Session.remote_function` itself. The core construct that knows about the udf is the [wrapper](https://github.com/googleapis/python-bigquery-dataframes/blob/4e7e67bfe07844b1faa9763c6ab1ad7ac89ed043/bigframes/functions/remote_function.py#L931) which is too low level to belong to the Session class, so its better to keep the complexity in the remote function module, for both creation and cleanup. 

And to your main point, even if we moved some of the stuff to the Session class to make it aware of the remote function artifacts, we would still need the lock to protect against concurrent usage of the session.

Also, FWIW, cleanup is a one-off functionality, so lock can have only insignificant overhead.

Happy to chat more f2f if needed.

shobsi: Sent a patch to define a `_RemoteFunctionSession` and manage artifacts within instead of a module level dict. For convenience of review, ignore the whitespace diff https://github.blog/changelog/2021-10-14-hiding-whitespace-is-now-remembered-for-each-pull-request/

TrevorBergeron (approved):

PR #823, opened 2024-07-04
fix: Fewer relation joins from df self-operations

TrevorBergeron:


tests/system/small/test_dataframe.py
```diff
@@ -30,6 +30,7 @@
 import bigframes._config.display_options as display_options
 import bigframes.core.indexes as bf_indexes
 import bigframes.dataframe as dataframe
+import bigframes.pandas
```
Genesis929, line 33: Seems import not used and maybe duplicated.

TrevorBergeron: removed in new revision

Genesis929 (approved):

PR #824, opened 2024-07-05
fix: remove the unused connection_client from the remote function

jiaxunwu:

### DIFF
--- a/bigframes/functions/remote_function.py
+++ b/bigframes/functions/remote_function.py
@@ -53,7 +53,6 @@ import google.api_core.exceptions
 import google.api_core.retry
 from google.cloud import (
     bigquery,
-    bigquery_connection_v1,
     functions_v2,
     resourcemanager_v3,
 )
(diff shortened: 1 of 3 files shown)


bigframes/functions/remote_function.py
```diff
@@ -654,9 +653,6 @@ def remote_function(
     output_type: Optional[type] = None,
     session: Optional[Session] = None,
     bigquery_client: Optional[bigquery.Client] = None,
-    bigquery_connection_client: Optional[
```
tswast, line 657: This is a breaking change. Please detect if this is set and warn with a [FutureWarning](https://docs.python.org/3/library/exceptions.html#FutureWarning) that this argument is deprecated, instead.

tswast (changes requested):

bigframes/functions/remote_function.py
jiaxunwu: Thanks! Are we allowed to check in the breaking change? I do find it very confusing as a customer.

shobsi: I think the savior is that this is marked as deprecated for external use, so we can make a breaking change here?

shobsi (commented):
Thanks for the catch. Looks like we missed removing from here when refactoring out the connection management (for reuse in ML modules).

bigframes/functions/remote_function.py
```diff
@@ -849,15 +841,6 @@ def remote_function(
             f"{constants.FEEDBACK_LINK}"
         )
 
-    # A BigQuery connection client is required to perform BQ connection operations
```
shobsi, line 852: This is one of the 4 client params only in the internal method,

bigquery_client
bigquery_connection_client
cloud_functions_client
resource_manager_client

None of them are needed to be provided in the publically documented entry point: https://cloud.google.com/python/docs/reference/bigframes/latest/bigframes.pandas#bigframes_pandas_remote_function

Wondering what is the value of removing only one of them.

PR #827, opened 2024-07-08
deps: update to ibis-framework 9.x and newer sqlglot

tswast:
This change is updating to Ibis-framework 9.x and a newer version of SQLGLot. The Ibis upgrading also removes previous version restrictions on certain packages. Specifically, it expands the allowable version of `pyarrow` (from `15.0.2` to `17.0.0`) and `numpy` (from `1.26.4` to `2.1.1`). 
- [X] Fixes internal issue 350749011 

Fixes internal issue 350749011 🦕


tswast:
Getting 

```
pytest tests/system/small/test_dataframe.py::test_dataframe_agg_int_single_string
...
../../envs/bigframes/lib/python3.11/site-packages/sqlglot/dialects/bigquery.py:53: in <listcomp>
    exp.PropertyEQ(this=exp.to_identifier(name), expression=fld)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

name = Column(
  this=Identifier(this=col_18, quoted=True)), quoted = None, copy = True

    def to_identifier(name, quoted=None, copy=True):
        """Builds an identifier.
    
        Args:
            name: The name to turn into an identifier.
            quoted: Whether to force quote the identifier.
            copy: Whether to copy name if it's an Identifier.
    
        Returns:
            The identifier ast node.
        """
    
        if name is None:
            return None
    
        if isinstance(name, Identifier):
            identifier = maybe_copy(name, copy)
        elif isinstance(name, str):
            identifier = Identifier(
                this=name,
                quoted=not SAFE_IDENTIFIER_RE.match(name) if quoted is None else quoted,
            )
        else:
>           raise ValueError(f"Name needs to be a string or an Identifier, got: {name.__class__}")
E           ValueError: Name needs to be a string or an Identifier, got: <class 'sqlglot.expressions.Column'>

../../envs/bigframes/lib/python3.11/site-packages/sqlglot/expressions.py:6798: ValueError
```

for `DataFrame.agg` and related tests like `describe`. Might have something to do with how we're unnesting memtables. Will have to investigate further.

tswast:
Re: `FAILED tests/system/small/test_pandas.py::test_get_dummies_dataframe[kwargs2] - AssertionError: DataFrame.iloc[:, 11] (column name="time_col.11:14:34.701606") are different`

Looks like the time scalar is losing microsecond precision.

```
  COALESCE(`t0`.`time_col` = time(11, 14, 34), FALSE) AS `col_15`,
  COALESCE(`t0`.`time_col` = time(11, 41, 43), FALSE) AS `col_17`,
  COALESCE(`t0`.`time_col` = time(12, 0, 0), FALSE) AS `col_19`,
  COALESCE(`t0`.`time_col` = time(15, 16, 17), FALSE) AS `col_21`,
  COALESCE(`t0`.`time_col` = time(23, 59, 59), FALSE) AS `col_23`,
```

bigframes/ml/preprocessing.py
```diff
@@ -305,8 +305,20 @@ def _compile_to_sql(
         array_split_points = {}
         if self.strategy == "uniform":
             for column in columns:
-                min_value = X[column].min()
-                max_value = X[column].max()
+                column_min = X[column].min()
+                column_max = X[column].max()
+
+                # Use Python value rather than Numpy value to serialization.
```
TrevorBergeron, line 311: This seems a bit fragile - should the sql generator be more strict in its formatting instead?

bigframes/core/compile/scalar_op_compiler.py
```diff
@@ -995,8 +995,14 @@ def eq_nulls_match_op(
     y: ibis_types.Value,
 ):
     """Variant of eq_op where nulls match each other. Only use where dtypes are known to be same."""
-    left = x.cast(ibis_dtypes.str).fillna(ibis_types.literal("$NULL_SENTINEL$"))
-    right = y.cast(ibis_dtypes.str).fillna(ibis_types.literal("$NULL_SENTINEL$"))
+    literal = ibis_types.literal("$NULL_SENTINEL$")
+    if hasattr(x, "fill_null"):
```
TrevorBergeron, line 999: What is causing the fillna/fill_nbull divide?

tests/system/small/test_numpy.py
```diff
@@ -70,6 +70,13 @@ def test_df_ufuncs(scalars_dfs, opname):
     ).to_pandas()
     pd_result = getattr(np, opname)(scalars_pandas_df[["float64_col", "int64_col"]])
 
+    # In NumPy versions 2 and later, `np.floor` and `np.ceil` now produce integer
+    # outputs for the "int64_col" column.
+    if opname in ["floor", "ceil"] and isinstance(
+        pd_result["int64_col"].dtypes, pd.Int64Dtype
+    ):
```
TrevorBergeron, line 77: Seems like a big behavior change?! I thought numpy 2 was backwards compatible?

bigframes/core/compile/single_column.py
```diff
@@ -179,4 +179,8 @@ def value_to_join_key(value: ibis_types.Value):
     """Converts nullable values to non-null string SQL will not match null keys together - but pandas does."""
     if not value.type().is_string():
         value = value.cast(ibis_dtypes.str)
-    return value.fillna(ibis_types.literal("$NULL_SENTINEL$"))
+    return (
+        value.fill_null(ibis_types.literal("$NULL_SENTINEL$"))
+        if hasattr(value, "fill_null")
+        else value.fillna(ibis_types.literal("$NULL_SENTINEL$"))
+    )
```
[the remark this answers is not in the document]
tswast: fillna was renamed to fill_null in https://github.com/ibis-project/ibis/pull/9300 (ibis 9.1.0)

bigframes/core/compile/scalar_op_compiler.py
tswast: Ibis 9.0.0 -> Ibis 9.1.0. We need Ibis 9.0.0 for Python 3.9 support, at least until we vendor the ibis expr APIs too, which we'll want to do sometime between now and bigframes 2.0.

third_party/bigframes_vendored/ibis/backends/sql/compilers/bigquery/__init__.py
tswast, line 1: I notice a typo in this file: `visti_StringFind` should be `visit_StringFind`

bigframes/ml/preprocessing.py
tswast: +1 This is a bit scary and may indicate a bigger problem. Would adding numpy literal support to `BigQueryCompiler.visit_NonNullLiteral` allow us to avoid this?

tests/system/small/test_numpy.py
tswast: There were behavior changes. See: https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#numpy-2-migration-guide

I suspect it may relate to this change: https://numpy.org/neps/nep-0050-scalar-promotion.html#nep50

Type promotion is no longer value dependent, only dtype dependent.

third_party/bigframes_vendored/ibis/backends/sql/compilers/bigquery/__init__.py
tswast: Sent https://github.com/ibis-project/ibis/pull/10119 as this is also an error upstream.

bigframes/core/compile/single_column.py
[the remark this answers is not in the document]
chelsea-lin: Yes, `fillna` was renamed to `fill_null` in Ibis 9.1.0. Using `fillna` with this version or later will trigger a deprecation warning. However, we cann't use `fill_null` because it's not available in Ibis 9.0.0. Since Ibis 9.1.0 and later dropped support for Python 3.9, we must use Ibis 9.0.0 to maintain compatibility with Python 3.9

tests/system/small/test_dataframe.py
```diff
@@ -4568,6 +4568,9 @@ def test_recursion_limit(scalars_df_index):
     scalars_df_index.to_pandas()
 
 
+@pytest.mark.skipif(
+    reason="Skip until query complexity error can be reliably triggered"
```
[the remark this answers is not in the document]
chelsea-lin: Done.

cpcloud:
Reading through some of the review here, I would really love for y'all to chime in more proactively on the Ibis issue tracker.

I understand that contributing to Ibis isn't top priority, but it might help ease some of the pain here if we had some advance notice about any difficulty a particular change might create. I also realize that may not be knowable until you actually do the work to upgrade to a later version of Ibis.

Generally speaking, if there's a decision that would cause lots of headaches, or that we expect might, we'd like to discuss and get feedback on it.

bigframes/ml/preprocessing.py
chelsea-lin: Moved changes to the ML SQL generation resided in `ml/sql.py`. Other data core APIs tests didn't have same complains so that `BigQueryCompiler.visit_NonNullLiteral` remain unaffected.

tswast (commented):
LGTM, thanks!

chelsea-lin:
Thanks for your input, @cpcloud. We appreciate you taking the time to engage with us on this thread.
We generally aim to report issues that are relevant to the wider Ibis community directly in the issue tracker, so everyone can benefit from the discussion and resolution. However, we also recognize that some issues might be specific to our use case or temporary, and in those cases, we'll handle them internally. 
We value your insights and want to ensure we're collaborating effectively. Are there any particular issues you've observed in our usage of Ibis that you'd like to discuss further?  Your feedback helps us to better understand your priorities and identify areas for improvement.

bigframes/core/compile/single_column.py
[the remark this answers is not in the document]
TrevorBergeron: Hmm, I do hate to have hasattr checks in our code. Should we just use fillna and suppress warnings?

tests/system/small/test_numpy.py
TrevorBergeron: Should we match the new behavior then? dtype-dependent is something we can actually emulate perfectly, unlike value-dependency

TrevorBergeron (approved):

tests/system/small/test_numpy.py
tswast: > Should we match the new behavior then? dtype-dependent is something we can actually emulate perfectly, unlike value-dependency

Yes. Maybe bigframes 2.0? Could you file an issue @TrevorBergeron ?

PR #842, opened 2024-07-15
refactor: Unordered mode supports user partial orders

TrevorBergeron:


bigframes/core/blocks.py
```diff
@@ -498,7 +498,7 @@ def to_pandas(
         sampling_method: Optional[str] = None,
         random_state: Optional[int] = None,
         *,
-        ordered: Optional[bool] = None,
+        ordered: bool = True,
```
tswast, line 501: Does this mean total ordering now? That is to say, can ordered = False support partial ordering?

Let's add this info to the docstring.

bigframes/core/compile/api.py
```diff
@@ -25,38 +25,44 @@
...
-def compile_raw(
-    node: bigframes.core.nodes.BigFrameNode,
-) -> Tuple[str, bigframes.core.ordering.TotalOrdering]:
-    """Compile node into sql that exposes all columns, including hidden ordering-only columns."""
-    ir = _STRICT_COMPILER.compile_ordered_ir(node)
-    sql = ir.raw_sql()
-    ordering_info = ir._ordering
-    assert ir.has_total_order
-    return sql, ordering_info  # type: ignore
+class SqlCompiler:
```
tswast, line 28: Nit: `SQLCompiler`

> When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

https://peps.python.org/pep-0008/#descriptive-naming-styles

tswast (commented):
Thanks! Looks good. A few nits.

bigframes/core/blocks.py
TrevorBergeron: Added docstring and indicate that ordered=True will sort rows, but whether that is total or not depends on session setting.

ordered=False will always be completely unordered.

bigframes/core/compile/api.py
TrevorBergeron: renamed

tswast (approved):

PR #844, opened 2024-07-16
fix: Fix 'sql' property for null index

TrevorBergeron:


tests/system/small/test_unordered.py
```diff
@@ -20,6 +20,15 @@
 from tests.system.utils import assert_pandas_df_equal, skip_legacy_pandas
 
 
+def test_unordered_mode_sql_no_hash(unordered_session):
+    bf_df = unordered_session.read_gbq(
+        "bigquery-public-data.ethereum_blockchain.blocks"
+    )
+    sql = bf_df.sql
+    assert "ORDER BY" not in sql.upper()
+    assert "farm_fingerprint" not in sql
```
shobsi, line 29: stronger assert with `sql.lower()`, or even `"FARM_FINGERPRINT" not in sql.upper()` for consistency with the previous line?

shobsi (approved):
left a minor comment

PR #851, opened 2024-07-22
feat: `df.apply(axis=1)` to support remote function with mutiple params

shobsi:


bigframes/dataframe.py
```diff
@@ -3432,65 +3432,95 @@ def apply(self, func, *, axis=0, args: typing.Tuple = (), **kwargs):
...
-            supported_arrow_types = tuple(
-                dtype.pyarrow_dtype
-                for dtype in remote_function_supported_dtypes
-                if isinstance(dtype, pandas.ArrowDtype)
-            )
-            supported_dtypes_hints = tuple(
-                str(dtype) for dtype in remote_function_supported_dtypes
-            )
+            # Check if the function is a remote function
+            if not hasattr(func, "bigframes_remote_function"):
```
GarrettWu, line 3436: Is it possible to check "isinstance"? Why use the attribute to determine the type?

GarrettWu (commented):
Do we have docs to update?

bigframes/dataframe.py
shobsi: the user defined function `func` is of `function` class, we just set extra properties on it when we deploy a remote function from it. We check one of those properties to confirm that it is indeed a remote function and not a plain python function.

GarrettWu (approved):

bigframes/dataframe.py
```diff
@@ -3446,65 +3446,96 @@ def apply(self, func, *, axis=0, args: typing.Tuple = (), **kwargs):
...
+                )
+            else:
+                # This is a special case where we are providing not-pandas-like
+                # extension. If the remote function can take one or more params
+                # then we assume that here the user intention is to use the
+                # column values of the dataframe as arguments to the function.
+                # For this to work the following condition must be true:
+                #   1. The number or input params in the function must be same
+                #      as the number of columns in the dataframe
+                #   2. The dtypes of the columns in the dataframe must be
```
TrevorBergeron, line 3518: Do we want to accept compatible dtypes? eg the column is int, but the function takes decimal?

bigframes/dataframe.py
```diff
@@ -3446,65 +3446,96 @@ def apply(self, func, *, axis=0, args: typing.Tuple = (), **kwargs):
...
+                    )
+                if udf_input_dtypes != tuple(self.dtypes.to_list()):
+                    raise ValueError(
+                        f"Remote function takes arguments of types {udf_input_dtypes} but DataFrame dtypes are {tuple(self.dtypes)}."
+                    )
+
+                series_list = [self[col] for col in self.columns]
+                reprojected_series = bigframes.series.Series(
+                    series_list[0]._block._force_reproject()
+                )
```
TrevorBergeron, line 3535: wyh do we need a force_reproject?

bigframes/core/compile/scalar_op_compiler.py
```diff
@@ -1463,6 +1472,19 @@ def case_when_op(*cases_and_outputs: ibis_types.Value) -> ibis_types.Value:
     return case_val.end()
 
 
+@scalar_op_compiler.register_nary_op(ops.NaryRemoteFunctionOp, pass_op=True)
+def nary_remote_function_op_impl(
+    *operands: ibis_types.Value, op: ops.NaryRemoteFunctionOp
+):
+    ibis_node = getattr(op.func, "ibis_node", None)
```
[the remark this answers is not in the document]
shobsi: Ack.

bigframes/dataframe.py
shobsi: function taking decimal is not something we support right now

https://github.com/googleapis/python-bigquery-dataframes/blob/a7d7197a32c55b989ae4ea8f6cf6e1c0f7184cd4/bigframes/dtypes.py#L618

shobsi: Having said that, are there other places where we reconcile dtypes across dataframes or in operations?

bigframes/dataframe.py
shobsi: I am just copy-pasting the pattern introduced in this change: https://github.com/googleapis/python-bigquery-dataframes/commit/ff3bb89cca2e29625166b4646a34ce6b889d58f9#diff-8718ceb6a8f6b68d7b06a15e84043fb866c500d5bfb1f33ad8c945f06815a140

Is the reasoning (got a bit detached unintentionally, sitting at the beginning of the function) still valid?

```python
# Reproject as workaround to applying filter too late. This forces the filter
# to be applied before passing data to remote function, protecting from bad
# inputs causing errors.
```

shobsi: Actually it does make a difference, quickly tested in https://github.com/googleapis/python-bigquery-dataframes/pull/874 and `series.mask` doctest is failing

shobsi: I moved the comment back close to reproject, PTAL

TrevorBergeron (approved):

shobsi:
> Do we have docs to update?

The code samples added should suffice for this change.

PR #852, opened 2024-07-22
test: fix mypy failures to unblock presubmit tests

shobsi:
The presubmit and continuous tests have started failing due to mypy. This change finds the failure points and puts in workarounds.

### DIFF
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -650,15 +650,17 @@ class Session(
 
         index_cols = _to_index_cols(index_col)
 
-        filters = list(filters)
-        if len(filters) != 0 or max_results is not None:
+        filters_copy1, filters_copy2 = itertools.tee(filters)
+        has_filters = len(list(filters_copy1)) != 0
+        filters = typing.cast(third_party_pandas_gbq.FiltersType, filters_copy2)
+        if has_filters or max_results is not None:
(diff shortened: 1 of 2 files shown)


bigframes/session/__init__.py
```diff
@@ -768,7 +770,7 @@ def _read_gbq_table(
         )
 
         columns = list(columns)
-        filters = list(filters)
+        filters = typing.cast(list, list(filters))
```
Genesis929, line 773: I'm curious why we need to cast to list here?

Genesis929 (approved):

bigframes/session/__init__.py
shobsi: otherwise mypy complains:

```
bigframes/session/__init__.py:773: error: Incompatible types in assignment (expression has type "list[Iterable[tuple[str, Literal['in', 'not in', '<', '<=', '==', '!=', '>=', '>', 'LIKE'], Any] | str | Any]]", variable has type "Iterable[tuple[str, Literal['in', 'not in', '<', '<=', '==', '!=', '>=', '>', 'LIKE'], Any]] | Iterable[Iterable[tuple[str, Literal['in', 'not in', '<', '<=', '==', '!=', '>=', '>', 'LIKE'], Any]]]")  [assignment]
```

PR #856, opened 2024-07-24
fix: reduce redundant `remote_function` deployments

shobsi:
See the demo at screencast/cast/NTA0OTk2MTc5MTA5NDc4NHwwYjBjNDBiMS01Mw

Fixes internal issue #349872122 🦕

### DIFF
--- a/bigframes/functions/remote_function.py
+++ b/bigframes/functions/remote_function.py
@@ -893,11 +893,16 @@ class _RemoteFunctionSession:
                 dynamically using the `bigquery_connection_client` assuming the user has necessary
                 priviliges. The PROJECT_ID should be the same as the BigQuery connection project.
             reuse (bool, Optional):
-                Reuse the remote function if is already exists.
-                `True` by default, which results in reusing an existing remote
+                Reuse the remote function if already exists.
+                `True` by default, which will result in reusing an existing remote
                 function and corresponding cloud function (if any) that was
                 previously created for the same udf.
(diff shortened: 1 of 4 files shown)


bigframes/functions/remote_function.py
```diff
@@ -167,7 +167,23 @@ def get_remote_function_locations(bq_location):
...
+    # A workaround of the same can be achieved by replacing the filename in the
+    # code object to a static value
+    # https://github.com/cloudpipe/cloudpickle/issues/120#issuecomment-338510661.
+    #
+    # To respect the user code/environment let's make this modification on a
+    # copy of the udf, not on the original udf itself.
+    def_copy = cloudpickle.loads(cloudpickle.dumps(def_))
+    def_copy.__code__ = def_copy.__code__.replace(
+        co_filename="bigframes_place_holder_filename"
+    )
```
TrevorBergeron, line 184: Is there a more efficient way to do this? Also do we want to replace filename for other code objects in the dependency tree? Seems the most efficient solution would require modifying cloudpickle a bit though

shobsi: We are not doing dependency tree yet, I'm looking into the cloudpickle (and the native pickle which it builds on top of) to evaluate what modification we can make. Consider this change as an instant work around to reduce re-deployments in two scenarios:

1. rerun of a cell in the same session
2. rerun of a notebook with explicitly named remote function

TrevorBergeron (approved):

PR #860, opened 2024-07-24
feat: Allow DataFrame.join for self-join on Null index

TrevorBergeron:

### DIFF
--- a/bigframes/ml/core.py
+++ b/bigframes/ml/core.py
@@ -83,7 +83,7 @@ class BaseBqml:
         """
         assert len(x.columns) == 1 and len(y.columns) == 1
 
-        input_data = x.cache().join(y.cache(), how="outer")
+        input_data = x.join(y, how="outer").cache()
         x_column_id, y_column_id = x._block.value_columns[0], y._block.value_columns[0]
 
         return self._apply_sql(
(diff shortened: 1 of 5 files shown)


tswast (commented):
Thanks Trevor!

Should we add an e2e test for the .fit(X, y) table with null index / unordered mode, too?

TrevorBergeron:
> Thanks Trevor!
> 
> Should we add an e2e test for the .fit(X, y) table with null index / unordered mode, too?

Good idea. This revealed that ml modules were caching pre-join, which is invalidated by row-identity join. Instead, I made it cache post-join.

bigframes/ml/core.py
```diff
@@ -326,7 +326,7 @@ def create_model(
         if y_train is None:
             input_data = X_train.cache()
         else:
-            input_data = X_train.cache().join(y_train.cache(), how="outer")
+            input_data = X_train.join(y_train.cache(), how="outer").cache()
```
GarrettWu, line 329: this should be y_train without cache() as well?

tests/system/large/ml/test_linear_model.py
```diff
@@ -150,6 +150,44 @@ def test_logistic_regression_configure_fit_score(penguins_df_default_index, data
     assert reloaded_model.class_weight is None
 
 
+def test_unordered_mode_logistic_regression_configure_fit_score(
+    unordered_session, penguins_table_id, dataset_id
+):
+    model = bigframes.ml.linear_model.LogisticRegression()
```
GarrettWu, line 156: Not a problem, but if we only pick one model to test some shared functionalities, usually the choice is LinearReg.

TrevorBergeron: ok, used linear regression model instead

bigframes/ml/core.py
TrevorBergeron: yes, fixed

tswast (approved):

PR #861, opened 2024-07-25
feat: Allow windowing in 'partial' ordering mode

TrevorBergeron:


bigframes/session/__init__.py
```diff
@@ -377,6 +378,10 @@ def slot_millis_sum(self):
         """The sum of all slot time used by bigquery jobs in this session."""
         return self._slot_millis_sum
 
+    @property
+    def allow_ambiguity(self) -> bool:
```
tswast, line 382: Does this need to be a public property? I'd prefer to keep it private until we have a request from users to modifying it.

Also, I'm not sold on the naming. Should it be specific and mention ambiguous orderings? Or rename to allow all forms of non-determinism?

TrevorBergeron: Yes, no reason to make this a public property. This is specifically for use in windowing - where ambiguous orderings produce ambiguous row values. Ordering ambiguity is introduced through "partial" ordering mode already.

bigframes/core/__init__.py
```diff
@@ -194,8 +194,15 @@ def promote_offsets(self, col_id: str) -> ArrayValue:
         """
         Convenience function to promote copy of column offsets to a value column. Can be used to reset index.
         """
-        if self.node.order_ambiguous and not self.session._strictly_ordered:
-            raise ValueError("Generating offsets not supported in unordered mode")
+        if self.node.order_ambiguous and not (self.session._strictly_ordered):
+            if not self.session._allows_ambiguity:
+                raise ValueError("Generating offsets not supported in unordered mode")
```
tswast, line 199: Nit: partial ordered mode

Let's try to be consistent with our naming of the feature.

bigframes/core/__init__.py
```diff
@@ -347,9 +354,16 @@ def project_window_op(
         # TODO: Support non-deterministic windowing
         if window_spec.row_bounded or not op.order_independent:
             if self.node.order_ambiguous and not self.session._strictly_ordered:
-                raise ValueError(
-                    "Order-dependent windowed ops not supported in unordered mode"
-                )
+                if not self.session._allows_ambiguity:
+                    raise ValueError(
+                        "Generating offsets not supported in unordered mode"
```
tswast, line 359: Nit: partial ordered mode

tswast (approved):

PR #868, opened 2024-07-29
feat: bigframes.bigquery.json_extract

chelsea-lin:
BEGIN_COMMIT_OVERRIDE
feat: Implement `bigframes.bigquery.json_extract` (#868)
END_COMMIT_OVERRIDE

- [X] Make sure to open an issue: Internal issue 346571659

Fixes internal bug: 346571659 🦕

### DIFF
--- a/bigframes/bigquery/__init__.py
+++ b/bigframes/bigquery/__init__.py
@@ -208,6 +208,41 @@ def json_set(
     return series
 
 
+def json_extract(
+    series: series.Series,
+    json_path: str,
+) -> series.Series:
+    """Extracts a JSON value and converts it to a SQL JSON-formatted `STRING` or `JSON`
+    value. This function uses single quotes and brackets to escape invalid JSONPath
+    characters in JSON keys.
+
+    **Examples:**
+
+        >>> import bigframes.pandas as bpd
+        >>> import bigframes.bigquery as bbq
+        >>> bpd.options.display.progress_bar = None
+
+        >>> s = bpd.Series(['{"class": {"students": [{"id": 5}, {"id": 12}]}}'])
+        >>> bbq.json_extract(s, json_path="$.class")
+        0    "{\\\"students\\\":[{\\\"id\\\":5},{\\\"id\\\":12}]}"
+        dtype: string
+
+    Args:
+        series (bigframes.series.Series):
+            The Series containing JSON data (as native JSON objects or JSON-formatted strings).
+        json_path (str):
+            The JSON path identifying the data that you want to obtain from the input.
+
+    Returns:
+        bigframes.series.Series: A new Series with the JSON or JSON-formatted STRING.
+    """
+    return series._apply_unary_op(ops.JSONExtract(json_path=json_path))
+
+
+# Search functions defined from
+# https://cloud.google.com/bigquery/docs/reference/standard-sql/search_functions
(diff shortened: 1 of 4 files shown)


tests/system/small/bigquery/test_json.py
```diff
@@ -110,3 +110,30 @@ def test_json_set_w_invalid_value_type():
...
+
+
+def test_json_extract_from_string():
+    s = bpd.Series(['{"a": {"b": [1, 2]}}', '{"a": {"c": 1}}', '{"a": {"b": 0}}'])
+    actual = bbq.json_extract(s, "$.a.b")
+    expected = _get_series_from_json(["[1,2]", None, "0"])
+    pd.testing.assert_series_equal(
+        actual.to_pandas(),
+        expected.to_pandas(),
+        check_names=False,
```
shobsi, line 133: why do we need this here but not in `test_json_extract_from_json`?

bigframes/operations/__init__.py
```diff
@@ -602,6 +602,22 @@ def output_type(self, *input_types):
         return dtypes.STRING_DTYPE
 
 
+## JSON Ops
+@dataclasses.dataclass(frozen=True)
+class JSONExtract(UnaryOp):
+    name: typing.ClassVar[str] = "json_extract"
+    json_path: str
+
+    def output_type(self, *input_types):
```
shobsi, line 611: what would be a good example of multiple input_types? Can they be different?

bigframes/bigquery/__init__.py
```diff
@@ -208,6 +208,41 @@ def json_set(
...
+            The JSON path identifying the data that you want to obtain from the input.
+
+    Returns:
+        bigframes.series.Series: A new Series with the JSON or JSON-formatted STRING.
+    """
+    return series._apply_unary_op(ops.JSONExtract(json_path=json_path))
+
+
+# Search functions defined from
+# https://cloud.google.com/bigquery/docs/reference/standard-sql/search_functions
```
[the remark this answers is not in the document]
chelsea-lin: Yes, added it for `vector_search` to align with other methods.

tests/system/small/bigquery/test_json.py
chelsea-lin: Thanks for looking the details. While the `dbjson` extension dtype is missing (working on [PR](https://github.com/googleapis/python-db-dtypes-pandas/pull/284)), we calls `_get_series_from_json` to return a JSON column from a BigQuery DataFrames. So the returned Series has a "data" name. That's why we have to ignore "check_names" in this test.

bigframes/operations/__init__.py
chelsea-lin: `UnaryOp` indicates that we only have one input. Similarly, `BinaryOp` can have two inputs. More examples can be found in this file.

tests/system/small/bigquery/test_json.py
shobsi: We are calling `_get_series_from_json` in both the tests, but using check_names=False only in this one, wondering why the difference

shobsi: Never mind, its because of the difference in `s = ...`

shobsi (approved):

PR #877, opened 2024-08-01
perf: Generate SQL with fewer CTEs

TrevorBergeron:


bigframes/core/compile/compiled.py
```diff
@@ -355,6 +346,9 @@ def _to_ibis_expr(
         return table
 
     def filter(self, predicate: ex.Expression) -> UnorderedIR:
+        if any(map(is_window, map(self._get_ibis_column, predicate.unbound_variables))):
```
tswast, line 349: Is this a fix for something? If so, maybe worth a mention in the changelog and corresponding test?

bigframes/core/compile/compiled.py
```diff
@@ -785,15 +779,32 @@ def promote_offsets(self, col_id: str) -> OrderedIR:
...
+            expr_builder = self.builder()
+            expr_builder.columns = [
+                self._compile_expression(
+                    ordering.total_order_col.scalar_expression
+                ).name(col_id),
+                *self.columns,
+            ]
+            return expr_builder.build()
+        # Cannot nest analytic expressions, so reproject to cte first if needed.
+        # Also ibis cannot window literals, so need to reproject those (even though this is legal in googlesql)
```
tswast, line 793: Maybe worth a bug upstream in ibis?

bigframes/core/compile/compiled.py
```diff
@@ -815,8 +825,24 @@ def project_window_op(
         window_spec: a specification of the window over which to apply the operator
         output_name: the id to assign to the output of the operator, by default will replace input col if distinct output id not provided
         never_skip_nulls: will disable null skipping for operators that would otherwise do so
-        skip_reproject_unsafe: skips the reprojection step, can be used when performing many non-dependent window operations, user responsible for not nesting window expressions, or using outputs as join, filter or aggregation keys before a reprojection
         """
+        # Cannot nest analytic expressions, so reproject to cte first if needed.
+        # Also ibis cannot window literals, so need to reproject those (even though this is legal in googlesql)
```
tswast, line 830: Same here. Let's link to an ibis bug.

bigframes/core/compile/compiled.py
```diff
@@ -1034,6 +1057,9 @@ def _to_ibis_expr(
         return table
 
     def filter(self, predicate: ex.Expression) -> OrderedIR:
+        if any(map(is_window, map(self._get_ibis_column, predicate.unbound_variables))):
+            # ibis doesn't support qualify syntax
```
tswast, line 1061: Might be worth an issue, too. I think there is some support in ibis for `having`, for example (though it's not really possible with the pandas API) if I remember correctly from my work on F1. https://ibis-project.org/tutorials/ibis-for-sql-users#having-clause

TrevorBergeron: created bug and referenced in comment

bigframes/core/compile/compiled.py
TrevorBergeron: created bug and referenced in comment

bigframes/core/compile/compiled.py
TrevorBergeron: created bug and referenced in comment

bigframes/core/compile/compiled.py
TrevorBergeron: This is not a fix, previously we would proactively reproject after windowing. With this new implementation, instead, we are maximally deferring, so need to check later on wherever windowed expressions are not allowed.

tswast (approved):

PR #884, opened 2024-08-06
docs: add streaming html docs

GarrettWu:
Verified locally.

### DIFF
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -760,7 +760,9 @@ class Session(
     ) -> streaming_dataframe.StreamingDataFrame:
         """Turn a BigQuery table into a StreamingDataFrame.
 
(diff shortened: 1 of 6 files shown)


jiaxunwu (approved):
for streaming.dataframe, could you add notes about what is supported there? for example, you can simply list the operations allowed at this moment.

chelsea-lin (approved):

chelsea-lin (approved):
Can you please link the staging doc with this change in the PR descriptions, if you have

GarrettWu:
> Can you please link the staging doc with this change in the PR descriptions, if you have

https://cloud.devsite.corp.google.com/python/docs/reference/bigframes/latest

GarrettWu:
> for streaming.dataframe, could you add notes about what is supported there? for example, you can simply list the operations allowed at this moment.

done

PR #892, opened 2024-08-08
docs: add columns for "requires ordering/index" to supported APIs summary

tswast:


TrevorBergeron:
There are a few APIs for which order- or index-dependence is conditional on the parameters passed. Presently, these apis have conditional statements inside the body instead of using the decorator. One such case is drop_duplicates. Not sure how we would want to label these APIs in the docs.

Genesis929 (approved):

PR #893, opened 2024-08-08
docs: fix the `DisplayOptions` doc rendering

shobsi:
- [x] Appropriate docs were updated (if necessary): **Before:** screenshot/9PL6mipmHHtg4x2, **After:** screenshot/BDLmQTTX8REjBpD

Fixes internal issue 358267007 🦕

### DIFF
--- a/third_party/bigframes_vendored/pandas/core/config_init.py
+++ b/third_party/bigframes_vendored/pandas/core/config_init.py
@@ -18,6 +18,7 @@ Encapsulates the configuration for displaying objects.
 **Examples:**
 
 Define Repr mode to "deferred" will prevent job execution in repr.
+
     >>> import bigframes.pandas as bpd
     >>> df = bpd.read_gbq("bigquery-public-data.ml_datasets.penguins")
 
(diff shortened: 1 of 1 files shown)


third_party/bigframes_vendored/pandas/core/config_init.py
```diff
@@ -26,25 +27,32 @@
     Computation deferred. Computation will process 28.9 kB
 
 Users can also get a dry run of the job by accessing the query_job property before they've run the job. This will return a dry run instance of the job they can inspect.
+
     >>> df.query_job.total_bytes_processed
     28947
 
 User can execute the job by calling .to_pandas()
+
     >>> # df.to_pandas()
```
chelsea-lin, line 36: should we remove `# ` here?

chelsea-lin (approved):
Left a nit comment. Overall LGTM.

third_party/bigframes_vendored/pandas/core/config_init.py
shobsi: I think let's keep it, my guess is that it is there to save downloading data to let the doctest run faster

PR #894, opened 2024-08-08
feat: support `Index.to_frame()`

mattyopl:
Intern Starter Task

can test by checking:

```
import pandas as pd
idx = pd.Index(['Ant', 'Bear', 'Cow'], name='animal')
df = idx.to_frame()
print(df)
df = idx.to_frame(False)
print(df)
df = idx.to_frame(name = "food")
print(df)
df = idx.to_frame(False, name = "food")
print(df)

import bigframes.core as core
idx = core.indexes.Index(['Ant', 'Bear', 'Cow'], name='animal')
df = idx.to_frame()
print(df)
df = idx.to_frame(False)
print(df)
df = idx.to_frame(name = "food")
print(df)
df = idx.to_frame(False, name = "food")
print(df)
```

Fixes internal 356891401 🦕


bigframes/core/indexes/base.py
```diff
@@ -114,6 +114,32 @@ def from_frame(
...
         index._linked_frame = frame
         return index
+    
+    def to_frame(
+        self, index: bool = True, name: blocks.Label | None = None
+    ) -> bigframes.dataframe.DataFrame:
+        import numpy as np
+        provided_name = name if name else self.name
+        provided_index = self.values if index else np.arange(len(self.values))
+        result = bigframes.dataframe.DataFrame({provided_name: self.values}, index= provided_index)
```
TrevorBergeron, line 124: If the index is a MultiIndex, the resulting DataFrame will need to have multiple columns (and names will need to be a list-like with length equal to the multi-index level depth).

bigframes/core/indexes/base.py
```diff
@@ -114,6 +114,32 @@ def from_frame(
         index = Index(frame._block)
         index._linked_frame = frame
         return index
+    
+    def to_frame(
+        self, index: bool = True, name: blocks.Label | None = None
+    ) -> bigframes.dataframe.DataFrame:
+        import numpy as np
+        provided_name = name if name else self.name
+        provided_index = self.values if index else np.arange(len(self.values))
```
TrevorBergeron, line 123: Just provide `None` if want default sequential index. np.arrange may create a very large in-memory array here if the original index is large (some BigFrames objects have billions of rows).

bigframes/core/indexes/base.py
```diff
@@ -115,6 +115,20 @@ def from_frame(
         index._linked_frame = frame
         return index
 
+    def to_frame(
+        self, index: bool = True, name: blocks.Label | None = None
+    ) -> bigframes.dataframe.DataFrame:
+        import numpy as np
```
GarrettWu, line 121: Put the import at the top of the file for easier management. Actually numpy is already imported in the file.

Import in the method only to avoid circular imports.

bigframes/core/indexes/base.py
```diff
@@ -115,6 +115,20 @@ def from_frame(
         index._linked_frame = frame
         return index
 
+    def to_frame(
```
GarrettWu, line 118: Plz add tests.

mattyopl: done in tests/unit/core/test_indexes.py :D

bigframes/core/indexes/base.py
```diff
@@ -115,6 +115,34 @@ def from_frame(
...
         return index
 
+    def to_frame(self, index: bool = True, name=None) -> bigframes.dataframe.DataFrame:
+        from bigframes.core.indexes.multi import MultiIndex
+
+        multi = isinstance(self, MultiIndex)
+
+        if multi:
+            columns = [
+                [self.values[j][i] for j in range(len(self.values))]
```
GarrettWu, line 125: self.values calls to_numpy, which downloads everything and not scalable. Maybe take a look at Index.from_frame() and Series.to_frame()

tests/system/small/test_index.py
```diff
@@ -320,6 +323,33 @@ def test_index_to_series(
...
 
+def test_index_to_frame():
+    index_to_frame_index_args = [True, False]
+    index_to_frame_name_args = [None, "food"]
+    pd_idx: pd.Index = pd.Index(
+        ["Ant", "Bear", "Cow"], name="animal", dtype="string[pyarrow]"
+    )
+    bf_idx = indexes.Index(["Ant", "Bear", "Cow"], name="animal")
+
+    for index_arg, name_arg in itertools.product(
```
GarrettWu, line 334: you can use @pytest.mark.parametrize to achieve combinations of params.

Which creates multiple separate tests. Easier to find the problem if anyone fails.

bigframes/core/indexes/base.py
```diff
@@ -115,6 +115,34 @@ def from_frame(
         index._linked_frame = frame
         return index
 
+    def to_frame(self, index: bool = True, name=None) -> bigframes.dataframe.DataFrame:
```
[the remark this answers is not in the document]
mattyopl: added back, still getting familiar with static type checker in Python

bigframes/core/indexes/base.py
mattyopl: fixed for `Index`. per offline conversation with @TrevorBergeron, efficient `MultiIndex` implementation is blocked by bugs in `DataFrame` construction with `MultiIndex`

bug TL;DR

1) Index with array values being created for DataFrame instead of MultiIndex
```
bf_idx = indexes.MultiIndex.from_arrays([["a", "b", "c"], ["d", "e", "f"]])
test = bigframes.dataframe.DataFrame({1: [2, 4, 6], 2: [1, 3, 5]}, index = bf_idx)
print(test)
```
2) block joining for DataFrame creation using MultiIndex and Dict of Series data throws
```
bf_idx = indexes.MultiIndex.from_arrays([["a", "b", "c"], ["d", "e", "f"]])
nlevels = bf_idx.nlevels
columns : list[bigframes.series.Series] = []
 for level in range(nlevels):
       series = self.get_level_values(level).to_series()
       columns.append(series)
data = {i: column for i, column in enumerate(columns)}
result = bigframes.dataframe.DataFrame(
            data, index= bf_idx
        )
```

tests/system/small/test_index.py
```diff
@@ -320,6 +321,30 @@ def test_index_to_series(
...
 
+@pytest.mark.parametrize("index_arg", [True, False])
+@pytest.mark.parametrize("name_arg", [None, "food"])
+def test_index_to_frame(index_arg, name_arg):
+    pd_idx: pd.Index = pd.Index(
+        ["Ant", "Bear", "Cow"], name="animal", dtype="string[pyarrow]"
+    )
+    bf_idx = indexes.Index(["Ant", "Bear", "Cow"], name="animal")
+
+    if name_arg is None:
```
GarrettWu, line 332: It doesn't need the if else condition. Just

```
pd_df = pd_idx.to_frame(index=index_arg, name=name_arg)
bf_df = bf_idx.to_frame(index=index_arg, name=name_arg)
```

is good enough

tests/system/small/test_multiindex.py
```diff
@@ -45,6 +46,27 @@ def test_multi_index_from_arrays():
     pandas.testing.assert_index_equal(bf_idx.to_pandas(), pd_idx)
 
 
+@pytest.mark.parametrize("index_arg", [True, False])
+@pytest.mark.parametrize("name_arg", [None, ["x", "y"]])
+def test_multi_index_to_frame(index_arg, name_arg):
+
+    pd_idx = pandas.MultiIndex.from_arrays([["a", "b", "c"], ["d", "e", "f"]])
+    bf_idx = indexes.MultiIndex.from_arrays([["a", "b", "c"], ["d", "e", "f"]])
+    if name_arg is None:
```
GarrettWu, line 55: same here.

GarrettWu (commented):
Generally LGTM.

tests/system/small/test_index.py
mattyopl: Pandas implementation of `name` arg handling is slightly different. If we set `name=None` for pandas, it will create a DataFrame with column names as string `None`, since technically the default is `lib.nodefault` (code pointer: https://github.com/pandas-dev/pandas/blob/v2.2.2/pandas/_libs/lib.pyi#L32), whereas use `None`: https://github.com/pandas-dev/pandas/blob/v2.2.2/pandas/core/indexes/base.py#L1607-L1666

mattyopl: Happy to add something similar to BigFrames if you think it is better to exactly mirror pandas functionality

GarrettWu: Ah, that's a weird behavior of pandas. Thanks for pointing out. Then we shouldn't let the default to be None. But should have sth similar. @TrevorBergeron

PR #899, opened 2024-08-14
refactor: Extract DataFrame execution to class

TrevorBergeron:

### DIFF
--- a/bigframes/session/_io/bigquery/__init__.py
+++ b/bigframes/session/_io/bigquery/__init__.py
@@ -18,7 +18,6 @@ from __future__ import annotations
 
 import datetime
 import itertools
-import os
 import re
 import textwrap
 import types
@@ -35,14 +34,17 @@ from bigframes.core import log_adapter
 import bigframes.core.compile.googlesql as googlesql
 import bigframes.core.sql
 import bigframes.formatting_helpers as formatting_helpers
+import bigframes.session.metrics
+
+CHECK_DRIVE_PERMISSIONS = "\nCheck https://cloud.google.com/bigquery/docs/query-drive-data#Google_Drive_permissions."
+
 
 IO_ORDERING_ID = "bqdf_row_nums"
 MAX_LABELS_COUNT = 64
 _LIST_TABLES_LIMIT = 10000  # calls to bqclient.list_tables
 # will be limited to this many tables
 
-LOGGING_NAME_ENV_VAR = "BIGFRAMES_PERFORMANCE_LOG_NAME"
-CHECK_DRIVE_PERMISSIONS = "\nCheck https://cloud.google.com/bigquery/docs/query-drive-data#Google_Drive_permissions."
+_MAX_CLUSTER_COLUMNS = 4
 
 
 def create_job_configs_labels(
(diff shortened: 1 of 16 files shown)


bigframes/session/executor.py
```diff
@@ -0,0 +1,329 @@
...
+
+# Max complexity that should be executed as a single query
+QUERY_COMPLEXITY_LIMIT = 1e7
+# Number of times to factor out subqueries before giving up.
+MAX_SUBTREE_FACTORINGS = 5
+
+_MAX_CLUSTER_COLUMNS = 4
+
+
+class BigQueryCachingExecutor:
```
chelsea-lin, line 46: Could you please add some comments to briefly explain how the cached executions works.

bigframes/session/executor.py
```diff
@@ -0,0 +1,329 @@
...
+        self.compiler: bigframes.core.compile.SQLCompiler = (
+            bigframes.core.compile.SQLCompiler(strict=strictly_ordered)
+        )
+        self.strictly_ordered: bool = strictly_ordered
+        self._cached_executions: weakref.WeakKeyDictionary[
+            nodes.BigFrameNode, nodes.BigFrameNode
+        ] = weakref.WeakKeyDictionary()
+        self.metrics = metrics
+
+    def to_sql(
```
chelsea-lin, line 70: Could you please explain the functionality of the public methods `to_sql`, `execute`, and `export` in more detail? The name export isn't immediately clear to me, so a more comprehensive explanation would be helpful.

bigframes/session/executor.py
```diff
@@ -0,0 +1,329 @@
...
+    def _run_execute_query(
+        self,
+        sql: str,
+        job_config: Optional[bq_job.QueryJobConfig] = None,
+        api_name: Optional[str] = None,
+    ) -> Tuple[bigquery.table.RowIterator, bigquery.QueryJob]:
+        """
+        Starts BigQuery query job and waits for results.
+        """
+        job_config = bigquery.QueryJobConfig()
```
chelsea-lin, line 172: The `job_config` overwrites the one in the function parameter. Could you double check if it's in purpose.

bigframes/session/_io/bigquery/__init__.py
```diff
@@ -222,17 +223,17 @@ def add_labels(job_config, api_name: Optional[str] = None):
 
 
 def start_query_with_client(
```
chelsea-lin, line 225: Can we make the `Session` as a class and then `bq_client` and `metrics` can be part of it?

chelsea-lin (changes requested):

bigframes/session/_io/bigquery/__init__.py
TrevorBergeron: Would rather pass more precise context, this makes it more flexible for use in other places, and reduces circular dependency between files.

bigframes/session/executor.py
TrevorBergeron: yeah, we don't want that. fixed in new revision

bigframes/session/executor.py
TrevorBergeron: Added a little docstring to each

bigframes/session/executor.py
TrevorBergeron: Added a bit of an explantation to class description.

chelsea-lin (approved):
Thanks for improving code readability consistently.

PR #913, opened 2024-08-22
refactor: Extract data loading logic into class

TrevorBergeron:


GarrettWu (commented):
What a large PR. Do we have a bug number or any other contexts? What are we trying to solve here?

TrevorBergeron:
> What a large PR. Do we have a bug number or any other contexts? What are we trying to solve here?

I do not have a bug number. Here is a previous commit in this effort: https://github.com/googleapis/python-bigquery-dataframes/commit/92fdb937220146f770c4ce32f8317eb55ece434d. 

The goal has been to make Session easier to understand by splitting it's state and functionality into distinct components. Mostly just moving code to new files right now. The hope is by creating some smaller components, it will be easier to optimize each component. Additionally, I do want to try making session thread-safe at some point, and pushing session state into individual objects will help with setting up necessary locks.

bigframes/session/loader.py
```diff
@@ -0,0 +1,641 @@
...
+    def __init__(
+        self,
+        session: bigframes.session.Session,
+        bqclient: bigquery.Client,
+        storage_manager: bigframes.session.temp_storage.TemporaryGbqStorageManager,
+        default_index_type: bigframes.enums.DefaultIndexKind,
+        scan_index_uniqueness: bool,
+        metrics: Optional[bigframes.session.metrics.ExecutionMetrics] = None,
+    ):
+        self.bqclient = bqclient
```
GarrettWu, line 103: Some members are "private" while others are public. Are they really different or they can be consistent?

bigframes/session/loader.py
```diff
@@ -0,0 +1,641 @@
...
+    This loader is constrained to loading local data and queries against data sources in the same region as the storage manager.
+    """
+
+    def __init__(
+        self,
+        session: bigframes.session.Session,
+        bqclient: bigquery.Client,
+        storage_manager: bigframes.session.temp_storage.TemporaryGbqStorageManager,
+        default_index_type: bigframes.enums.DefaultIndexKind,
+        scan_index_uniqueness: bool,
```
GarrettWu, line 100: Can we add docs for the args? At least scan_index_uniqueness and metrics aren't clear at first glance.

bigframes/session/executor.py
```diff
@@ -147,6 +157,35 @@ def export(
             job_config=job_config,
         )
 
+    def export_gcs(
+        self,
+        array_value: bigframes.core.ArrayValue,
+        col_id_overrides: Mapping[str, str],
+        uri: str,
+        format: Literal["JSON", "CSV", "PARQUET"],
```
GarrettWu, line 165: nit: Python usually takes lower cases

bigframes/session/executor.py
```diff
@@ -147,6 +157,35 @@ def export(
...
+        export_options: Dict[str, Union[bool, str]],
```
GarrettWu, line 166: Any chance to make these consistent? Mapping and Dict. Prefer Mapping as a more general type.

bigframes/session/executor.py
TrevorBergeron: changed to lower case

bigframes/session/loader.py
TrevorBergeron: added

bigframes/session/loader.py
TrevorBergeron: made them all private

GarrettWu (approved):

PR #914, opened 2024-08-22
fix: struct field non-nullable type issue.

Genesis929:
This issue is related to pypi.ipynb.

Fixes b/361145903 🦕


shobsi (approved):
Let's find a way to write a system test to showcase the value of this change yaqs/4269233226351902720 in a later PR

PR #918, opened 2024-08-23
perf: Improve repr performance

TrevorBergeron:


bigframes/core/__init__.py
```diff
@@ -70,23 +70,7 @@ def from_pyarrow(cls, arrow_table: pa.Table, session: Session):
             iobytes.getvalue(),
             data_schema=schema,
             session=session,
-        )
-        return cls(node)
-
-    @classmethod
-    def from_cached(
```
tswast, line 77: I assume this was dead code?

bigframes/core/blocks.py
```diff
@@ -210,8 +211,7 @@ def shape(self) -> typing.Tuple[int, int]:
             except Exception:
                 pass
 
-        iter, _ = self.session._execute(row_count_expr, ordered=False)
-        row_count = next(iter)[0]
+        row_count = self.session._executor.get_row_count(self.expr)
```
tswast, line 214: How do we decide what operations warrant a new method on the executor?

bigframes/core/nodes.py
```diff
@@ -341,15 +363,14 @@ def order_ambiguous(self) -> bool:
...
-        return self
+    @property
+    def row_count(self) -> typing.Optional[int]:
+        return self.n_rows
 
 
 ## Put ordering in here or just add order_by node above?
 @dataclass(frozen=True)
-class ReadTableNode(BigFrameNode):
+class ReadTableNode(LeafNode):
```
tswast, line 373: Do we want a `row_count` here? If there are no filters applied, we should be able to get this from table metadata (though it could be slightly inaccurate if there was anything in the streaming buffer, which can show up in query results before the count arrives at the table metadata).

bigframes/core/nodes.py
```diff
@@ -398,6 +415,11 @@ def relation_ops_created(self) -> int:
         # Assume worst case, where readgbq actually has baked in analytic operation to generate index
         return 3
 
+    @property
+    def supports_fast_head(self) -> bool:
+        # TODO: Be more lenient for small tables, or those clustered on non-sequential order key
```
tswast, line 420: +1, `LIMIT` clause is quite cheap on clustered tables.

bigframes/core/tree_properties.py
```diff
@@ -32,14 +32,45 @@ def local_only(node: nodes.BigFrameNode) -> bool:
...
+    children_peekable = all(can_fast_peek(child) for child in node.child_nodes)
     self_peekable = not node.non_local
     return children_peekable and self_peekable
 
 
+def can_fast_head(node: nodes.BigFrameNode) -> bool:
+    """Can get head fast if can push head operator down to leafs and operators preserve rows."""
+    if isinstance(node, nodes.LeafNode):
+        return node.supports_fast_head
+    # TODO: In theory we can push head down through concat, but requires some dedicated logic
```
tswast, line 47: Are you thinking something like: limit applied to each + an overall limit clause? Does the BigQuery engine not push down `limit` through a union if we apply it at the top level?

bigframes/session/executor.py
```diff
@@ -161,20 +163,74 @@ def dry_run(self, array_value: bigframes.core.ArrayValue, ordered: bool = True):
...
-        return self._run_execute_query(
-            sql=sql,
-        )
+        return self._run_execute_query(sql=sql)
+
+    def head(
+        self, array_value: bigframes.core.ArrayValue, n_rows: int
+    ) -> tuple[bigquery.table.RowIterator, bigquery.QueryJob]:
+        """
+        A 'peek' efficiently accesses a small number of rows in the dataframe.
```
tswast, line 187: Update this docstring.

bigframes/session/executor.py
```diff
@@ -161,20 +163,74 @@ def dry_run(self, array_value: bigframes.core.ArrayValue, ordered: bool = True):
...
+
+        head_plan = generate_head_plan(plan, n_rows)
+        sql = self.compiler.compile_ordered(head_plan)
+
+        # TODO(swast): plumb through the api_name of the user-facing api that
+        # caused this query.
+        return self._run_execute_query(sql=sql)
+
+    def get_row_count(self, array_value: bigframes.core.ArrayValue) -> int:
+        # optimized plan less likely to have count-destroying operators like filter or join
```
tswast, line 217: What's this comment mean? Do you mean that if the count is not None, then we have an optimized plan?

bigframes/session/executor.py
```diff
@@ -218,7 +274,7 @@ def _run_execute_query(
             else:
                 raise
 
-    def _with_cached_executions(self, node: nodes.BigFrameNode) -> nodes.BigFrameNode:
+    def _get_optimized_plan(self, node: nodes.BigFrameNode) -> nodes.BigFrameNode:
```
tswast, line 277: Could we get a docstring for this? Or maybe rename? I'd love to understand when we'd use this? It seems we'd always want to use an "optimized" plan, as no downsides are documented here.

bigframes/core/__init__.py
TrevorBergeron: yes, probably from a recent refactor

bigframes/core/blocks.py
TrevorBergeron: Its a bit difficult to draw the line, but executor captures differents intents for consuming the dataframe, while nodes will represent value transformations not necessarily associated with immediate intent to consume. I think maybe some of this head stuff would be better handled with a bfet-level slice operator, but we don't have that at present.

bigframes/core/nodes.py
TrevorBergeron: Yes, that is probably right. In new revision will stuff some more table metadata into these nodes and use that to provide row_count where possible.

bigframes/core/nodes.py
TrevorBergeron: From what I can tell, this is not actually the case yet. All the files still get dispatched for ORDER BY+LIMIT, and so full cost is paid. Only filter over materialized row numbers is granting significant saving from my experiements. This may change in time.

bigframes/core/tree_properties.py
TrevorBergeron: I guess I didn't think too hard about this. I am not aware of engine-level ordered head optimization that would prevent dispatching all the underlying data

bigframes/session/executor.py
TrevorBergeron: done

bigframes/session/executor.py
TrevorBergeron: removed comment. essential idea is to use the row_num metadata from cached execution where possible

bigframes/session/executor.py
TrevorBergeron: added docstring. only optimization right now is to apply caching. no real downsides at this stage, other than mutating the structure of the tree, which is fine at execution stage (don't want to do it earlier as it would interfere with implicit joining).

bigframes/core/nodes.py
```diff
@@ -341,20 +363,38 @@ def order_ambiguous(self) -> bool:
...
-        return self
+    @property
+    def row_count(self) -> typing.Optional[int]:
+        return self.n_rows
 
 
-## Put ordering in here or just add order_by node above?
 @dataclass(frozen=True)
-class ReadTableNode(BigFrameNode):
+class GbqTable:
```
tswast, line 372: This is so we can get something hashable? So we can make sure we only have the fields we care about? A docstring with the purpose would be helpful here.

tswast (approved):

PR #923, opened 2024-08-23
perf: Join op discards child ordering in unordered mode

TrevorBergeron:
…er mode


bigframes/core/compile/compiler.py
```diff
@@ -76,13 +76,22 @@ def _compile_node(
...
+            if self.strict:
+                left_ordered = self.compile_ordered_ir(node.left_child)
+                right_ordered = self.compile_ordered_ir(node.right_child)
+                return bigframes.core.compile.single_column.join_by_column_ordered(
+                    left=left_ordered,
+                    right=right_ordered,
+                    join=node.join,
+                )
+            else:
+                left_unordered = self.compile_unordered_ir(node.left_child)
```
tswast, line 88: I'm curious if wee need to do anything to handle `sort=True` behavior from the [pandas join](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.join.html#pandas-dataframe-join)? I'm guessing "no", as I suspect this would be implemented as a sort after the join.

Also, let's add a comment explaining this optimization:

```suggestion
                # In general, joins are an ordering destroying operation.
                # With ordering_mode = "partial", make this explicit. In
                # this case, we don't need to provide a deterministic ordering.
                left_unordered = self.compile_unordered_ir(node.left_child)
```

bigframes/core/compile/compiler.py
```diff
@@ -268,8 +277,16 @@ def compile_projection(self, node: nodes.ProjectionNode, ordered: bool = True):
...
+            if self.strict:
+                compiled_ordered = [
+                    self.compile_ordered_ir(node) for node in node.children
+                ]
+                return concat_impl.concat_ordered(compiled_ordered)
+            else:
+                compiled_unordered = [
+                    self.compile_unordered_ir(node) for node in node.children
+                ]
+                return concat_impl.concat_unordered(compiled_unordered).as_ordered_ir()
```
tswast, line 289: This optimization worries me a little more than the join optimization. Could we update some documentation for `concat` to call out this fact? I don't find this intuitive.

bigframes/core/compile/compiler.py
TrevorBergeron: added comment

TrevorBergeron: And yes, `sort=True` sorts as an additional operation and so will still apply: https://github.com/googleapis/python-bigquery-dataframes/blob/main/bigframes/core/blocks.py#L2073-L2080

bigframes/core/compile/compiler.py
TrevorBergeron: Hmm, could be confusing if concat doesn't conserve ordering. Reverting this aspect of the change - only join will drop child ordering for now.

tswast (approved):

PR #929, opened 2024-08-24
chore: add tpch q19-22 and benchmark readme

Genesis929:

### DIFF
--- a/third_party/bigframes_vendored/tpch/queries/q22.py
+++ b/third_party/bigframes_vendored/tpch/queries/q22.py
@@ -1,0 +1,52 @@
+# Contains code from https://github.com/pola-rs/tpch/blob/main/queries/polars/q22.py
+
+import bigframes
+import bigframes.pandas as bpd
+
+
+def q(dataset_id: str, session: bigframes.Session):
+    customer = session.read_gbq(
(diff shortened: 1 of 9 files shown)


third_party/bigframes_vendored/tpch/queries/q22.py
```diff
@@ -0,0 +1,52 @@
...
+    customer = session.read_gbq(
+        f"bigframes-dev-perf.{dataset_id}.CUSTOMER",
+        index_col=bigframes.enums.DefaultIndexKind.NULL,
+    )
+    orders = session.read_gbq(
+        f"bigframes-dev-perf.{dataset_id}.ORDERS",
+        index_col=bigframes.enums.DefaultIndexKind.NULL,
+    )
+
+    country_codes = ["13", "31", "23", "29", "30", "18", "17"]
```
sycai, line 17: nit: does it make sense to comment on the country names of these codes?

sycai (approved):

third_party/bigframes_vendored/tpch/queries/q22.py
Genesis929: May not be necessary to do so; the test data is generated for benchmarking purposes, and the country names are not relevant.

PR #931, opened 2024-08-27
deps: re-introduce support for numpy 1.24.x

### LINKED ISSUE #687: bigframes 1.5.0 example showing module 'numpy' has no attribute 'dtypes'
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/python-bigquery-dataframes/issues
  - Search StackOverflow: https://stackoverflow.com/questions/tagged/google-cloud-platform+python

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS type and version:
  - Python version: `python --version`
  - pip version: `pip --version`
  - `bigframes` version: `pip show bigframes`

```python
import sys
import bigframes
import google.cloud.bigquery
import ibis
import pandas
import pyarrow
import sqlglot

print(f"Python: {sys.version}")
print(f"bigframes=={bigframes.__version__}")
print(f"google-cloud-bigquery=={google.cloud.bigquery.__version__}")
print(f"ibis=={ibis.__version__}")
print(f"pandas=={pandas.__version__}")
print(f"pyarrow=={pyarrow.__version__}")
print(f"sqlglot=={sqlglot.__version__}")
```

```
Python: 3.10.9 (main, Mar  1 2023, 12:33:47) [Clang 14.0.6 ]
bigframes==1.5.0
google-cloud-bigquery==3.22.0
ibis==8.0.0
pandas==1.5.3
pyarrow==12.0.1
sqlglot==20.11.0
```

#### Steps to reproduce

  1. Running https://github.com/googleapis/python-bigquery-dataframes/blob/main/notebooks/generative_ai/large_language_models.ipynb
  2. Gets module 'numpy' has no attribute 'dtypes' in cell 5

#### Code example

```
df = pd.DataFrame(
        {
            "prompt": ["What is BigQuery?", "What is BQML?", "What is BigQuery DataFrame?"],
        })
bf_df = bigframes.pandas.read_pandas(df)
```

#### Stack trace
```
# example
```

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

tswast:
We explicitly use the numpy module, so it should be an explicit dependency in `setup.py`. Also, make sure we test with the minimum supported version by setting the minimum version in the constraints file (see: https://github.com/googleapis/synthtool/issues/1901 and https://github.com/googleapis/python-test-utils/pull/8).

- [N/A] Appropriate docs were updated (if necessary)

Fixes #687 🦕

### DIFF
--- a/bigframes/dtypes.py
+++ b/bigframes/dtypes.py
@@ -205,67 +205,74 @@ NUMERIC_BIGFRAMES_TYPES_PERMISSIVE = NUMERIC_BIGFRAMES_TYPES_RESTRICTIVE + [
 
 
 ## dtype predicates - use these to maintain consistency
-def is_datetime_like(type: ExpressionType) -> bool:
-    return type in (DATETIME_DTYPE, TIMESTAMP_DTYPE)
+def is_datetime_like(type_: ExpressionType) -> bool:
+    return type_ in (DATETIME_DTYPE, TIMESTAMP_DTYPE)
 
 
-def is_date_like(type: ExpressionType) -> bool:
-    return type in (DATETIME_DTYPE, TIMESTAMP_DTYPE, DATE_DTYPE)
+def is_date_like(type_: ExpressionType) -> bool:
+    return type_ in (DATETIME_DTYPE, TIMESTAMP_DTYPE, DATE_DTYPE)
 
 
-def is_time_like(type: ExpressionType) -> bool:
-    return type in (DATETIME_DTYPE, TIMESTAMP_DTYPE, TIME_DTYPE)
+def is_time_like(type_: ExpressionType) -> bool:
+    return type_ in (DATETIME_DTYPE, TIMESTAMP_DTYPE, TIME_DTYPE)
 
 
-def is_binary_like(type: ExpressionType) -> bool:
-    return type in (BOOL_DTYPE, BYTES_DTYPE, INT_DTYPE)
+def is_binary_like(type_: ExpressionType) -> bool:
+    return type_ in (BOOL_DTYPE, BYTES_DTYPE, INT_DTYPE)
 
 
-def is_string_like(type: ExpressionType) -> bool:
-    return type in (STRING_DTYPE, BYTES_DTYPE)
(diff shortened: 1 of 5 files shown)


bigframes/dtypes.py
```diff
@@ -205,93 +205,100 @@ class SimpleDtypeInfo:
 
 
 ## dtype predicates - use these to maintain consistency
-def is_datetime_like(type: ExpressionType) -> bool:
-    return type in (DATETIME_DTYPE, TIMESTAMP_DTYPE)
+def is_datetime_like(type_: ExpressionType) -> bool:
```
tswast, line 208: `type` is a built-in. Avoid conflicts by using `type_`.

> `single_trailing_underscore_`: used by convention to avoid conflicts with Python keyword

https://peps.python.org/pep-0008/#descriptive-naming-styles

In this case, `type` is a function, not a keyword, so not strictly needed, but `type` is a common enough function that it is best to avoid conflicts, IMO.

shobsi (approved):

setup.py
```diff
@@ -49,6 +49,7 @@
     "google-cloud-storage >=2.0.0",
     "ibis-framework[bigquery] >=8.0.0,<9.0.0dev",
     "jellyfish >=0.8.9",
+    "numpy >=1.24.0",
```
shobsi, line 52: Does it help with the issue in the email? The user already had Numpy 1.24.0 in the issue they reported

tswast: Yes. That's the whole issue. We weren't testing against this version so we broke them. Now we are.

shobsi: Ah, so the issue is with this particular version `numpy==1.24.0`, not necessarily all subsequent versions `numpy>=1.24.0`. E.g. colab comes with `numpy==1.26.4` and read_gbq works just fine there.

PR #932, opened 2024-08-27
chore: fix api coverage script in nightly build, ensure scripts tests run

tswast:

Fixes internal issue 362260020 (hopefully) 🦕

### DIFF
--- a/scripts/publish_api_coverage.py
+++ b/scripts/publish_api_coverage.py
@@ -286,6 +286,8 @@ def build_api_coverage_table(bigframes_version: str, release_version: str):
     sklearn_cov_df["module"] = "bigframes.ml"
     combined_df = pd.concat([pandas_cov_df, sklearn_cov_df])
     combined_df["timestamp"] = pd.Timestamp.now()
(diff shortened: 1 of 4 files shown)


scripts/publish_api_coverage.py
```diff
@@ -164,13 +164,20 @@ def generate_pandas_api_coverage():
                 token_type = "property"
 
             is_in_bigframes = hasattr(bigframes_obj, member)
-            requires_index = False
-            requires_ordering = False
+            requires_index = ""
```
GarrettWu, line 167: Nit: I guess changing it from bool to "" and "Y" for type consistency? But bool seems more readable.

GarrettWu (approved):

scripts/publish_api_coverage.py
tswast: I'd like to support "partial" in future, so using string as a stand-in for an enum.

tswast:
Using admin privileges as presubmit failure for encryption is unrelated.

PR #934, opened 2024-08-28
test: allow more retries on claude tests

shobsi:
This should accommodate more flakiness in the not-so-stable claude endpoints.

### DIFF
--- a/tests/system/small/ml/test_llm.py
+++ b/tests/system/small/ml/test_llm.py
@@ -409,6 +409,7 @@ def test_gemini_text_generator_predict_with_params_success(
     "model_name",
     ("claude-3-sonnet", "claude-3-haiku", "claude-3-5-sonnet", "claude-3-opus"),
 )
+@pytest.mark.flaky(retries=3, delay=120)
 def test_claude3_text_generator_create_load(
     dataset_id, model_name, session, session_us_east5, bq_connection
 ):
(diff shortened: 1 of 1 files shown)


tswast (approved):
I wonder if these are incorrectly classified as "small"? The flakiness on presubmits is very concerning to me.

tswast:
I'm using admin permissions to merge this as well, as the claude tests are indeed passing and the encryption failure is unrelated.

shobsi:
> I wonder if these are incorrectly classified as "small"? The flakiness on presubmits is very concerning to me.

@GarrettWu how do we think of the LLM integration tests being under small tests. IIRC there have been instabilities in the other LLM tests (Palm2, Gemini) as well. Should we move them under e2e tests?

GarrettWu:
> > I wonder if these are incorrectly classified as "small"? The flakiness on presubmits is very concerning to me.
> 
> @GarrettWu how do we think of the LLM integration tests being under small tests. IIRC there have been instabilities in the other LLM tests (Palm2, Gemini) as well. Should we move them under e2e tests?

They are considered small since they are just small and short tests. Usually we put training tests to large.
There are 2 different issues with this:
1. load failed for 500: this one is new and it is a real issue instead of flakiness. I'd bring this up in today's meeting with BQML.
2. flakiness on the content: we have some updates that don't check the contents. While some tests are missed for that change. I'd send a PR to update.

PR #935, opened 2024-08-28
test: stop checking text generation contents

GarrettWu:
After the change, the .predict check shouldn't be flaky. While still keeping the "retry" label since:
1. CREATE MODEL has new flakiness b/362768658.
2. To avoid unexpected flakiness to block

I'll keep an eye on the tests. Eventually we should remove the "retry" labels, at least for GA models.

### DIFF
--- a/tests/system/small/ml/test_llm.py
+++ b/tests/system/small/ml/test_llm.py
@@ -74,10 +74,9 @@ def test_create_text_generator_model_default_session(
         llm_text_df = bpd.read_pandas(llm_text_pandas_df)
 
         df = model.predict(llm_text_df).to_pandas()
-        assert df.shape == (3, 4)
-        assert "ml_generate_text_llm_result" in df.columns
-        series = df["ml_generate_text_llm_result"]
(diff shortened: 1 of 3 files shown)


tests/system/small/ml/test_llm.py
```diff
@@ -444,10 +435,9 @@ def test_claude3_text_generator_predict_default_params_success(
         model_name=model_name, connection_name=bq_connection, session=session
     )
     df = claude3_text_generator_model.predict(llm_text_df).to_pandas()
-    assert df.shape == (3, 3)
-    assert "ml_generate_text_llm_result" in df.columns
-    series = df["ml_generate_text_llm_result"]
-    assert all(series.str.len() > 20)
```
shobsi, line 450: Is this precisely the part which was failing due to instability?

GarrettWu: It is the flakiness of the .predict. In fact this time it isn't instability but a quota issue. b/362768658. 

It is not the issue of create model.

shobsi (approved):

PR #939, opened 2024-08-29
chore: define a tree node for DQL statements.

sycai:


bigframes/core/nodes.py
```diff
@@ -786,3 +786,45 @@ def relation_ops_created(self) -> int:
...
+        # items = tuple(
+        #     typing.cast(schemata.SchemaItem, bigframes.dtypes.convert_schema_field(field))
+        #     for field in self.physical_schema
+        # )
+        items = tuple(map(_bq_to_bf_schema, self.physical_schema))
+        return schemata.ArraySchema(items)
+
+    @property
+    def variables_introduced(self) -> int:
+        return 0
```
tswast, line 812: I think we use this to determine query complexity. 0 is likely way too small. @TrevorBergeron what are your thoughts?

A lower bound could be the number of columns like here: https://github.com/googleapis/python-bigquery-dataframes/blob/7ed39629c638874d8e9cc3c7a9b3ec92ad480eca/bigframes/core/nodes.py#L305 ?

We could also estimate the number of joins with some regex?

Or estimate the number of joins with this stat from a query dryrun: https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobStatistics2.FIELDS.referenced_tables

TrevorBergeron: Estimating this robustly seems quite difficult, as I don't think dry runs provide the query plan. An estimate can be generated using schema size and counting joins using a regex, but this does risk wildly underestimating complexity in some cases.

sycai: The response of a query dry run contains the schema field: https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#response-body. Does it make sense to use the size of that field plus OVERHEAD? @tswast @TrevorBergeron

tswast: > Does it make sense to use the size of that field plus OVERHEAD?

It'd be an underestimate in many cases, but would at least give a lower bound.

How about `"number of fields" + ("number of tables referenced" * OVERHEAD)`, since we know there will be at least one join per table referenced.

sycai: Okay. I will go head and use the formula `"number of fields" + ("number of tables referenced" * OVERHEAD)`.

PR #940, opened 2024-08-29
feat: enable read_csv() to process other files

chelsea-lin:
- [X] Make sure to open an issue as a bug/issue(internal issue 361420947) before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea

Fixes internal issue 361420947 🦕

### DIFF
--- a/tests/system/small/test_session.py
+++ b/tests/system/small/test_session.py
@@ -1051,7 +1051,7 @@ def test_read_csv_local_w_usecols(session, scalars_pandas_df_index, engine):
 )
 def test_read_csv_others(session, scalars_pandas_df_index, engine):
     # TODO: Update to `main` branch after merging https://github.com/googleapis/python-bigquery-dataframes/pull/938
-    uri = "https://raw.githubusercontent.com/googleapis/python-bigquery-dataframes/main_chelsealin_addcsvfile/tests/data/people.csv"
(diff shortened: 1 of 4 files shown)


tests/system/small/test_session.py
```diff
@@ -1036,6 +1036,26 @@ def test_read_csv_local_w_usecols(session, scalars_pandas_df_index, engine):
...
 
 
+@pytest.mark.parametrize(
+    "engine",
+    [
+        pytest.param(
+            "bigquery",
+            id="bq_engine",
+            marks=pytest.mark.xfail(
+                raises=NotImplementedError,
```
GarrettWu, line 1046: So it only available in default engine? We should update the docs to be explicit. Also should update the doc that we support global paths(currently says local only).

chelsea-lin: Good catch! Done.

tests/system/small/test_session.py
```diff
@@ -1036,6 +1036,26 @@ def test_read_csv_local_w_usecols(session, scalars_pandas_df_index, engine):
...
+            ),
+        ),
+        pytest.param(None, id="default_engine"),
+    ],
+)
+def test_read_csv_others(session, scalars_pandas_df_index, engine):
+    # TODO: Update to `main` branch after merging https://github.com/googleapis/python-bigquery-dataframes/pull/938
```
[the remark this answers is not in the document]
chelsea-lin: Done.

GarrettWu (approved):

PR #941, opened 2024-08-30
feat: allow setting table labels in `to_gbq`

mattyopl:

Fixes b/312727426 🦕


mattyopl:
tested: https://screenshot.googleplex.com/3iDSwZSmyuEXq8j

bigframes/dataframe.py
```diff
@@ -3106,6 +3107,12 @@ def to_gbq(
                 + constants.DEFAULT_EXPIRATION,
             )
 
+        if len(labels) != 0:
```
chelsea-lin, line 3110: We probably want to add a `system` test so that the test coverage didn't get down:

https://screenshot.googleplex.com/73WshjpaD6NZpjD from https://fusion2.corp.google.com/ci/kokoro/prod:bigframes%2Fpresubmit%2Fpresubmit/activity/90e96fb7-f11b-4590-aab4-e1563516b6b2/log

mattyopl: done!

bigframes/dataframe.py
```diff
@@ -3027,6 +3027,7 @@ def to_gbq(
         index: bool = True,
         ordering_id: Optional[str] = None,
         clustering_columns: Union[pandas.Index, Iterable[typing.Hashable]] = (),
+        labels: dict[str, str] = {},
```
chelsea-lin, line 3030: Please update the doc of this function too:
https://github.com/googleapis/python-bigquery-dataframes/blob/main/third_party/bigframes_vendored/pandas/core/frame.py#L385

chelsea-lin (commented):
Overall looks good to me, if Garrett wants to take another look.

bigframes/dataframe.py
```diff
@@ -3106,6 +3107,12 @@ def to_gbq(
                 + constants.DEFAULT_EXPIRATION,
             )
 
+        if len(labels) != 0:
+            client = bigquery.Client()
```
GarrettWu, line 3111: Instead of create a new client every time, we should reuse the df.session.bqclient.

GarrettWu (commented):
The PR title should be "feat".

bigframes/dataframe.py
mattyopl: done!

bigframes/dataframe.py
mattyopl: done!

GarrettWu (approved):

tests/system/small/test_dataframe.py
```diff
@@ -4657,6 +4658,17 @@ def test_to_gbq_and_create_dataset(session, scalars_df_index, dataset_id_not_cre
     assert not loaded_scalars_df_index.empty
 
 
+def test_to_gbq_table_labels(scalars_df_index):
+    destination_table = "bigframes-dev.bigframes_tests_sys.table_labels"
+    result_table = scalars_df_index.to_gbq(
+        destination_table, labels={"test": "labels"}, if_exists="replace"
+    )
+    client = bigquery.Client()
```
chelsea-lin, line 4666: `client = scalars_df_index._session.bqclient`. Usually, we don't create a separate BQ client, which is expensive and required additional setup, such as default project, token etc.

bigframes/dataframe.py
```diff
@@ -3106,6 +3107,12 @@ def to_gbq(
                 + constants.DEFAULT_EXPIRATION,
             )
 
+        if len(labels) != 0:
+            client = self._session.bqclient
```
chelsea-lin, line 3111: Setting up labels currently requires additional client calls, which is acceptable but not ideal for performance. Could you try setting the label directly at the `bigquery.table.TableReference destination`? While `TableReference` doesn't have a public API for this, you can experiment with the internal call similar to [Table.label](https://github.com/googleapis/python-bigquery/blob/ea69fe315592f3a73417f175b1fe4543203cb716/google/cloud/bigquery/table.py#L480). If this works, we can explore adding a new public API to TableReference. +@tswast has more experience with the Python client API and could provide additional insights.

mattyopl: I removed one client call by using the TableReference returned by the Query job, but as far as I can tell we might still have to make another API call for `update_table`. Looking at seeing if I can change table labels upon creation but I haven't found where that's being done yet.

chelsea-lin: Thanks for exploring it.

chelsea-lin (approved):

tests/system/small/test_dataframe.py
```diff
@@ -4657,6 +4657,17 @@ def test_to_gbq_and_create_dataset(session, scalars_df_index, dataset_id_not_cre
     assert not loaded_scalars_df_index.empty
 
 
+def test_to_gbq_table_labels(scalars_df_index):
+    destination_table = "bigframes-dev.bigframes_tests_sys.table_labels"
```
shobsi, line 4661: Ideally the test should not be hard coupled with a project/dataset. Do we have to use a fixed table name? I think the test would still serve its purpose if we did `to_gbq`  without an explicit table arg and asserted labels on the returned anonymous table.

third_party/bigframes_vendored/pandas/core/frame.py
```diff
@@ -467,6 +468,9 @@ def to_gbq(
                 clustering order within the Index/DataFrame columns follows the order
                 specified in `clustering_columns`.
 
+            labels (dict[str, str], default None):
+                Specifies table labels within BigQuery
```
shobsi, line 472: nit, (but since this goes in the public doc) should end with the period `.`

tests/system/small/test_dataframe.py
mattyopl: done in https://github.com/googleapis/python-bigquery-dataframes/pull/968!

third_party/bigframes_vendored/pandas/core/frame.py
mattyopl: done in https://github.com/googleapis/python-bigquery-dataframes/pull/968!

PR #948, opened 2024-09-03
chore: fix the test cover workflow

chelsea-lin:
This change is fixing a regression of the `cover` workflow. The `upload-artifact` excludes hidden files by default [here](https://github.com/actions/upload-artifact/commit/50769540e7f4bd5e21e526ee35c689e35e0d6874).

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -30,6 +30,7 @@ jobs:
       with:
         name: coverage-artifact-${{ matrix.python }}
         path: .coverage-${{ matrix.python }}
+        include-hidden-files: true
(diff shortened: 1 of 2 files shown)


GarrettWu (approved):

tswast (approved):

shobsi (approved):

chelsea-lin:
The change gets updated by owlbot so the change should be fixed here: https://github.com/googleapis/synthtool/pull/2008
This change excludes the `unittest.yml` from the owlbot so that we can unblock BF workflow earlier if the synthtool fixing takes a longer time.

PR #950, opened 2024-09-04
fix: support `read_gbq_function` for axis=1 application

shobsi:

Fixes internal issue 362988965 🦕

### DIFF
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -1223,6 +1223,7 @@ class Session(
     def read_gbq_function(
         self,
         function_name: str,
+        is_row_processor: bool = False,
     ):
         """Loads a BigQuery function from BigQuery.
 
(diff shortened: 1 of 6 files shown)


bigframes/session/__init__.py
```diff
@@ -1264,13 +1265,45 @@ def read_gbq_function(
             <BLANKLINE>
             [3 rows x 3 columns]
 
+        You can even use a function with multiple inputs. For example, let's use
```
tswast, line 1268: Nit from documentation style guide:

> In general, address the reader of your documents using the second person instead of the first person: use you or your instead of we, our, or us.

https://developers.google.com/style/person

Drop the "let's" (the `'s` means "us"):

```suggestion
        You can even use a function with multiple inputs. For example, use
```

tswast (approved):

PR #951, opened 2024-09-04
chore: temporarily disable struct accessor tests

shobsi:
This is to unblock PRs until the real issue is identified and fixed in b/362738764.


tswast (changes requested):
doctest isn't a required check, so this feels unnecessary. Plus the `doctest: +SKIP` will end up in the API reference docs, which isn't ideal.

My understanding with conversation with @Genesis929 is that we can alphabetize the struct subfields in the sample data we create as a better temporary fix.

tswast:
Closing in favor of https://github.com/googleapis/python-bigquery-dataframes/pull/953#pullrequestreview-2280586198

Issue #647, opened 2024-04-30
Allow querying the bigframes dataframe with custom BigQuery SQL like pandasql


NiloFreitas:
###  **Feature Request**
Supposed I used bigframes and processed my data like this:
`res_df = df.assign(pred=input_remote_function.apply(generate_text))`

And got this output dataframe **res_df**:

uri                                                                                  | pred                                                                        |
-------------------------------------------------- | -----------------------------------------------|
gs://dataproc-metastore-public-binaries/ads_ba... | {"interpretation": " The primary message of th...
gs://dataproc-metastore-public-binaries/ads_ba... | {"interpretation": " The primary message of th...
gs://dataproc-metastore-public-binaries/ads_ba... | {"interpretation": " The primary message of th...

As you can see, the **pred** column is a JSON string.

 ### **Describe the solution you'd like**

I would like to use [BigQuery JSON Functions](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions) to explode this column into a column for each JSON attribute.

For example, I could write a custom BigQuery SQL code to run for this dataframe, like this:
`res_df = df.sqldf("SELECT *, JSON_EXTRACT(pred, '$.interpretation') AS interpretation FROM df")
`

Getting as a result the **res_df**:
uri                 | pred                                                                          | interpretation                  |
------------ | ------------------------------------------------|-------------------------|
gs://datap... | {"interpretation": " The primary message of th... |  The primary messag...  |
gs://datap... | {"interpretation": " The primary message of th... |  The primary messag...  |
gs://datap... | {"interpretation": " The primary message of th... |  The primary messag...  |

### **Describe alternatives you've considered**

As an alternative I need to use remote functions, like this one, to explode the column:

```
@bpd.remote_function(
    [str],
    str,
    bigquery_connection=CONNECTION_ID
)
def extract_json_attribute(input_content: str) -> str:
    
    pred = input_content.split("|||")[0]
    json_attribute = input_content.split("|||")[1]
    
    import json
    return json.loads(pred)[json_attribute]
```

```
input_remote_function_int = res_df['pred'] + '|||interpretation'
df = res_df.assign(interpretation=input_remote_function_int.apply(extract_json_attribute))
```
Which I cannot pass multiple parameter because of https://github.com/googleapis/python-bigquery-dataframes/issues/646

 ### **Additional context**

The pandasql allow us to query pandas DataFrames using SQL syntax.
[...]

PR #955, opened 2024-09-04
feat: add bigframes.ml.compose.SQLScalarColumnTransformer to create custom SQL-based transformations

### LINKED ISSUE #954: Add possibility to create Custom-Transformers in compose.ColumnTransformer
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

 **Is your feature request related to a problem? Please describe.**
Transform-Only models (compose.ColumnTransformer) are a performant way to do the data preprocessing.
Also it is possible to bundle the transformation together with an estimator into a ML-Model, which is then able to do the preprocessing and the inferencing together.

There exists a number of Transformations (ML.STANDARD_SCALER, ..., ML.IMPUTER) which cover standard use cases.
In addition we have custom transformations, which we would like to integrate also into the model.

ColumnTransformer does not support other transformers. It is limited to the ML.xxx transformers.

 **Describe the solution you'd like**
We would like to add a possibility to add custom transformations, so that it is possible to add new custom encoders and integrate them seamlessly to the existing ones.

 **Describe alternatives you've considered**
Currently we have two work-streams. 
One evaluates the BigQuery "CREATE MODEL ... TRANSFORM ..." SQL approach.
And the other one analyzes the BigFrames capabilities.

With the support of Custom-Transformators in BigFrames it would be possible to interchange between both architectures.
The transformation created with BigFrames could also be used in BigQuery and vice-versa.
BigFrames is the prefered option for data scientists which have a strong Pandas background but less SQL experience.
So it would be a one-time effort to create the custom transformers which need SQL code.
Then these Custom-Transformers can be used in the BigFrames code.

 **Additional context**
We discussed the issue with Google Experts and they proposed to create a pull-request to discuss the pros and cons.

ferenc-hechler:

Fixes #954 🦕

This PR adds the possibility to create custom-transformers to be used in ColumnTransformer.
The base idea is to implement the _compile_to_sql() and _parse_from_sql() methods which are called from ColumnTransformer.
Each CustomTransformer registers itself in a registry and in addition to _BQML_TRANSFROM_TYPE_MAPPING, which contains the ML. transformers, this registry is queried, when transformers are created from reading in a model.

A very simple example is the Identity-Transformer. It does not change the value:

```
  column AS ident_column
```

The implementation looks like this:

```
from bigframes.ml.compose import CustomTransformer
from typing import List, Optional, Union, Dict
import re

IDENT_BQSQL_RX = re.compile("^(?P<colname>[a-z][a-z0-9_]+)$", flags=re.IGNORECASE)

class IdentityTransformer(CustomTransformer):
    _CTID = "IDENT"

    def custom_compile_to_sql(self, X: bpd.DataFrame, column: str) -> str:
        return f"{column}"

    @classmethod
    def custom_parse_from_sql(cls, config: Optional[Union[Dict, List]], sql: str) -> tuple[CustomTransformer, str]:
        col_label = IDENT_BQSQL_RX.match(sql).group("colname")
        return cls(), col_label

CustomTransformer.register(IdentityTransformer)
```

The class CustomTransformer is new and was added in this PR to bigframes.ml.compose, where also ColumnTransformer is implemented.
It provides some base functionality, like iterating over the column list, which can be reused for all inherited classes. 
It also maintains the registry of custom transformers.
The class variable `_CTID` is used for identifying the transform and has to be unique over all registered custom transformers.

Another example which replaces a string column with its length and uses a configurable default-value for NULL values.
The SQL for implementing this transformation looks like this:

```
  CASE WHEN {column} IS NULL THEN {default_value} ELSE LENGTH({column}) END AS len_{column}
```

Here the corresponding code 

```
LEN_BQSQL_RX = re.compile("^CASE WHEN (?P<colname>[a-z][a-z0-9_]*) IS NULL THEN (?P<defaultvalue>[-]?[0-9]+) ELSE LENGTH[(](?P=colname)[)] END$", flags=re.IGNORECASE)

class LengthTransformer(CustomTransformer):
    _CTID = "LEN"
    _DEFAULT_VALUE_DEFAULT = -1
    
    def __init__(self, default_value: Optional[int] = None):
        self.default_value = default_value

    def custom_compile_to_sql(self, X: bpd.DataFrame, column: str) -> str:
        default_value = (
            self.default_value
            if self.default_value is not None
            else LengthTransformer._DEFAULT_VALUE_DEFAULT
        )
        return f"CASE WHEN {column} IS NULL THEN {default_value} ELSE LENGTH({column}) END"
    
    @classmethod
    def custom_parse_from_sql(cls, config: Optional[Union[Dict, List]], sql: str) -> tuple[CustomTransformer, str]:
        m = LEN_BQSQL_RX.match(sql)
        col_label = m.group("colname")
        default_value = int(m.group("defaultvalue"))
        return cls(default_value), col_label

CustomTransformer.register(LengthTransformer)
```

The custom_compile_to_sql() implementation is straight forward.
In custom_parse_from_sql() the configured parameters are reverse engineered from the SQL code using RegEx.

But this might not always be possible or useful, so we added a feature to persist config parameters along with the sql.
This is done by using a comment section in the SQL.

```
  CASE WHEN {column} IS NULL THEN {default_value} ELSE LENGTH({column}) END /*LEN([99])*/ AS len_{column}
```
So, the comment `/*<custom-transformer-id>(<json-args>)*/` is stored in the SQL text of the model and can be used to retrieve this information when parsing the SQL. 
Okay, this is a hack and It will not work, when the model strips away the comments. 
But at the moment it works and if this will change in future, another solution can be found to store this information in the model.

The CustomTransformer class implements the serialization of the config, so only the method get_persistent_config() has to be implemented returning a dict or list:

```
    def get_persistent_config(self, column: str) -> Optional[Union[Dict, List]]:
        return [self.default_value]
```

When parsing the SQL, this persistent_config is retrieved from the SQL and given as additional input to the parse method:

```
    @classmethod
    def custom_parse_from_sql(cls, persistent_config: Optional[Union[Dict, List]], sql: str) -> tuple[CustomTransformer, str]:
        col_label = LEN2_BQSQL_RX.match(sql).group("colname")
        default_value = persistent_config[0]   # get default value from persistent_config instead of parsing SQL
        return cls(default_value), col_label
```

Custom transformers can be used in the same way, as the standard transformers, like LabelEncoder.
This piece of codes creates a ColumnTransformer, fits it and stores the Transform-Only model.

```
df = bpd.read_gbq(SELECT_TABLE)
col_trans = ColumnTransformer([
    ("identity", IdentityTransformer(), ['address', 'pack', 'store_location']),
    ("length", Length1Transformer(-2), ['store_location']),
    ("length", Length2Transformer(99), ['address']),
    ("labelenc", LabelEncoder(), ['county', 'category_name']),
])
col_trans = col_trans.fit(df)
col_trans.to_gbq(TRANSFORM_MODEL, replace=True)
```

IdentityTransformer, Length1Transformer and Length2Transformer are inhreited from CustomTransformer, while LabelEncoder is from bigframes.ml.preprocessing.

Loading and applying the model is done with the following code:

```
df = bpd.read_gbq(SELECT_TABLE)
transform_model = bpd.get_global_session().read_gbq_model(TRANSFORM_MODEL)
df_trans = transform_model.transform(df)
```

transform_model is the ColumnTransformer which we persisted above.

Here is the complete source code:
[custom_transformers.zip](https://github.com/user-attachments/files/16873888/custom_transformers.zip)


tswast:
Thanks so much for the Pull Request and the detailed discussion @ferenc-hechler !

> So, the comment /*<custom-transformer-id>(<json-args>)*/ is stored in the SQL text of the model and can be used to retrieve this information when parsing the SQL.
> Okay, this is a hack and It will not work, when the model strips away the comments.

This is a fascinating approach, and not one that I've considered. I had considered embedding parameters in SQL, but not how one could recover them as you do here. I'll check in with some backend BQML folks to see if this comment embedding is something we could rely on.

bigframes/ml/compose.py
```diff
@@ -46,6 +48,114 @@
...
+            f"{self.custom_compile_to_sql(X, column)} {self._get_sql_comment(column)} AS {self.get_target_column_name(column)}"
+            for column in columns
+        ]
+
+    def get_target_column_name(self, column: str) -> str:
+        return f"{self._CTID.lower()}_{column}"
+
+    @classmethod
+    @abc.abstractclassmethod
+    def custom_compile_to_sql(cls, X: bpd.DataFrame, column: str) -> str:
```
tswast, line 103: As much as possible, I'd like to avoid having to fallback to SQL as part of our public interfaces. I do also see the utility in providing such an extensions point, though.

If we do provide such hooks, I'd like it to be via a more general purpose API such as https://github.com/googleapis/python-bigquery-dataframes/issues/647 and have CustomTransform be able to hook into our compilation somehow to generate a TRANSFORM clause instead of a full table expression (e.g. SELECT ... FROM ...).

tswast (commented):
Thanks so much for your work here and the code examples in the tests. I definitely want to provide something like this, so this is a great starting point. I'll do some design work this week to come up with some proposals for how we might hook into our compilation steps to be able to do something like this in a less SQL-forward way.

bigframes/ml/compose.py
ferenc-hechler: just my 2 cents:

looking at the into of the pandasql README:

```
pandasql [...] provides a more familiar way of manipulating and cleaning data for people new to Python or pandas [using SQL syntax].
```

So it is thought for developers who are already familiar with SQL and new to pandas.
Our intention to use BigFrames is to please developers who are familiar with Python and pandas but do not know much about SQL.
:-)

I understand your concerns to break the BigFrames philosophie to abstract all the BQ SQL stuff behind the pandas interface. 

ColumnTransformer already is a very good abstraction of the TRANSFORM clause.
It manages very good all the "ML." transformations but lacks support for other transformations (mostly CASE WHEN).

So, BQ SQL experts can create a set of CustomTransformers and the DataScientists using BigFrames will never get in touch with BQ SQL. They use the CustomTransformers in the same way, as they use the predefined transformers.

bigframes/ml/compose.py
```diff
@@ -46,6 +48,114 @@
...
+        if args != "":
+            config = json.loads(args)
+        else:
+            config = None
+        sql = m.group("sql").strip()
+        return cls.custom_parse_from_sql(config, sql)
+
+    @classmethod
+    @abc.abstractclassmethod
+    def custom_parse_from_sql(
```
tswast, line 137: How important is this feature to you? Do you expect folks will need to inspect the learned parameters even after loading the trained transform from BigQuery? Without this feature, folks would still be able to inspect parameters after calling `fit()`.

bigframes/ml/compose.py
tswast: I see. So, the custom transformation API is targeted at a different user base. I'd still like to do some thinking to see if there's a way we can offer a SQL escape hatch in line with long term plans.

bigframes/ml/compose.py
ferenc-hechler: I could think about a VertexAI-Pipeline (Kubeflow), which creates a transform-only model in the first component and in following steps it tests different estimators and finally creates a ml.Pipeline which contains the transformation together with the best estimator together in one Model. So, the final model contains the initially created tranformation and the prediction together.

Each VertexAI component is executed in a seperate process, so the transformation created using fit() is not available in the succeeding steps. It has to be loaded.

But we could simplify things and instead of parsing sql we could only use the persisted config. 
Parsing the sql was neccessary to retrieve the original column label. 
But this could also be added to the persisted config `get_persistent_config()`.

tswast: I see you use `TRANSFORM_ONLY` models in the description of this PR. I'll try to use such a model with `read_gbq_model()`. If so, I'm wondering if it's necessary to save the parameters in comments.

tswast: I see we have a limitation currently in BigQuery DataFrames. I created a `TRANSFORM_ONLY` model

```
CREATE OR REPLACE MODEL `swast-scratch`.bigframes_pr955.species_length
TRANSFORM(CASE WHEN species IS NULL THEN -1 ELSE LENGTH(species) END AS len1_species)
  OPTIONS (
    model_type = 'transform_only')
AS (
  SELECT *
  FROM `bigquery-public-data.ml_datasets.penguins`
);
```

But this "model" failed to load with `len1_species = bpd.read_gbq_model("swast-scratch.bigframes_pr955.species_length")`. It's failing on this assertion: https://github.com/googleapis/python-bigquery-dataframes/blob/569a7ad5a2fe72a8d3deb1304eb7180b176d8830/bigframes/ml/compose.py#L168

I will continue to investigate this. IMO we should allow for this kind of TRANSFORM_ONLY model to be used from bigframes.

ferenc-hechler: Yes, in my sample code I saved the ml.ColumnTransformer and not a ml.Pipeline.
When reading the model with read_gbq_model() the ColumnTransformer is returned again, execpt for the case, that all columns are transformed in the same way, then the Transformer itself is "merged" together and the ML.xxx transformer is returned. To simplify the CustomTransformer code I disabled merging if a CustomTransformer is contained.

Putting the parameters into the comment inside the sql of each column is a workaround.
Maybe we can persist this info at another place, e.g. the Description of the model.
Then each CustomTransformer returns the plain SQL in compile_to_sql() and the ColumnTransformer writes the list of params (retrieved with get_persistent_config()) for each CustomTransformer together with the CustomTransformer-ID and the column name into the model description. 

I am fine, if ColumnTransformer is not supporting the edge case of an empty transformation.

tswast (commented):
Thanks for your patience as I try to understand this change and your requirements.

Some questions for you:

1. How important is it for the custom transformation to be deployed as part of the Pipeline to BQML? Another way to phrase this: does all preprocessing need to happen in the TRANSFORM clause?
2. If (1) is important, then how important is it that one can recover the original Pipeline after training? That is to say, if it were possible to use the trained model for predictions without being able to inspect exactly what model / transformations were applied from Python, would that be acceptable?

Overall, I'm warming up to the idea of a SQL-based extension mechanism / helpers, but I am concerned about

1. Offering such a bigframes.ml specific way to hook into how BigQuery DataFrames compiles its SQL.
2. Trying to recover the original pipeline via special comments.

Long-term, I'd like for our extension mechanism to be provided by the existing scalar BigQuery DataFrames APIs, such as `Series.case_when`.

I envision something like the following:

```python
class OneHotEncodingPlusOutliers([bigframes.ml.ScalarTransformerMixin](http://bigframes.ml.scalartransformermixin/)):
	def __init__(self, column_name, num_values):
		self.column_name = column_name
		self.num_values = num_values

	def fit(self, X):
		# Find unique values in column_name.
		values = X[[self.column_name]].groupby(self.column_name, as_index=False).size()
		self.top_values_ = values.sort_values("size").head(self.num_values).to_list()

	def transform(self, X):
		# Note: This can only do scalar operations.
		original = X[self.column_name]
		X = X.drop(columns=[self.column_name])
		X = X.assign(**{
			f"{self.column_name}_{value}": original.eq(value)			for value in self.top_values_
		})
		X = X.assign(**{f"{self.column_name}_others": ~original.isin(self.top_values)})
		return X
```

This will take some time, though. Probably longer than the timelines we'd like. Compilation is changing, so extracting the SQL for these scalar transformations is a bit blocked by https://github.com/googleapis/python-bigquery-dataframes/pull/827

Some alternatives to consider:

(a). Make bigframes better able to utilize TRANSFORM_ONLY models that have already been deployed to BQML "models" by making them compatible with `read_gbq_model()`. This would required relaxing some of the validations, as you've done in this PR. We're tracking this request internally in issue 365996054.

A downside of this approach is that the BQML TRANSFORM clause cannot contain a call to ML.PREDICT / ML.TRANSFORM. To allow these TRANSFORM_ONLY models in a Pipeline would require changes to our compilation and possibly our execution. All transforms that happen before (and including) one of these opaque read_gbq_model() TRANSFORM_ONLY models cannot be deployed with the final BQML model for the Pipeline. Instead, we'd have to embed these transformations in the SQL table expression representing the training data and/or execute several queries.

(b). If (a) is acceptable, then I would be open to some helpers in bigframes to make deploying custom TRANSFORM_ONLY models based on SQL easier.

(c). If (a/b) is not acceptable, perhaps there is some intermediary step we can take towards the long-term vision that would unblock you?

ferenc-hechler:
We discussed your questions internally and came to the following conclusions.

> 1. How important is it for the custom transformation to be deployed as part of the Pipeline to BQML? 
> Another way to phrase this: 
> does all preprocessing need to happen in the TRANSFORM clause?

If all preprocessing could happen in one TRANSFORM clause, this would be the ideal case (top scenario).

image

But we already recognized that our existing Transformer-Framework has no limitation on transforming 
a column multiple times. So, we would most likely end up with multiple input transformers and only the 
last one can be integrated into the bundle (middle scenario).

To answer your question: If we anyway have to apply other transformations, we do not need to bundle
the last transformation with the estimator (scenario 3). So, we are fine if this would not be supported.


> [...] how important is it that one can recover the original Pipeline after training? 
> That is to say, if it were possible to use the trained model for predictions without being able 
> to inspect exactly what model / transformations were applied from Python, 
> would that be acceptable?

That one was harder to answer. Let´s formulate this way:
It would not be a show-stopper for us, but we are loosing transparency.
If a Model has bad prediction performance in production, the reason for this has to be analyzed. 
And for the DataScientists the fitted-parameters are a valuable hint,
to find out what might be different between training and production data.

Some kind of data-shift/-skew can be detected based on the transformer-model parameters.


A question:
Where do the ML.xxx transformers store there fitted parameters? 
Could we use the same place for storing our custom transformer parameters?


> Long term [...] I envision something like the following:  
> class OneHotEncodingPlusOutliers( [...] 

This is a good example, what is not possible using plain ColumnTransformer transformations.
If this is possible (and performant) we do not need to talk any more about generating SQLs   :-) 

But, as we focus on a Transformer-Framework which creates the transformations based on a given configuration, 
we can manage such transformations outside of the ColumnTransformer fit() method.
The elementary column-transformers would then just be EqualityTransformers for the same column with different output names.

```
col_trans = ColumnTransformer([
    ("book_cat", EqualityTransformer("book"), ['category']),
    ("record_cat", EqualityTransformer("record"), ['category']),
    ("magazine_cat", EqualityTransformer("magazine"), ['category']),
    ...
])
```

And in our Transformer-Framework this would be a config like:

```
preprocessing:
  stages:
    - stage_name: "CategoricalEncoder"
      stage_params:
        cardinalityThreshold: 10
        newCategory: "other"
        frequencyThreshold: .01
```

This means, the transformer frameworks creates one-hot-encoding columns if the cardinality is below 10, 
otherwise another encoder (Mean-Target) is used, which would then look like this:

```
col_trans = ColumnTransformer([
    ("tme", TargetMeanEncoder(), ['category']),
    ...
])
```

And the parameters for the TargetMeanEncoder would then be derived in the ColumnTransformer.fit() call.

```json
{"book": 33.27,"record": 18.45,"magazin": 4.99,...}
```

Finally the SQL generated would be a case-when.

So, we are able to split the preprocessing and are not limited to one-column transformations.
But the result will always be a list of ColumnTransformers which are then applied sequentially.


> (a). Make bigframes better able to utilize TRANSFORM_ONLY models 
> that have already been deployed to BQML

Does this mean, that unknown transformations are not dropped but passed through as black-box transformers?
I think this would be good, independently of our needs.


> A downside of this approach is that the BQML TRANSFORM clause 
> cannot contain a call to ML.PREDICT / ML.TRANSFORM.

As already mentioned above, this is fine for us.


> (b). If (a) is acceptable, then I would be open to some helpers in bigframes 
> to make deploying custom TRANSFORM_ONLY models based on SQL easier.

If we also get the possibility to store our fitted parameters somewhere, 
this is a good alternative to pimp up ColumnTransformer.


> (c). If (a/b) is not acceptable, perhaps there is some intermediary step 
> we can take towards the long-term vision that would unblock you?

It feels like playing a book adventure and after each chapter you have to decide, 
which chapter to continue reading until you find the treasure...    :-)

ferenc-hechler:
Just tested loading a ColumnTransformer containing CustomTransformers and adding it to a pipeline and then train a Linear-Regression model including the transformation:

```
col_trans = bpd.get_global_session().read_gbq_model(transformer_name)
model = LinearRegression(fit_intercept=False)
pipeline = Pipeline([
  ('preproc', col_trans),
  ('linreg', model)
])
pipeline.fit(X_train, y_train)
pipeline.to_gbq(full_model_name, replace=True)
```

That worked without any problems.
Here is the final SQL generated from the pipeline.fit() step:

```
CREATE OR REPLACE MODEL `de0360-pkce-spoke2-al-dev0`.`_4159a931697fdd40d66d89a5a31999be23c1295f`.`88a63fa5508c41688499916e032e68d9`
TRANSFORM(
  address /*CT.IDENT()*/ AS ident_address,
  CASE WHEN address IS NULL THEN 99 ELSE LENGTH(address) END /*CT.LEN2([99])*/ AS len2_address,
  store_location /*CT.IDENT()*/ AS ident_store_location,
  ML.STANDARD_SCALER(sale_dollars) OVER() AS standard_scaled_sale_dollars,
  ML.STANDARD_SCALER(volume_sold_gallons) OVER() AS standard_scaled_volume_sold_gallons,
  CASE WHEN store_location IS NULL THEN -2 ELSE LENGTH(store_location) END /*CT.LEN1()*/ AS len1_store_location,
  ML.LABEL_ENCODER(county, 1000000, 0) OVER() AS labelencoded_county,
  ML.LABEL_ENCODER(category_name, 1000000, 0) OVER() AS labelencoded_category_name,
  pack /*CT.IDENT()*/ AS ident_pack,
  sale_dollars)
OPTIONS(
  model_type="LINEAR_REG",
  data_split_method="NO_SPLIT",
  optimize_strategy="auto_strategy",
  fit_intercept=False,
  l2_reg=0.0,
  max_iterations=20,
  learn_rate_strategy="line_search",
  min_rel_progress=0.01,
  calculate_p_values=False,
  enable_global_explain=False,
  INPUT_LABEL_COLS=["sale_dollars"])
AS SELECT
  t0.`col_9` AS `address`,
  t0.`col_10` AS `pack`,
  t0.`col_11` AS `store_location`,
  t0.`col_12` AS `county`,
  t0.`col_13` AS `category_name`,
  t0.`col_14` AS `volume_sold_gallons`,
  t0.`col_16` AS `sale_dollars`
FROM (
  SELECT
    t1.`col_9`,
    t1.`col_10`,
    t1.`col_11`,
    t1.`col_12`,
    t1.`col_13`,
    t1.`col_14`,
    t1.`col_16`
  FROM `de0360-pkce-spoke2-al-dev0`._4159a931697fdd40d66d89a5a31999be23c1295f.bqdf20240917_sessionc82c96_bd13a2e0379c47b796f26ff3b554ed74 AS t1
) AS t0
```

bigframes/ml/compose.py
```diff
@@ -46,6 +48,114 @@
 )
 
 
+CUSTOM_TRANSFORMER_SQL_RX = re.compile(
+    "^(?P<sql>.*)/[*]CT.(?P<id>[A-Z]+[A-Z0-9]*)[(](?P<config>[^*]*)[)][*]/$",
+    re.IGNORECASE,
+)
+
+
+class CustomTransformer(base.BaseTransformer):
```
tswast, line 57: Per my most recent comment, would you be open to renaming this to `SQLScalarColumnTransformer` and accepting `sql` and `columns` as arguments to `__init__`?

This naming would help us align with a similar `bigframes.bigquery.sql_scalar(...)` feature I'd like to add to support folks that want a SQL escape hatch on the `bigframes.pandas.*` APIs.

bigframes/ml/compose.py
```diff
@@ -147,10 +254,30 @@ def camel_to_snake(name):
...
+                    found_transformer = True
+
+            if not found_transformer:
+                if not transform_sql.startswith("ML.") and "/*CT." not in transform_sql:
+                    continue  # ignore other patterns, only report unhandled known patterns
+                if transform_sql.startswith("ML."):
+                    raise NotImplementedError(
+                        f"Unsupported transformer type. {constants.FEEDBACK_LINK}"
+                    )
+                raise ValueError("Missing custom transformer")
```
tswast, line 278: Rather than ValueError and NotImplementedError, could we provide a general purpose UnknownTransformer class? I'm aware this would lose some transparency, but maybe it could include an attribute that shows the SQL substring it corresponds to?

bigframes/ml/compose.py
tswast: We'll maybe need some additional infrastructure in our compilation to allow nesting multiple `SQLScalarColumnTransformer`s on the same column, but it's possible you've already thought of that :-)

ferenc-hechler: > I propose we offer a subset of what you've written here with the following:
> `my_transformer = SQLScalarColumnTransformer(`

So, this is a fixed transformer class to be used inside a ColumnTransformer that wraps all non-"ML." transformations by defining the SQL text with a placeholder for the input column.

This would give us the full flexibility to generate any kind of column transformations in our transformer framework.

When reading in a TRANSFORM_ONLY model, all "ML." lines will be replaced with the corresponding Standard-Transformers and all other lines will be translated into SQLScalarColumnTransformer.

That sound good to me.

ferenc-hechler: > renaming this to `SQLScalarColumnTransformer` and accepting `sql` and columns as arguments to `__init__`?
yes, I will rewrite the PR.

Proposal: 
```
__init__(self, sql, columns, output_column_prefix="transformed_", output_columns=None)
```

So, the first syntax would be

```
SQLScalarColumnTransformer(
    "CASE WHEN {0} IS NULL THEN -1 ELSE LENGTH({0}) END",
    columns=["city", "county"],
    output_column_prefix="len_"
)
```

the second syntax would be 


```
SQLScalarColumnTransformer(
    "CASE WHEN {0} IS NULL THEN -1 ELSE LENGTH({0}) END",
    columns=["city", "county"],
    output_columns=["len_city", "county"]
)
```

if not needed I would omit the dtype param.

ferenc-hechler: > We'll maybe need some additional infrastructure in our compilation 
> to allow nesting multiple SQLScalarColumnTransformers on the same column, 
> but it's possible you've already thought of that :-)

ColumnTransformer (=BigQuery TRANSFORM_ONLY model) does not support nesting columns.
The only way to apply transformers iteratively on the same column is to create multiple ColumnTransformers and apply them iteratively.

bigframes/ml/compose.py
ferenc-hechler: > Rather than ValueError and NotImplementedError, 
> could we provide a general purpose UnknownTransformer class? [...]

The `NotImplementedError` was not changed from the original code.
So, I would leave it as it is.
And the ValueError should not happen any more, 
because any not `ML.xxx` sql will be translated into an `SQLScalarColumnTransformer`

bigframes/ml/compose.py
tswast: `__init__(self, sql, columns, output_column_prefix="transformed_", output_columns=None)`

I was imagining `columns` as the columns that go into the SQL formatter, so one could create features that combine multiple columns. For example:

```python
SQLScalarColumnTransformer(
    "CASE WHEN {0} IS NULL OR {1} IS NULL THEN -1 ELSE LENGTH({0}) + LENGTH({1}) END",
    columns=["city", "county"],
    output_name="len_city_and_county",
)
```

To transform multiple columns, one could create multiple SQLScalarColumnTransformer instances and use them in a Pipeline.

tswast: As with BQML, all other columns are passed through. So one wouldn't need to name all columns that need to be transformed in one instance.

ferenc-hechler: If I anyway have to mention the columns in the constructor, I could immediately replace the placeholders in the sql.
This is anyway possible. So I can access any column name in the sql.
But when the same sql should be applied to different columns the placeholder makes sense.

I am still thinking, that SQLScalarColumnTransformer is only viable inside a ColumnTransformer, right.

tswast: > I could immediately replace the placeholders in the sql

I was thinking about this more from the bigframes.pandas perspective where the DataFrame column names don't necessarily correspond to a single column, possibly a whole expression. With bigframes.ml, since ML models are more dependent on the correct names for transform and fit steps, this is indeed less of a concern.

I like your proposal, supporting both `output_column_prefix` and `output_columns`, given this bigframes.ml context.

That said, a problem I anticipate is that this approach could cause problems if someone uses the same output column name more than once, such as if they are applying several transforms to the same column with the intention of replacing the original column each time.

Perhaps we can detect when multiple transforms in a pipeline define the one of the same output columns for now? Or does this fail when we try to execute it in BigQuery anyway due to multiple columns with the same name?

Long-term, I think my concerns about nesting transforms will be better addressed as we refactor [bigframes.ml](http://bigframes.ml/) to use more of the bigframes.pandas expression logic.

ferenc-hechler: I see a problem reverse engineering the column parameter from the SQL when loading the model.

Just a quick example:

```
col_trans = ColumnTransformer([
    ("lenenc", SQLScalarColumnTransformer(CASE WHEN {0} IS NULL THEN -1 ELSE LENGTH({0}) END), ["city", "county"]),
    ("labelenc", LabelEncoder(), ['county', 'category_name']),
    ...
])
```

After Storing this ColumnTransformer into a TRANSFORM_ONLY model and loading it, it looks different:

col_trans = ColumnTransformer([
    ("lenenc", SQLScalarColumnTransformer(CASE WHEN city IS NULL THEN -1 ELSE LENGTH(city) END), [???]),
    ("lenenc", SQLScalarColumnTransformer(CASE WHEN county IS NULL THEN -1 ELSE LENGTH(county) END), [???]),
    ("labelenc", LabelEncoder(), ['county']),
    ("labelenc", LabelEncoder(), ['category_name']),
    ...
])

Column Transformer uses _parse_from_sql to retrieve the input-column name:

https://github.com/googleapis/python-bigquery-dataframes/blob/4221632e69f1e4866d5db3cc183021ae8f80b245/bigframes/ml/preprocessing.py#L70-L77

But for arbitrary SQL the input column can not be retrieved as simple as `sql[sql.find("(") + 1 : sql.find(")")]`

tswast: I'd rather not spend too much time trying to restore SQLScalarColumnTransformer as it was originally. Someone could hardcode multiple column names in the SQL string and not even really use the formatting feature.

What if we allowed `columns` to be empty? Alternatively, perhaps a placeholder value could be used for `columns` in such cases, such as `[None]`?

ferenc-hechler: Created a first version of the SQLScalarColumnTransformer:

https://github.com/ferenc-hechler/python-bigquery-dataframes/blob/a53b51ccb359f342a0605fb8feb3b32a3fd59dda/bigframes/ml/compose.py#L57-L76

tswast:
presubmit failure shows some type checker errors:

```
nox > Running session mypy
nox > Creating virtual environment (virtualenv) using python3.10 in .nox/mypy
nox > python -m pip install .
nox > python -m pip install tabulate jinja2 pytest-mock mypy pytest-asyncio pytest-xdist xarray google-cloud-testutils types-python-dateutil 'pytest<8.0.0dev' asyncmock types-protobuf openpyxl types-requests pytest-timeout mock pandas-stubs pytest-retry pytest-cov types-tabulate types-setuptools
nox > mypy bigframes tests/system tests/unit --check-untyped-defs --explicit-package-bases '--exclude="^third_party"'
[bigframes/ml/compose.py:182](https://cs.corp.google.com/piper///depot/google3/bigframes/ml/compose.py?l=182): error: Argument 1 to "add" of "set" has incompatible type "tuple[Any, SQLScalarColumnTransformer, str]"; expected "tuple[str, OneHotEncoder | StandardScaler | MaxAbsScaler | MinMaxScaler | KBinsDiscretizer | LabelEncoder | PolynomialFeatures | SimpleImputer, str | list[str]]"  [arg-type]
[bigframes/ml/compose.py:189](https://cs.corp.google.com/piper///depot/google3/bigframes/ml/compose.py?l=189): error: Incompatible types in assignment (expression has type "ColumnTransformer", variable has type "SQLScalarColumnTransformer")  [assignment]
[bigframes/ml/compose.py:192](https://cs.corp.google.com/piper///depot/google3/bigframes/ml/compose.py?l=192): error: Incompatible return value type (got "SQLScalarColumnTransformer", expected "ColumnTransformer")  [return-value]
[tests/unit/ml/test_compose.py:194](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=194): error: List item 0 has incompatible type "tuple[str, SQLScalarColumnTransformer, list[str]]"; expected "tuple[str, OneHotEncoder | StandardScaler | MaxAbsScaler | MinMaxScaler | KBinsDiscretizer | LabelEncoder | PolynomialFeatures | SimpleImputer, str | Iterable[str]]"  [list-item]
[tests/unit/ml/test_compose.py:199](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=199): error: List item 1 has incompatible type "tuple[str, SQLScalarColumnTransformer, list[str]]"; expected "tuple[str, OneHotEncoder | StandardScaler | MaxAbsScaler | MinMaxScaler | KBinsDiscretizer | LabelEncoder | PolynomialFeatures | SimpleImputer, str | Iterable[str]]"  [list-item]
[tests/unit/ml/test_compose.py:200](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=200): error: List item 2 has incompatible type "tuple[str, SQLScalarColumnTransformer, list[str]]"; expected "tuple[str, OneHotEncoder | StandardScaler | MaxAbsScaler | MinMaxScaler | KBinsDiscretizer | LabelEncoder | PolynomialFeatures | SimpleImputer, str | Iterable[str]]"  [list-item]
[tests/unit/ml/test_compose.py:225](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=225): error: List item 0 has incompatible type "tuple[str, SQLScalarColumnTransformer, list[str]]"; expected "tuple[str, OneHotEncoder | StandardScaler | MaxAbsScaler | MinMaxScaler | KBinsDiscretizer | LabelEncoder | PolynomialFeatures | SimpleImputer, str | Iterable[str]]"  [list-item]
[tests/unit/ml/test_compose.py:230](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=230): error: List item 1 has incompatible type "tuple[str, SQLScalarColumnTransformer, list[str]]"; expected "tuple[str, OneHotEncoder | StandardScaler | MaxAbsScaler | MinMaxScaler | KBinsDiscretizer | LabelEncoder | PolynomialFeatures | SimpleImputer, str | Iterable[str]]"  [list-item]
[tests/unit/ml/test_compose.py:231](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=231): error: List item 2 has incompatible type "tuple[str, SQLScalarColumnTransformer, list[str]]"; expected "tuple[str, OneHotEncoder | StandardScaler | MaxAbsScaler | MinMaxScaler | KBinsDiscretizer | LabelEncoder | PolynomialFeatures | SimpleImputer, str | Iterable[str]]"  [list-item]
[tests/unit/ml/test_compose.py:270](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=270): error: Argument "X" to "_compile_to_sql" of "SQLScalarColumnTransformer" has incompatible type "None"; expected "DataFrame"  [arg-type]
[tests/unit/ml/test_compose.py:279](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=279): error: Argument "X" to "_compile_to_sql" of "SQLScalarColumnTransformer" has incompatible type "None"; expected "DataFrame"  [arg-type]
[tests/unit/ml/test_compose.py:288](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=288): error: Argument "X" to "_compile_to_sql" of "SQLScalarColumnTransformer" has incompatible type "None"; expected "DataFrame"  [arg-type]
[tests/unit/ml/test_compose.py:304](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=304): error: Cannot assign to a method  [method-assign]
[tests/unit/ml/test_compose.py:449](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=449): error: Incompatible types in assignment (expression has type "ColumnTransformer | OneHotEncoder | StandardScaler | MaxAbsScaler | MinMaxScaler | KBinsDiscretizer | LabelEncoder | PolynomialFeatures | SimpleImputer", variable has type "ColumnTransformer")  [assignment]
[tests/unit/ml/test_compose.py:460](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=460): error: Incompatible types in assignment (expression has type "ColumnTransformer | OneHotEncoder | StandardScaler | MaxAbsScaler | MinMaxScaler | KBinsDiscretizer | LabelEncoder | PolynomialFeatures | SimpleImputer", variable has type "ColumnTransformer")  [assignment]
[tests/unit/ml/test_compose.py:485](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=485): error: Incompatible types in assignment (expression has type "BQMLModel", variable has type "BqmlModel | None")  [assignment]
[tests/unit/ml/test_compose.py:508](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=508): error: List item 0 has incompatible type "tuple[str, SQLScalarColumnTransformer, list[str]]"; expected "tuple[str, OneHotEncoder | StandardScaler | MaxAbsScaler | MinMaxScaler | KBinsDiscretizer | LabelEncoder | PolynomialFeatures | SimpleImputer, str | Iterable[str]]"  [list-item]
[tests/unit/ml/test_compose.py:513](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=513): error: List item 1 has incompatible type "tuple[str, SQLScalarColumnTransformer, list[str]]"; expected "tuple[str, OneHotEncoder | StandardScaler | MaxAbsScaler | MinMaxScaler | KBinsDiscretizer | LabelEncoder | PolynomialFeatures | SimpleImputer, str | Iterable[str]]"  [list-item]
[tests/unit/ml/test_compose.py:514](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=514): error: List item 2 has incompatible type "tuple[str, SQLScalarColumnTransformer, list[str]]"; expected "tuple[str, OneHotEncoder | StandardScaler | MaxAbsScaler | MinMaxScaler | KBinsDiscretizer | LabelEncoder | PolynomialFeatures | SimpleImputer, str | Iterable[str]]"  [list-item]
[tests/unit/ml/test_compose.py:518](https://cs.corp.google.com/piper///depot/google3/tests/unit/ml/test_compose.py?l=518): error: Argument 1 to "_compile_to_sql" of "ColumnTransformer" has incompatible type "None"; expected "DataFrame"  [arg-type]
```

ferenc-hechler:
> mypy bigframes tests/system tests/unit --check-untyped-defs --explicit-package-bases '--exclude="^third_party"'

Fixed the typing errors.

bigframes/ml/compose.py
```diff
@@ -46,6 +46,28 @@
 )
 
 
+class SQLScalarColumnTransformer(base.BaseTransformer):
```
tswast, line 49: Please make a docstring for this class, ideally with an examples section that passes doctest. See: https://github.com/googleapis/python-bigquery-dataframes/blob/6b34244e9ebe747161da21af2f31ee537b653093/third_party/bigframes_vendored/pandas/core/frame.py#L245-L275 for an example of a google-style docstring.

bigframes/ml/compose.py
```diff
@@ -46,6 +46,28 @@
...
+    def __init__(self, sql: str, target_column="transformed_{0}"):
+        super().__init__()
+        self.sql = sql
+        self.target_column = target_column
```
tswast, line 53: Are these intended to be public properties? If so, could you make them private here and add a proper `@property` getter for each.

If not, please make these "private"

```suggestion
        self._sql = sql
        self._target_column = target_column
```

bigframes/ml/compose.py
```diff
@@ -79,13 +105,25 @@ def _keys(self):
...
         """The collection of transformers as tuples of (name, transformer, column)."""
         result: List[
             Tuple[
                 str,
-                Union[preprocessing.PreprocessingType, impute.SimpleImputer],
+                Union[
+                    preprocessing.PreprocessingType,
+                    impute.SimpleImputer,
+                    SQLScalarColumnTransformer,
+                ],
```
tswast, line 126: We're getting enough of these now that it's probably worth creating a constant to represent this union of types.

See: https://github.com/googleapis/python-bigquery-dataframes/blob/6b34244e9ebe747161da21af2f31ee537b653093/bigframes/dtypes.py#L31-L38 for an example.

bigframes/ml/compose.py
```diff
@@ -234,6 +295,22 @@ def fit(
         self._extract_output_names()
         return self
 
+    # Overwrite the implementation in BaseTransformer, as it only supports the "ML." transformers.
+    # TODO: clarify if this should be changed in base.BaseTransformer?
```
tswast, line 299: Fixing this in the base class would make sense to me. I see no technical reason why BaseTransformer should be restricted to `ML.*` transformers.

tswast (commented):
Looking good, thanks! A few comments. 

@GarrettWu could you take a look, too?

.project
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
GarrettWu, line 1: Are these 2 local dev files? Should they be excluded?

tests/unit/ml/test_compose.py
```diff
@@ -173,3 +179,355 @@ def test_columntransformer_repr_matches_sklearn():
     )
 
     assert bf_column_transformer.__repr__() == sk_column_transformer.__repr__()
+
```
GarrettWu, line 182: Thanks for adding unit tests. Can you also add system tests to verify that 
1. .fit() and .transform() work as expected in standalone transformer
2. .fit() and .predict() work as expected in pipeline
3. to_gbq() and the returned(loaded) transformer works as expected, if supported.

You can find examples in tests/system/

bigframes/ml/compose.py
GarrettWu, line 49: If inherits from BaseTransformer, then to_gbq() method is provided. While I don't think this class supports the operation as it is only one column SQL.

ferenc-hechler:
Ready for next review.   :-)

ferenc-hechler:
ready for next review   :-)

ferenc-hechler:
Sorry, was not aware, that nox can also check the docs.

```
 nox -s docs
```

is now successful locally.

bigframes/ml/compose.py
```diff
@@ -46,6 +46,93 @@
...
+
+    Create a single column transformer in plain sql.
+    This transformer can only be used inside ColumnTransformer.
+
+    When creating an instance '{0}' can be used as placeholder
+    for the column to transform:
+
+        SQLScalarColumnTransformer("{0}+1")
+
+    The default target column gets the prefix 'transformed\_'
```
tswast, line 61: Looks like the linter is complaining about this one:

```suggestion
    The default target column gets the prefix 'transformed_'
```

Alternatively, maybe switching to an `r"""` above will fix that?

bigframes/ml/compose.py
```diff
@@ -46,6 +46,93 @@
...
+
+    Create a single column transformer in plain sql.
+    This transformer can only be used inside ColumnTransformer.
+
+    When creating an instance '{0}' can be used as placeholder
+    for the column to transform:
+
+        SQLScalarColumnTransformer("{0}+1")
+
+    The default target column gets the prefix 'transformed_'
```
tswast, line 61: Let's try `r` here to fix the docs.

```suggestion
    r"""
    Wrapper for plain SQL code contained in a ColumnTransformer.

    Create a single column transformer in plain sql.
    This transformer can only be used inside ColumnTransformer.

    When creating an instance '{0}' can be used as placeholder
    for the column to transform:

        SQLScalarColumnTransformer("{0}+1")

    The default target column gets the prefix 'transformed\_'
```

bigframes/ml/compose.py
```diff
@@ -46,6 +46,93 @@
...
+    def _compile_to_sql(
+        self, X: bpd.DataFrame, columns: Optional[Iterable[str]] = None
+    ) -> List[str]:
+        if columns is None:
+            columns = X.columns
+        result = []
+        for column in columns:
+            current_sql = self._sql.format(column)
+            current_target_column = self._target_column.format(column)
+            result.append(f"{current_sql} AS {current_target_column}")
```
tswast, line 112: Because of the flexible column names feature, `column` and `current_target_column` could contain spaces and other characters. https://cloud.google.com/bigquery/docs/schemas#flexible-column-names

Let's quote these with backticks to support such column names.

```suggestion
            current_sql = self._sql.format(f"`{column}`")
            current_target_column = self._target_column.format(column)
            result.append(f"{current_sql} AS `{current_target_column}`")
```

tswast (commented):
Love it! One suggestion and then let's make sure the `e2e` tests pass before merging.

tswast (approved):
Thanks so much for your contribution and patience!

bigframes/ml/compose.py
```diff
@@ -46,6 +46,101 @@
...
+        <BLANKLINE>
+        [3 rows x 2 columns]
+
+    SQLScalarColumnTransformer can be combined with other transformers, like StandardScaler:
+
+        >>> col_trans = ColumnTransformer([
+        ...     ("identity", SQLScalarColumnTransformer("{0}", target_column="{0}"), ["col1", col5"]),
+        ...     ("increment", SQLScalarColumnTransformer("{0}+1", target_column="inc_{0}"), "col2"),
+        ...     ("stdscale", preprocessing.StandardScaler(), "col3"),
+        ...     ...
```
tswast, line 93: Not sure if this will cause a doctest failure. Maybe put in a comment?

```suggestion
        ...     # ...
```

bigframes/ml/compose.py
```diff
@@ -46,6 +46,101 @@
...
+        0                 5                 8
+        1                15                 6
+        2                 4                15
+        <BLANKLINE>
+        [3 rows x 2 columns]
+
+    SQLScalarColumnTransformer can be combined with other transformers, like StandardScaler:
+
+        >>> col_trans = ColumnTransformer([
+        ...     ("identity", SQLScalarColumnTransformer("{0}", target_column="{0}"), ["col1", col5"]),
```
tswast, line 90: Syntax error.

```suggestion
        ...     ("identity", SQLScalarColumnTransformer("{0}", target_column="{0}"), ["col1", "col5"]),
```

```
__________ [doctest] bigframes.ml.compose.SQLScalarColumnTransformer ___________
[gw2] linux -- Python 3.12.0 /tmpfs/src/github/python-bigquery-dataframes/.nox/doctest/bin/python
080            transformed_name  transformed_city
081         0                 5                 8
082         1                15                 6
083         2                 4                15
084         <BLANKLINE>
085         [3 rows x 2 columns]
086 
087     SQLScalarColumnTransformer can be combined with other transformers, like StandardScaler:
088 
089         >>> col_trans = ColumnTransformer([
UNEXPECTED EXCEPTION: SyntaxError('unterminated string literal (detected at line 2)', ('<doctest bigframes.ml.compose.SQLScalarColumnTransformer[7]>', 2, 87, '    ("identity", SQLScalarColumnTransformer("{0}", target_column="{0}"), ["col1", col5"]),', 2, 87))
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/doctest.py", line 1357, in __run
    exec(compile(example.source, filename, "single",
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<doctest bigframes.ml.compose.SQLScalarColumnTransformer[7]>", line 2
    ("identity", SQLScalarColumnTransformer("{0}", target_column="{0}"), ["col1", col5"]),
                                                                                      ^
SyntaxError: unterminated string literal (detected at line 2)
/[tmpfs/src/github/python-bigquery-dataframes/bigframes/ml/compose.py:89](https://cs.corp.google.com/piper///depot/google3/tmpfs/src/github/python-bigquery-dataframes/bigframes/ml/compose.py?l=89): UnexpectedException
FAILED bigframes/ml/compose.py::bigframes.ml.compose.SQLScalarColumnTransformer
```

bigframes/ml/compose.py
```diff
@@ -130,8 +225,10 @@ def camel_to_snake(name):
             if "transformSql" not in transform_col_dict:
                 continue
             transform_sql: str = transform_col_dict["transformSql"]
-            if not transform_sql.startswith("ML."):
-                continue
+
+            # workaround for bug in bq_model returning " AS `...`" suffix for flexible names
+            if cls.AS_FLEXNAME_SUFFIX_RX.match(transform_sql):
+                transform_sql = cls.AS_FLEXNAME_SUFFIX_RX.match(transform_sql).group(1)
```
tswast, line 231: Getting some mypy errors here. I think because mypy isn't smart enough to see the null check here. Let's try this instead:

```suggestion
            flex_name_match = cls.AS_FLEXNAME_SUFFIX_RX.match(transform_sql)
            if flex_name_match:
                transform_sql = flex_name_match.group(1)
```

```
nox > mypy bigframes tests/system tests/unit --check-untyped-defs --explicit-package-bases '--exclude="^third_party"'
[bigframes/ml/compose.py:231](https://cs.corp.google.com/piper///depot/google3/bigframes/ml/compose.py?l=231): error: Item "None" of "Match[str] | None" has no attribute "group"  [union-attr]
Found 1 error in 1 file (checked 235 source files)
nox > Command mypy bigframes tests/system tests/unit --check-untyped-defs --explicit-package-bases '--exclude="^third_party"' failed with exit code 1
nox > Session mypy failed.
```

tswast (commented):
A couple of small fixes needed for the test suite.

tswast:
Notebook failure `FAILED notebooks/remote_functions/remote_function.ipynb:: - ResourceExhausted: 429 Too many operations are currently being executed, try again later.` not caused by this PR and can be ignored.

tswast:
This will be included in the 1.20.0 release. Keep an eye on https://github.com/googleapis/python-bigquery-dataframes/pull/1017

Thanks again for your help @ferenc-hechler !

ferenc-hechler:
@tswast  thanks for all your support and reviews 🥇 
It really helps us to have this solution implemented!

PR #956, opened 2024-09-04
test: retry streaming tests to accommodate flakiness

shobsi:

### DIFF
--- a/tests/system/large/test_streaming.py
+++ b/tests/system/large/test_streaming.py
@@ -14,10 +14,13 @@
 
 import time
 
+import pytest
+
 import bigframes
 import bigframes.streaming
 
 
+@pytest.mark.flaky(retries=2, delay=120)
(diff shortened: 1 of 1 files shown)


tests/system/large/test_streaming.py
```diff
@@ -14,10 +14,13 @@
 
 import time
 
+import pytest
+
 import bigframes
 import bigframes.streaming
 
 
+@pytest.mark.flaky(retries=2, delay=120)
```
GarrettWu, line 23: I think we don't need to apply delay. Or 10s could be enough unless the service is having a large outage.

GarrettWu (approved):

tests/system/large/test_streaming.py
shobsi: done, along with retry=3

PR #964, opened 2024-09-05
docs: add docstring return type section to BigQueryOptions class

arwas11:


bigframes/_config/bigquery_options.py
```diff
@@ -235,6 +268,10 @@ def kms_key_name(self) -> Optional[str]:
         Cloud KMS CryptoKey Encrypter/Decrypter IAM role in the key's project.
         For more information, see https://cloud.google.com/bigquery/docs/customer-managed-encryption#assign_role
         Assign the Encrypter/Decrypter.
+
+        Returns:
+            None or str:
+                Customer managed encryption key as a string; otherwise None.
```
tswast, line 274: It's not the actual key, it's the name of the key.

arwas11: Done.

bigframes/_config/sampling_options.py
```diff
@@ -33,14 +33,44 @@ class SamplingOptions:
...
 
     def with_max_download_size(self, max_rows: Optional[int]) -> SamplingOptions:
+        """Configures the maximum download size for data sampling in MB
+
+        Args:
+            max_rows (None or int):
+                An int value for the maximum row size.
+
+        Returns:
+            SamplingOptions:
```
tswast, line 43: Please use the full path as a customer would use it. Otherwise, Cloud RAD might not find it. In this case `bigframes._config.sampling_options.SamplingOptions`. That way it can link to https://cloud.google.com/python/docs/reference/bigframes/latest/bigframes._config.sampling_options.SamplingOptions

Note: in this case it's a bit less essential as it's on the same page, but I'd like us to set a good example for others to follow.

tswast (approved):

sycai (approved):

OKEAMAH (commented):
Okeamah.eth

PR #966, opened 2024-09-05
refactor: Simplify join node definition

TrevorBergeron:

### DIFF
--- a/bigframes/core/__init__.py
+++ b/bigframes/core/__init__.py
@@ -17,9 +17,8 @@ from dataclasses import dataclass
 import datetime
 import functools
 import io
-import itertools
 import typing
-from typing import Iterable, Optional, Sequence
+from typing import Iterable, Optional, Sequence, Tuple
 import warnings
 
 import google.cloud.bigquery
@@ -191,18 +190,13 @@ class ArrayValue:
             nodes.ConcatNode(children=tuple([self.node, *[val.node for val in other]]))
         )
 
-    def project_to_id(self, expression: ex.Expression, output_id: str):
+    def compute_values(self, assignments: Sequence[Tuple[ex.Expression, str]]):
         return ArrayValue(
-            nodes.ProjectionNode(
-                child=self.node,
-                assignments=(
-                    (
-                        expression,
-                        output_id,
-                    ),
-                ),
-            )
+            nodes.ProjectionNode(child=self.node, assignments=tuple(assignments))
         )
+
+    def project_to_id(self, expression: ex.Expression, output_id: str):
+        return self.compute_values(((expression, output_id),))
 
     def assign(self, source_id: str, destination_id: str) -> ArrayValue:
         if destination_id in self.column_ids:  # Mutate case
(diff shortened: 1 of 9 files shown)


bigframes/core/__init__.py
```diff
@@ -341,124 +335,32 @@ def _reproject_to_table(self) -> ArrayValue:
...
-            rows.append(row)
-
-        return ArrayValue.from_pyarrow(pa.Table.from_pylist(rows), session=self.session)
-
     def relational_join(
         self,
         other: ArrayValue,
-        join_def: join_def.JoinDefinition,
-    ) -> ArrayValue:
+        conditions: typing.Tuple[typing.Tuple[str, str], ...] = (),
```
tswast, line 341: [Wikipedia](https://en.wikipedia.org/wiki/Join_(SQL)) calls these predicates, or more specifically "join predicates". That said, I do see [Google SQL calls these join conditions](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#join_conditions).

Note: we will eventually want to support more than just equality, such as geospatial join predicates (https://carto.com/blog/guide-to-spatial-joins-and-predicates-with-sql), so `Tuple` doesn't seem like the right type.

bigframes/core/__init__.py
```diff
@@ -341,124 +335,32 @@ def _reproject_to_table(self) -> ArrayValue:
...
         join_node = nodes.JoinNode(
             left_child=self.node,
             right_child=other.node,
-            join=join_def,
+            conditions=conditions,
+            type=type,
         )
-        return ArrayValue(join_node)
+        l_size = len(self.node.schema)
+        l_mapping = {
```
tswast, line 351: I'm curious what the purpose of these mappings is? Could you give more explanation in a docstring, please?

A guess: is it so we don't actually have to explicitly rename the columns in the SQL compilation step? If so, would it be better to switch to some offset-based logic now instead of mapping strings?

bigframes/core/blocks.py
```diff
@@ -1080,33 +1080,33 @@ def aggregate_all_and_stack(
...
+            stacked_expr, (_, value_col_ids, passthrough_cols,) = unpivot(
+                expr_with_offsets,
+                row_labels=self.column_labels,
+                unpivot_columns=[tuple(self.value_columns)],
                 passthrough_columns=[*self.index_columns, offset_col],
             )
             index_aggregations = [
                 (ex.UnaryAggregation(agg_ops.AnyValueOp(), ex.free_var(col_id)), col_id)
-                for col_id in [*self.index_columns]
+                for col_id in passthrough_cols[:-1]
```
tswast, line 1091: Why every column except the last one? Could you have a comment here explaining, please?

bigframes/core/blocks.py
```diff
@@ -1368,23 +1366,17 @@ def calculate_pairwise_metric(self, op=agg_ops.CorrOp()):
...
-        index_col_ids = [
-            guid.generate_guid() for i in range(self.column_labels.nlevels)
-        ]
         input_count = len(self.value_columns)
         unpivot_columns = tuple(
-            (
-                guid.generate_guid(),
-                tuple(expr.column_ids[input_count * i : input_count * (i + 1)]),
-            )
+            tuple(expr.column_ids[input_count * i : input_count * (i + 1)])
```
tswast, line 1371: I'm happy to see this. Should make our SQL more deterministic.

bigframes/core/blocks.py
```diff
@@ -1604,7 +1596,7 @@ def promote_offsets(self, label: Label = None) -> typing.Tuple[Block, str]:
             Block(
                 expr,
                 index_columns=self.index_columns,
-                column_labels=self.column_labels.insert(0, label),
+                column_labels=self.column_labels.insert(len(self.column_labels), label),
```
tswast, line 1599: Is it going to be a problem that the label moves from the start to the end?

bigframes/core/blocks.py
TrevorBergeron: Added comment. These correspond to the passthrough_columns argument in unpivot.

bigframes/core/__init__.py
TrevorBergeron: Callers used to provide the input_id->output_id mapping themselves through the join_def. I'm slowly taking power away from callers to provide the internal ids, so instead of accepting mappings from caller, this method now provides them to callers. I do want to eventually move to entirely offset-based column addressing, but its a multi-step process.

bigframes/core/__init__.py
TrevorBergeron: Wikipedia uses the term "condition" plenty as well - seems to be an accepted term. As for spatial predicates - can we leave those for later? Not sure how yet how I would want to represent those. I'm sure we will have one or two more refactors by then as we move towards offset-based indexing.

bigframes/core/blocks.py
TrevorBergeron: caused a few issues - but all resolved now. I want nodes to add variables to the end as it preserves the offsets of the existing variables - this will make some planned bfet transformations simpler.

tswast (approved):

PR #973, opened 2024-09-10
feat: support describe for non-numerical type string

sycai:
This change adds the "include" param in the describe method. It can only be set as None(default) or 'all'

I only enabled "count" and "nunique" for string types. More non-numerical types can be added later.

### DIFF
--- a/bigframes/dataframe.py
+++ b/bigframes/dataframe.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import datetime
 import inspect
+import itertools
 import re
 import sys
 import textwrap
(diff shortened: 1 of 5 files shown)


bigframes/dataframe.py
```diff
@@ -1913,6 +1915,11 @@ def _reindex_rows(
     def _reindex_columns(self, columns):
         block = self._block
         new_column_index, indexer = self.columns.reindex(columns)
+
+        if indexer is None:
+            # The new index is the same as the old one. Do nothing.
```
tswast, line 1920: I'm curious how you discovered this. Worth separating out into it's own PR (with tests).

tswast (approved):
Thanks!

bigframes/dataframe.py
sycai: This was discovered when I wrote an ealier version of the "describe()": 

I didn't check the emptiness of aggregated dataframes, and always concated them in the end. Then there was an edge case where the original dataframe contains only numerical columns, and as a result I was basically reindexing the columns with their original order, which triggered this bug.

PR #977, opened 2024-09-11
refactor: Limit access to ArrayValue node field.

TrevorBergeron:


GarrettWu (approved):
Is ArrayValue class still needed? Now it is only a wrapper of BigFramesNode.

TrevorBergeron:
> Is ArrayValue class still needed? Now it is only a wrapper of BigFramesNode.

Its a wrapper yes, but a useful one. It allows us to change the implementation of the node structure without rewriting the callers (in most cases). Also, I plan to add a bit of name aliasing to ArrayValue while changing BigFramesNode to use integer ids.

PR #981, opened 2024-09-12
feat: add `subset` parameter to `DataFrame.dropna` to select which columns to consider

tswast:

Fixes internal issue 366248570
🦕


tswast:
Looks like some real test failures. I'll look at these later today.

tswast:
> Looks like some real test failures. I'll look at these later today.

Looks like there was a bug with `subset=None`. I believe this is now fixed.

third_party/bigframes_vendored/pandas/core/frame.py
```diff
@@ -1662,6 +1664,15 @@ def dropna(
             <BLANKLINE>
             [3 rows x 3 columns]
 
+        Define in which columns to look for missing values.
```
Genesis929, line 1667: From doctest.
@@ -1,5 +1,5 @@
    -       name        toy       born
    -1    Batman  Batmobile 1940-04-25
    -2  Catwoman   Bullwhip        NaT
    +       name        toy        born
    +1    Batman  Batmobile  1940-04-25
    +2  Catwoman   Bullwhip        <NA>
     <BLANKLINE>
     [2 rows x 3 columns]

tswast: Thanks, fixed! The diff got a little big because I pulled in https://github.com/googleapis/python-bigquery-dataframes/pull/987. It should be easier to review again once that is merged.

Genesis929 (approved):

bigframes/dataframe.py
```diff
@@ -2033,8 +2034,23 @@ def dropna(
...
+                f"subset only supported when axis=0. {constants.FEEDBACK_LINK}"
+            )
+
         if axis_n == 0:
-            result = block_ops.dropna(self._block, self._block.value_columns, how=how)  # type: ignore
+            # subset needs to be converted into column IDs, not column labels.
+            if subset is None:
+                subset_ids = None
+            elif not utils.is_list_like(subset):
+                subset_ids = [self._block.label_to_col_id[subset]]
```
Genesis929, line 2047: The result of this seems to be a list of tuples, and don't work.

tswast: This is supposed to cover the single column string case. A loop would give one character per iteration. I'll take a look at the error.

Genesis929: In single column case the subset_ids seems looks like [(col_id,)], I assume one loop is needed for the tuple?

Genesis929: From the defination ```def label_to_col_id(self) -> typing.Mapping[Label, typing.Sequence[str]]```, I think self._block.label_to_col_id[subset] will be a sequence(tuple)?  So I think a loop like for ```id in self._block.label_to_col_id[label]``` should be included?

tswast: Chatted offline. `self._block.label_to_col_id[subset]` is a tuple. I've corrected.

PR #985, opened 2024-09-13
fix: DataFrameGroupby.agg now works with unnamed tuples

TrevorBergeron:


tests/system/small/test_groupby.py
sycai, line 1: Maybe we could also add test cases for:
* v is not a tuple
* v is a tuple but its length is not 2
?

sycai (changes requested):

tests/system/small/test_groupby.py
TrevorBergeron: added

sycai (approved):

PR #987, opened 2024-09-13
refactor: remove circular dependencies preventing local doctest runs

tswast:
With this change I can once again run

```
pytest --doctest-modules third_party/bigframes_vendored/pandas/core/frame.py
```

Note: having multiple `version.py` files should be fine. release-please will update all such files it finds.

### DIFF
--- a/third_party/bigframes_vendored/version.py
+++ b/third_party/bigframes_vendored/version.py
@@ -1,4 +1,4 @@
-# Copyright 2023 Google LLC
+# Copyright 2022 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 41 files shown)


third_party/bigframes_vendored/version.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2022 Google LLC
```
sycai, line 1: nit: maybe 2024 ?

sycai (approved):

third_party/bigframes_vendored/version.py
tswast: Since it's copied from a filed that was copyright 2022, I'm inclined to leave this as-is. Not that there's much to copyright :-)

PR #990, opened 2024-09-16
chore: re-enable the unittest workflow in Owlbot configuration

chelsea-lin:
This change is re-enabling the unittest workflow in Owlbot configuration. This workflow was previously disabled in #948 but is now reintroduced due to a fix in Synthtool.


shobsi (commented):
Does it make something pass that is failing right now? Or it is redundant (we don't need anymore) because the hidden directories are honored by default now?

chelsea-lin:
@shobsi It is not fixing anything but it's nice to bring the exclude file back to sync up with the upstream template. The file was excluded because of the hidden directories bug was not fixed yet in the template but can be quick fixed in BF.

shobsi (approved):

PR #992, opened 2024-09-16
chore: Use fixture for list/str accessor tests

sycai:


tests/data/repeated.jsonl
```diff
@@ -0,0 +1,4 @@
+{"rowindex": 0, "list_col": [1]}
+{"rowindex": 1, "list_col": [1,2]}
+{"rowindex": 2, "list_col": [1,2,3]}
+{"rowindex": 3, "list_col": [1,2,3,4]}
```
chelsea-lin, line 4: To robust the array supports, it would be nice to test different dtypes of an array:
1. array of `int64`, `bool`, `bytes`, `date`, `datetime`, `geography`, `numeric`, `float64`, `time`, `timestamp`. 
2. array of an array
3. array of an struct, which has a known issue on the constructor: https://github.com/googleapis/python-bigquery/issues/1992

We can add more comprehensive tests for other data types in a follow-up PR.

chelsea-lin (commented):
Thank you!

tests/data/repeated.jsonl
sycai: Sounds good. Added some more common types.

shobsi (approved):

PR #996, opened 2024-09-17
test: restore original udf in the remote function test, use supported type in the test data

chelsea-lin:


tests/system/large/test_remote_function.py
```diff
@@ -1744,6 +1750,8 @@ def serialize_row(row):
                 }
             )
 
+        pd_result = pd_df.apply(serialize_row, axis=1)
```
shobsi, line 1753: duplicate

tests/system/large/test_remote_function.py
```diff
@@ -1726,12 +1726,18 @@ def test_df_apply_axis_1_complex(session, pd_df):
 
         def serialize_row(row):
             custom = {
-                "name": row.name.item() if hasattr(row.name, "item") else row.name,
-                "index": [
-                    idx.item() if hasattr(idx, "item") else idx for idx in row.index
-                ],
+                "name": tuple(
```
shobsi, line 1729: doesn't the existing test highlight a gap between pandas and BigFrames. I'm concerned that by changing the test we can make it pass but leave the original issue unaddressed

shobsi: After discussing offline, we are now changing just the data in the tests. The remote function tests involved in this PR have custom udf that inspects the index data types. In pandas the index type can be be numpy based  (e.g. `Int64`) or non-numpy (e.g. `int64`). But in bigframes the types are always numpy based, so that is what we are going to use for these tests.

TrevorBergeron (approved):

PR #998, opened 2024-09-18
feat: allow multiple columns input for llm models

GarrettWu:


bigframes/ml/llm.py
```diff
@@ -244,7 +244,7 @@ def predict(
 
         Args:
             X (bigframes.dataframe.DataFrame or bigframes.series.Series):
-                Input DataFrame or Series, which contains only one column of prompts.
+                Input DataFrame or Series, can contain one or more columns. If multiple columns in the DataFrame, it must contain a "prompt" column for prediction.
```
sycai, line 247: nit: "If multiple columns are in the DataFrame, they must ..." and for other docs too

tests/system/small/ml/test_llm.py
```diff
@@ -297,6 +311,20 @@ def test_embedding_generator_predict_series_success(
...
+    palm2_embedding_generator_model, llm_text_df
+):
+    llm_text_df["additional_col"] = 1
+    llm_text_df = llm_text_df.rename(columns={"prompt": "content"})
+    df = palm2_embedding_generator_model.predict(llm_text_df).to_pandas()
+    assert df.shape == (3, 5)
+    assert "text_embedding" in df.columns
+    series = df["text_embedding"]
+    value = series[0]
+    assert len(value) == 768
```
sycai, line 325: nit: maybe we could coalesce line 323 - 325 into a single line?
assert len(df[..][0]) == 768

tests/system/small/ml/test_llm.py
```diff
@@ -338,6 +366,27 @@ def test_text_embedding_generator_predict_default_params_success(
...
+):
+    llm_text_df["additional_col"] = 1
+    llm_text_df = llm_text_df.rename(columns={"prompt": "content"})
+    text_embedding_model = llm.TextEmbeddingGenerator(
+        model_name=model_name, connection_name=bq_connection, session=session
+    )
+    df = text_embedding_model.predict(llm_text_df).to_pandas()
+    assert df.shape == (3, 5)
+    assert "ml_generate_embedding_result" in df.columns
+    series = df["ml_generate_embedding_result"]
```
sycai, line 385: nit: coalesce the following lines?

sycai (changes requested):

bigframes/ml/llm.py
```diff
@@ -307,14 +307,10 @@ def predict(
...
 
-        if len(X.columns) != 1:
-            raise ValueError(
-                f"Only support one column as input. {constants.FEEDBACK_LINK}"
-            )
-
-        # BQML identified the column by name
-        col_label = cast(blocks.Label, X.columns[0])
-        X = X.rename(columns={col_label: "prompt"})
+        if len(X.columns) == 1:
```
shobsi, line 310: I think we should make another check in the else clause - that the multi-column input does have a "prompt" column. Also add test for that scenario

bigframes/ml/llm.py
GarrettWu: "it" refers to the DataFrame. Can add "are" in "If multiple columns are in the DataFrame"

tests/system/small/ml/test_llm.py
GarrettWu: sure, actually I'll rewrite the tests. Also some are already removed in a recent PR.

bigframes/ml/llm.py
GarrettWu: @tswast had a suggestion that we shouldn't do much client side checks. I'm trying to follow: if the error message is meaningful to the user, then rely on server side checks. Otherwise we have to wrap server error messages or return client side error messages.

bigframes/ml/llm.py
GarrettWu: done

tests/system/small/ml/test_llm.py
GarrettWu: done

sycai (approved):

PR #1005, opened 2024-09-20
refactor: re-arrange `describe()` logic into two helper methods

sycai:
It feels like an overkill to introduce class-level constants.


tswast (approved):
I like it, thanks!

PR #1010, opened 2024-09-21
feat: add bigframes.bigquery.approx_top_count

chelsea-lin:
feat: add bigframes.bigquery.approx_top_count

- [X] Fixes internal issue 363004193 

Fixes internal issue 363004193 🦕


bigframes/bigquery/__init__.py
```diff
@@ -272,6 +272,44 @@ def json_extract_array(
...
+    The number parameter specifies the number of elements returned.
+
+    Each `STRUCT` contains two fields. The first field (named `value`) contains an input
+    value. The second field (named `count`) contains an `INT64` specifying the number
+    of times the value was returned.
+
+    Returns `NULL` if there are zero input rows.
+
+    **Examples:**
+        >>> import bigframes.pandas as bpd
```
shobsi, line 293: I think we need newline after **Examples:** to render it nicely

chelsea-lin: Good catches! Fixed.

Genesis929 (approved):

tests/system/small/bigquery/test_approx_agg.py
```diff
@@ -0,0 +1,76 @@
+# Copyright 2024 Google LLC
```
shobsi, line 1: Are these tests deterministic, given they are about an approx function?

chelsea-lin: I observed consistent results from local running. The consistency might be due to the small dataset.

PR #1013, opened 2024-09-23
docs: limit pypi notebook to 7 days and add more info about differences with partial ordering mode

tswast:


tswast:
Notebook failure is "FAILED notebooks/remote_functions/remote_function_usecases.ipynb:: - ResourceExhausted: 429 Too many operations are currently being executed, try again later." not this notebook.

notebooks/dataframes/pypi.ipynb
```diff
@@ -40,6 +40,16 @@
    "execution_count": 2,
```
GarrettWu, line 40: <pre class="ql-syntax" spellcheck="false">Line #5.    last_30_days = now - datetime.timedelta(days=30)
</pre><p>This is not needed then.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/googleapis/python-bigquery-dataframes/pull/1013/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

GarrettWu (approved):

notebooks/dataframes/pypi.ipynb
tswast: <p>Thanks. I just noticed that the <code>max()</code> call is doing a 4+ TB query. I will try to fix that. The trouble with using <code>last_7_days</code> for both is that I was getting empty results for the deps DataFrame. I will use the 30 day comparison for that to avoid the expensive <code>max()</code> call.  </p>

PR #1018, opened 2024-09-24
feat: add DeprecationWarning for PaLM2TextEmbeddingGenerator

GarrettWu:


bigframes/deprecation.py
```diff
@@ -0,0 +1,92 @@
...
+        def decorator(func1):
+
+            if inspect.isclass(func1):
+                fmt1 = "Call to deprecated class {name}. {reason}"
+            else:
+                fmt1 = "Call to deprecated function {name}. {reason}"
+
+            @functools.wraps(func1)
+            def new_func1(*args, **kwargs):
+                warnings.simplefilter("always", DeprecationWarning)
```
tswast, line 48: Please don't override people's default filters. Just use FutureWarning as recommended here: https://docs.python.org/3/library/warnings.html

See this note:

> Changed in version 3.7: Previously DeprecationWarning and FutureWarning were distinguished based on whether a feature was being removed entirely or changing its behaviour. They are now distinguished based on their intended audience and the way they’re handled by the default warnings filters.

bigframes/deprecation.py
tswast, line 1: Can we put this in `bigframes._core`? I know we haven't been the best at not polluting the `bigframes.*` namespace with private stuff, but since this isn't an API intended to be used by end users, I'd like to keep it clear that it's "private" to our package.

tswast:
Consider using https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.deprecated instead, as that also works well with type checkers such as `mypy`.

bigframes/deprecation.py
GarrettWu: removed the file, using typing_extension instead.

GarrettWu:
> Consider using https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.deprecated instead, as that also works well with type checkers such as `mypy`.

Done

bigframes/deprecation.py
GarrettWu: Done.

tests/system/small/ml/test_llm.py
```diff
@@ -383,3 +384,11 @@ def test_llm_gemini_pro_score_params(llm_fine_tune_df_default_index):
         ],
         index=6,
     )
+
+
+def test_palm2_text_embedding_deprecated():
+    with pytest.warns(exceptions.ApiDeprecationWarning):
+        try:
+            llm.PaLM2TextEmbeddingGenerator()
+        except (Exception):
```
sycai, line 393: It feels like we are also ignoring all the exceptions? I wonder if we could just let it be raised and fail this test, if such a case occurs.

sycai (approved):

tests/system/small/ml/test_llm.py
GarrettWu: The backend is already unavailable now. The exception is expected. We only want to check the warnings.

PR #1020, opened 2024-09-25
feat: add ml.model_selection.cross_validate support

GarrettWu:

Fixes #360015492 🦕

### DIFF
--- a/bigframes/ml/model_selection.py
+++ b/bigframes/ml/model_selection.py
@@ -165,13 +165,13 @@ def cross_validate(
 
     result: dict[str, list] = {"test_score": [], "fit_time": [], "score_time": []}
     for X_train, X_test, y_train, y_test in cv.split(X, y):  # type: ignore
-        fit_start_time = time.time()
+        fit_start_time = time.perf_counter()
         estimator.fit(X_train, y_train)
-        fit_time = time.time() - fit_start_time
+        fit_time = time.perf_counter() - fit_start_time
 
-        score_start_time = time.time()
+        score_start_time = time.perf_counter()
         score = estimator.score(X_test, y_test)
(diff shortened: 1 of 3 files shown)


bigframes/ml/model_selection.py
```diff
@@ -147,3 +149,37 @@ def split(
...
+    if cv is None:
+        cv = KFold(n_splits=5)
+    elif isinstance(cv, int):
+        cv = KFold(n_splits=cv)
+
+    result: dict[str, list] = {"test_score": [], "fit_time": [], "score_time": []}
+    for X_train, X_test, y_train, y_test in cv.split(X, y):  # type: ignore
+        fit_start_time = time.time()
+        estimator.fit(X_train, y_train)
+        fit_time = time.time() - fit_start_time
```
Genesis929, line 170: Maybe use perf_counter

From https://superfastpython.com/time-time-vs-time-perf_counter/:

The time.time() function should no longer be used for benchmarking because it is adjustable, non-monotonic, and lower resolution.

Genesis929 (approved):

bigframes/ml/model_selection.py
```diff
@@ -147,3 +149,37 @@ def split(
...
+
+        score_start_time = time.time()
```
Genesis929, line 172: Maybe use perf_counter. From chatgpt
time.time(): It is subject to system clock adjustments or skew. The system time can be adjusted backward and forward by the operating system, which can lead to inaccurate or unreliable results when measuring short durations or intervals. This makes it less suitable for performance testing where precise and stable measurements are critical.

GarrettWu: Done.

PR #1023, opened 2024-09-26
refactor: move reader functions from `__init__.py` to a separate file under the pandas package

sycai:


bigframes/pandas/readers.py
tswast, line 1: Could we make this `[bigframes/pandas/io/parsers/readers.py](https://github.com/pandas-dev/pandas/blob/23c497bb2f7e05af1fda966e7fb04db942453559/pandas/io/parsers/readers.py#L720)` to better match the pandas code layout?

tswast: Or `bigframes/pandas/io/api.py` if we want to go one layer of indirection fewer

https://github.com/pandas-dev/pandas/blob/23c497bb2f7e05af1fda966e7fb04db942453559/pandas/__init__.py#L139

sycai: Sounds good. I put all the functions under the `api.py` file since they are relatively simple

chelsea-lin (approved):
LGTM overall. Please check the two failing tests in the pre-submit.

bigframes/pandas/readers.py
chelsea-lin: The docs of these methods are stayed at `third_party/bigframes_vendored/pandas/io/parsers/readers.py`. Should we keep them in the similar path?

PR #1024, opened 2024-09-26
feat: add `bigframes.bigquery.create_vector_index` to assist in creating vector index on `ARRAY<FLOAT64>` columns

tswast:

Fixes internal issue 365541165 🦕


bigframes/bigquery/_operations/search.py
```diff
@@ -0,0 +1,243 @@
...
+    Creates a new vector index on a column of a table.
+
+    This method calls the `CREATE VECTOR INDEX DDL statement
+    <https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_vector_index_statement>`_.
+
+    """
+    import bigframes.pandas
+
+    if index_name is None:
+        index_name = table_id.split(".")[-1]
```
TrevorBergeron, line 60: would reading into a TableReference be more robust?

tswast: Pretty sure it's exactly the same logic there, but makes sense from the perspective of having one place for that logic (in case "." is allowed as a character in the table name someday).

tswast: Done in latest commit (along with `identifier` logic).

TrevorBergeron (approved):

PR #1028, opened 2024-09-27
fix: remove pre-caching of remote function results

shobsi:
This will save redundant remote function execution as reported in b/370088754. The trade-off is that any remote function integration issues will be caughts only at a later point through a usage triggered sql execution. Why the caching is not working as indended in the reported use case in the bug? as per TrevorBergeron@ "the cached execution is useless when it is implicitly joined back to the base dataframe"

Fixes internal issue 370088754 🦕

### DIFF
--- a/bigframes/dataframe.py
+++ b/bigframes/dataframe.py
@@ -3567,11 +3567,7 @@ class DataFrame(vendored_pandas_frame.DataFrame):
                     ops.NaryRemoteFunctionOp(func=func), series_list[1:]
                 )
             result_series.name = None
-
-            # Return Series with materialized result so that any error in the remote
-            # function is caught early
-            materialized_series = result_series.cache()
-            return materialized_series
+            return result_series
 
         # Per-column apply
         results = {name: func(col, *args, **kwargs) for name, col in self.items()}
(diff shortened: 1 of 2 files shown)


GarrettWu (dismissed):

GarrettWu:
So after the change, when does the execution happens? 
```
df["col2"] = df["col1"].apply(remote_function)
```
or
```
df.to_gbq()
```

if at the later stage, if the user calls exec APIs (such as __repr__) multiple times, will the function be executed multiple times?

GarrettWu:
Is there a way that we can add test to ensure that with multiple operations, there won't be redundant function calls?

shobsi:
> So after the change, when does the execution happens?
> 
> ```
> df["col2"] = df["col1"].apply(remote_function)
> ```
> 
> or
> 
> ```
> df.to_gbq()
> ```
> 
> if at the later stage, if the user calls exec APIs (such as **repr**) multiple times, will the function be executed multiple times?

It would be the latter. After this change the behavior would be same as the standard bigframes behavior - if the result is cached it won't be executed, else it would be.

shobsi:
> Is there a way that we can add test to ensure that with multiple operations, there won't be redundant function calls?

It's tied to caching. Chatted with @TrevorBergeron that caching is not a guarantee at this point but more like an internal optimization. Asserting how many times a function was called is a bit of a stretc,h just like asseting how many number of sql jobs were run in an operation, so I'd skip adding a test. Let me know if you have s strong opinion otherwise.

GarrettWu (approved):

PR #1031, opened 2024-09-30
fix: fix generic error message when entering an incorrect column name

arwas11:

Fixes #328113227 🦕


bigframes/core/groupby/__init__.py
```diff
@@ -88,9 +88,12 @@ def __getitem__(
             keys = list(key)
         else:
             keys = [key]
-        columns = [
-            col_id for col_id, label in self._col_id_labels.items() if label in keys
-        ]
+
+        for col_id, label in self._col_id_labels.items():
+            if label in keys:
+                columns = [col_id]
```
chelsea-lin, line 94: The `col_id` should be append to the `columns` by calling `columns.append(col_id)`. Also, `columns=[]` need to be called to initialize it first.

tests/system/small/test_groupby.py
```diff
@@ -421,6 +421,20 @@ def test_dataframe_groupby_getitem(
     pd.testing.assert_series_equal(pd_result, bf_result, check_dtype=False)
 
 
+def test_dataframe_groupby_getitem_error(
+    scalars_df_index,
+    scalars_pandas_df_index,
+):
+    col_names = ["float64_col", "int64_col", "bool_col", "string_col"]
+    with pytest.raises(KeyError, match="\"Columns not found: 'not_in_group'\""):
```
chelsea-lin, line 429: Could you please add another tests with two unknown columns? Thanks!

chelsea-lin (approved):

PR #1032, opened 2024-10-01
fix: Ensure no double execution for to_pandas

TrevorBergeron:

### DIFF
--- a/tests/system/small/test_dataframe.py
+++ b/tests/system/small/test_dataframe.py
@@ -4831,17 +4831,23 @@ def test_dataframe_explode(col_names, ignore_index, session):
     metrics = session._metrics
     df = bpd.DataFrame(data, session=session)
     pd_df = df.to_pandas()
+    pd_result = pd_df.explode(col_names, ignore_index=ignore_index)
+    bf_result = df.explode(col_names, ignore_index=ignore_index)
+
+    # Check that to_pandas() results in at most a single query execution
     execs_pre = metrics.execution_count
+    bf_materialized = bf_result.to_pandas()
+    execs_post = metrics.execution_count
+
     pd.testing.assert_frame_equal(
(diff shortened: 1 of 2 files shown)


tests/system/small/test_dataframe.py
```diff
@@ -4821,20 +4821,27 @@ def test_to_gbq_table_labels(scalars_df_index):
...
+    execs_pre = metrics.execution_count
     pd.testing.assert_frame_equal(
         df.explode(col_names, ignore_index=ignore_index).to_pandas(),
         pd_df.explode(col_names, ignore_index=ignore_index),
         check_index_type=False,
         check_dtype=False,
     )
+    execs_post = metrics.execution_count
+    # we test this property on this method in particular as compilation
+    # is non-deterministic and won't use the query cache as implemented
```
sycai, line 4843: nit: I would also document that this is to check that the execution was performed only once

tests/system/small/test_dataframe.py
```diff
@@ -4821,20 +4821,27 @@ def test_to_gbq_table_labels(scalars_df_index):
...
         "A": [[0, 1, 2], [], [3, 4]],
         "B": 3,
         "C": [["a", "b", "c"], np.nan, ["d", "e"]],
     }
-    df = bpd.DataFrame(data)
+
+    metrics = session._metrics
+    df = bpd.DataFrame(data, session=session)
     pd_df = df.to_pandas()
+    execs_pre = metrics.execution_count
```
sycai, line 4834: nit: put this line right below line 4831 or move line 4831 above this line, just to keep together statements that are related

sycai (approved):

tests/system/small/test_dataframe.py
TrevorBergeron: reorganized a bit to make clear that only care about executions from the `to_pandas()` invocation.

PR #1039, opened 2024-10-02
chore: Add an ExperimentOptions class to control experiments

sycai:


jiaxunwu (approved):

chelsea-lin (approved):
LGTM!

chelsea-lin:
FYI @TrevorBergeron, the semantic operators would be launched as experimental APIs, when the configurations are enabled.

PR #1045, opened 2024-10-03
chore: Implement semantic map

sycai:


bigframes/operations/semantics.py
```diff
@@ -39,7 +39,7 @@ def filter(self, instruction: str, model):
                 "The {food} is healthy."
 
             model:
-                A LLM model provided by Bigframes ML package.
+                A GeminiTextGenerator provided by Bigframes ML package.
```
chelsea-lin, line 42: Could you please add an assert in the codes to check the model type is correct or not:
```
        if not isinstance(model, TextEmbeddingGenerator):
            raise TypeError(f"Expected a test embedding model but got: {type(model)}")
```
The semantics operators might required other model, such as `sem_index` needs an embedding model.

bigframes/operations/semantics.py
```diff
@@ -61,30 +114,13 @@ def filter(self, instruction: str, model):
                 raise ValueError(f"Column {column} not found.")
 
         # Replace column references with names.
-        instruction = instruction.format(**{col: col for col in columns})
+        user_instruction = user_instruction.format(**{col: col for col in columns})
 
         prompt_df = self._df.copy()
```
chelsea-lin, line 119: nit:  maybe we can just call `self._df[columns].copy()`? Though BF has optimizations already to skip unused columns in query.

bigframes/operations/semantics.py
```diff
@@ -50,8 +50,61 @@ def filter(self, instruction: str, model):
...
+            results["ml_generate_text_llm_result"].str.lower().str.contains("true")
+        ]
+
+    def map(self, instruction: str, model):
+        """
+        Filters the DataFrame with the semantics of the user instruction.
+
+        Args:
+            instruction:
+                An instruction on how to filter the data. This value must contain
```
chelsea-lin, line 71: The document may be copied from `filter`?

bigframes/operations/semantics.py
```diff
@@ -50,8 +50,61 @@ def filter(self, instruction: str, model):
...
+
+        return self._df[
+            results["ml_generate_text_llm_result"].str.lower().str.contains("true")
+        ]
+
+    def map(self, instruction: str, model):
+        """
+        Filters the DataFrame with the semantics of the user instruction.
+
+        Args:
```
chelsea-lin, line 69: Could we please add some simple examples in the doc string?

bigframes/operations/semantics.py
```diff
@@ -50,8 +50,61 @@ def filter(self, instruction: str, model):
...
+        results = typing.cast(
+            Series,
+            model.predict(self._make_prompt(instruction, output_instruction))[
+                "ml_generate_text_llm_result"
+            ],
+        )
+
+        from bigframes.core.reshape import concat
+
+        return concat([self._df, results.rename("map_result")], axis=1)
```
chelsea-lin, line 103: I am not sure which one is better:
1. Find a special name, such as `_semantic_map_result_<index>`? Also check if `_semantic_map_result_<index>` is a new column for the existing df.
2. Ask user to provide a custom column name?

tests/system/small/operations/test_semantics.py
```diff
@@ -62,3 +62,53 @@ def test_filter_invalid_instruction_raise_error(instruction, gemini_flash_model)
...
+        gemini_flash_model,
+    ).to_pandas()
+    # Result sanitation
+    actual_df["map_result"] = actual_df["map_result"].str.strip().str.lower()
+
+    expected_df = pd.DataFrame(
+        {
+            "ingredient_1": ["Burger Bun", "Soy Bean"],
+            "ingredient_2": ["Beef Patty", "Bittern"],
+            "map_result": ["burger", "tofu"],
```
chelsea-lin, line 88: Can LM return the same results for this questions? Maybe just ask `Which city is the capital of {country}`?

chelsea-lin (changes requested):

bigframes/operations/semantics.py
sycai: Type check added.

bigframes/operations/semantics.py
sycai: SG!

bigframes/operations/semantics.py
sycai: Good catch!

bigframes/operations/semantics.py
sycai: Sure!

bigframes/operations/semantics.py
sycai: Asking for column name inputs sounds better

tests/system/small/operations/test_semantics.py
sycai: For now LM responses are pretty stable for obvious questions. I think we should be fine here.

bigframes/operations/semantics.py
```diff
@@ -49,9 +67,89 @@ def filter(self, instruction: str, model):
...
+
+        results = typing.cast(
+            DataFrame, model.predict(self._make_prompt(instruction, output_instruction))
+        )
+
+        return self._df[
+            results["ml_generate_text_llm_result"].str.lower().str.contains("true")
+        ]
+
+    def map(self, instruction: str, result_column_name: str, model):
```
chelsea-lin, line 84: nit: output_column. This is trying to align with pandas style.

bigframes/operations/semantics.py
```diff
@@ -61,30 +159,20 @@ def filter(self, instruction: str, model):
                 raise ValueError(f"Column {column} not found.")
 
         # Replace column references with names.
-        instruction = instruction.format(**{col: col for col in columns})
+        user_instruction = user_instruction.format(**{col: col for col in columns})
 
-        prompt_df = self._df.copy()
+        prompt_df = self._df[columns].copy()
```
chelsea-lin, line 164: When `columns` specifies only a single column, let's verify its data type. If it's a single string (rather than a list),  `_df[columns]` might return a Pandas Series instead of a DataFrame. To ensure this behavior is correct, could we please a test case to double-check this scenario? Thanks!

chelsea-lin (approved):
Left few comments. Overall LGTM. Thanks!

bigframes/operations/semantics.py
sycai: Sure. Test added.

bigframes/operations/semantics.py
sycai: Good idea!

PR #1048, opened 2024-10-03
feat: update LLM generators to warn user about model name instead of raising error.

Genesis929:

### DIFF
--- a/bigframes/ml/llm.py
+++ b/bigframes/ml/llm.py
@@ -154,8 +154,11 @@ class PaLM2TextGenerator(base.BaseEstimator):
             )
 
         if self.model_name not in _TEXT_GENERATOR_ENDPOINTS:
-            raise ValueError(
-                f"Model name {self.model_name} is not supported. We only support {', '.join(_TEXT_GENERATOR_ENDPOINTS)}."
+            warnings.warn(
+                f"Model name '{self.model_name}' is not supported. "
+                f"We currently support the following models: {', '.join(_TEXT_GENERATOR_ENDPOINTS)}. "
+                "However, model names can change, and the supported models may be outdated. "
+                "You can try using this model name, but it might result in unexpected errors."
             )
 
         options = {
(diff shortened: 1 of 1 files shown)


bigframes/ml/llm.py
```diff
@@ -154,8 +154,11 @@ def _create_bqml_model(self):
             )
 
         if self.model_name not in _TEXT_GENERATOR_ENDPOINTS:
-            raise ValueError(
-                f"Model name {self.model_name} is not supported. We only support {', '.join(_TEXT_GENERATOR_ENDPOINTS)}."
+            warnings.warn(
+                f"Model name '{self.model_name}' is not supported. "
+                f"We currently support the following models: {', '.join(_TEXT_GENERATOR_ENDPOINTS)}. "
```
shobsi, line 159: Can we say "We currently support" -> "We are currently aware of"

bigframes/ml/llm.py
```diff
@@ -154,8 +154,11 @@ def _create_bqml_model(self):
...
+                "However, model names can change, and the supported models may be outdated. "
+                "You can try using this model name, but it might result in unexpected errors."
```
shobsi, line 161: Can we say "You can try using this model name, but it might result in unexpected errors." -> "You should use this model name only if you are sure that it is supported in BigQuery."

@GarrettWu

bigframes/ml/llm.py
```diff
@@ -484,8 +487,11 @@ def _create_bqml_model(self):
             )
 
         if self.model_name not in _PALM2_EMBEDDING_GENERATOR_ENDPOINTS:
-            raise ValueError(
-                f"Model name {self.model_name} is not supported. We only support {', '.join(_PALM2_EMBEDDING_GENERATOR_ENDPOINTS)}."
+            warnings.warn(
+                f"Model name '{self.model_name}' is not supported. "
+                f"We currently support the following models: {', '.join(_PALM2_EMBEDDING_GENERATOR_ENDPOINTS)}. "
+                "However, model names can change, and the supported models may be outdated. "
```
shobsi, line 493: The last two lines are common, we can define it once as a constant

bigframes/ml/llm.py
Genesis929: Updated. That sounds better, thanks.

bigframes/ml/llm.py
Genesis929: Updated.

bigframes/ml/llm.py
Genesis929: The warning message is updated to be a template.

shobsi (approved):

PR #1051, opened 2024-10-04
chore: implement semantic join

sycai:


bigframes/operations/semantics.py
```diff
@@ -170,9 +167,93 @@ def _make_prompt(self, user_instruction: str, output_instruction: str):
...
+
+        for col in columns:
+            if col in self._df.columns and col in other.columns:
+                raise ValueError(f"Ambiguous column reference: {col}")
+            elif col in self._df.columns:
+                left_columns.append(col)
+            elif col in other.columns:
+                right_columns.append(col)
+            elif col.endswith("_left") and col[: -len("_left")] in self._df.columns:
+                left_columns.append(col)
```
chelsea-lin, line 232: `col` -> `col[: -len("_left")]`? If so, could you please add a test to cover these two lines?

bigframes/operations/semantics.py
```diff
@@ -170,9 +167,93 @@ def _make_prompt(self, user_instruction: str, output_instruction: str):
...
+            elif col.endswith("_right") and col[: -len("_right")] in other.columns:
+                right_columns.append(col)
+            else:
+                raise ValueError(f"Column {col} not found")
+
+        if not left_columns or not right_columns:
```
chelsea-lin, line 238: Please double check, the `left_columns` and `right_columns` is not referred in the `merge` operations.

chelsea-lin (changes requested):

bigframes/operations/semantics.py
sycai: This is intentional for overlapping columns, as they will be appended with "_left" and "_right" suffixes. Otherwise, the semantic filter would complain that the renamed columns are not found.

The self join test should've covered this

bigframes/operations/semantics.py
sycai: Yeah this is intentional. Since we are using cross join, we are simply doing a cartesian product on rows from the two data frames. There's no need to use a join key.

These variables are here to help check that columns from both data frames are referenced.

bigframes/operations/semantics.py
sycai: Plus, they are not used in the join, so the names shouldn't matter :)

bigframes/operations/semantics.py
```diff
@@ -170,9 +167,93 @@ def _make_prompt(self, user_instruction: str, output_instruction: str):
...
+            raise ValueError()
```
chelsea-lin, line 239: nit: please add error message here.

tests/system/large/operations/test_semantics.py
```diff
@@ -160,3 +160,100 @@ def test_map_invalid_model_raise_error():
...
+    )
+
+
+@pytest.mark.parametrize(
+    "instruction",
+    [
+        "No column reference",
+        pytest.param("{city} is in {continent}", id="non_existing_column"),
+        pytest.param("{city} is in {country}", id="ambiguous_column"),
+        pytest.param("{city_left} is in {country}", id="suffix_on_unique_column"),
```
chelsea-lin, line 235: `city_left` didn't raise an error at first according to: https://screenshot.googleplex.com/3bMRpw7BzRNTV6Y

bigframes/operations/semantics.py
chelsea-lin: Got it! Yeah, they're referred by `semantics.filter` API.

bigframes/operations/semantics.py
chelsea-lin: Thanks! After talking offline, the `suffixes` parameter in the `merge` method can rewrite the column name here.

chelsea-lin (approved):
Left 2 small comments. Overall LGTM.

bigframes/operations/semantics.py
sycai: Done.

tests/system/large/operations/test_semantics.py
sycai: Updated the validation logic. Now we can handle this scenario properly.

PR #1054, opened 2024-10-04
chore: add row_limit for text generators

sycai:
This helps prevent unexpected large amount of data being processed by LLM

### DIFF
--- a/bigframes/ml/llm.py
+++ b/bigframes/ml/llm.py
@@ -598,6 +598,8 @@ class TextEmbeddingGenerator(base.BaseEstimator):
         connection_name (str or None):
             Connection to connect with remote service. str of the format <PROJECT_NUMBER/PROJECT_ID>.<LOCATION>.<CONNECTION_ID>.
             If None, use default connection in session context.
+        row_limit (int or None):
+            The maximum number of rows that this model can process.
     """
 
     def __init__(
@@ -608,6 +610,7 @@ class TextEmbeddingGenerator(base.BaseEstimator):
         ] = "text-embedding-004",
         session: Optional[bigframes.Session] = None,
         connection_name: Optional[str] = None,
+        row_limit: Optional[int] = None,
     ):
         self.model_name = model_name
         self.session = session or bpd.get_global_session()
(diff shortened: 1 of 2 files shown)


shobsi:
> This helps prevent unexpected large amount of data being processed by LLM

Is this to server a customer feedback or to support our test infra?

bigframes/ml/llm.py
```diff
@@ -598,6 +598,8 @@ class TextEmbeddingGenerator(base.BaseEstimator):
         connection_name (str or None):
             Connection to connect with remote service. str of the format <PROJECT_NUMBER/PROJECT_ID>.<LOCATION>.<CONNECTION_ID>.
             If None, use default connection in session context.
+        row_limit (int or None):
```
chelsea-lin, line 601: nit: `max_cols`, which is used in other pandas methods.

bigframes/ml/llm.py
```diff
@@ -598,6 +598,8 @@ class TextEmbeddingGenerator(base.BaseEstimator):
...
+            The maximum number of rows that this model can process.
```
chelsea-lin, line 602: nit: `The maximum number of rows that this model can process per call`. The model can be called multiple times.

chelsea-lin (approved):
LGTM with 2 nit comments.

bigframes/ml/llm.py
sycai: SG!

bigframes/ml/llm.py
sycai: Done

bigframes/ml/llm.py
```diff
@@ -598,8 +598,8 @@ class TextEmbeddingGenerator(base.BaseEstimator):
         connection_name (str or None):
             Connection to connect with remote service. str of the format <PROJECT_NUMBER/PROJECT_ID>.<LOCATION>.<CONNECTION_ID>.
             If None, use default connection in session context.
-        row_limit (int or None):
-            The maximum number of rows that this model can process.
+        max_rows (int or None):
+            The maximum number of rows that this model can process per call.
```
TrevorBergeron, line 602: My feeling is this is better as a bigframes-wide setting. Alternatively, if we are just focusing on sem-join cardinality explosion, we can apply a narrow limit just for that case

chelsea-lin:
Added `do not merge` for further API reviewing.

PR #1055, opened 2024-10-05
refactor: Introduce slice op to model array slicing

TrevorBergeron:


bigframes/core/rewrite.py
```diff
@@ -381,3 +383,169 @@ def common_selection_root(
...
+    if start or ((start is not None) and step < 0):
+        if start > 0 and step > 0:
+            start_cond = ops.ge_op.as_expr(forward_offsets, scalar_exprs.const(start))
+        elif start > 0 and step < 0:
+            start_cond = ops.le_op.as_expr(forward_offsets, scalar_exprs.const(start))
+        elif start < 0 and step > 0:
+            start_cond = ops.le_op.as_expr(
+                reverse_offsets, scalar_exprs.const(-start - 1)
+            )
+        elif start < 0 and step < 0:
```
shobsi, line 514: maybe leave this as `else #start < 0 and step < 0:` for readabaility, I needed to look through the conditions to be sure that `start_cond` is guaranteed to be defined at line 518

same for `stop` and `step`

bigframes/core/__init__.py
```diff
@@ -229,6 +225,23 @@ def order_by(self, by: Sequence[OrderingExpression]) -> ArrayValue:
     def reversed(self) -> ArrayValue:
         return ArrayValue(nodes.ReversedNode(child=self.node))
 
+    def slice(
+        self, start: Optional[int], stop: Optional[int], step: Optional[int]
+    ) -> ArrayValue:
+        if self.node.order_ambiguous and not (self.session._strictly_ordered):
+            warnings.warn(
+                "Window ordering may be ambiguous, this can cause unstable results.",
```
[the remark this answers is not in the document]
TrevorBergeron: Not really - all slice operations address some aspect of the total ordering, which may not be stable under partial ordering. The only case where it may not be worth warning is a pure reverse slice (df[::-1]) where offsets aren't being addressed.

bigframes/core/rewrite.py
shobsi: Thanks!

shobsi (approved):

PR #1057, opened 2024-10-07
feat: support array output in `remote_function`

shobsi:
  - `remote_function`: screen/5rMtCZVaUYKdqxP
  - `Series.apply`: screen/9HkKMuWxMvbbPgf 
  - `DataFrame.apply`: screen/BoXH9A7d4hGpETu

Fixes internal issue 298876217	 🦕


bigframes/session/__init__.py
```diff
@@ -1365,6 +1373,12 @@ def read_gbq_function(
                 Whether the function is a row processor. This is set to True
                 for a function which receives an entire row of a DataFrame as
                 a pandas Series.
+            output_type (type, default None):
+                Python type equivalent to the BigQuery return type. This is used
+                to coerce the return value of the BigQuery function to BigFrames
+                type. For example, if the BigQuery function returns a JSON
+                serialized array of integers such as "[1, 2, 3]", then user can
+                set `output_type=list[int]` to read it as array of integers [1, 2, 3].
```
TrevorBergeron, line 1381: I don't know if this makes sense for existing bq functions. Its more helpful for local python functions where the output type is otherwise ambiguous. Users should call astype or similar methods to explicitly coerce the result.

bigframes/series.py
```diff
@@ -1501,6 +1501,18 @@ def apply(
...
+        # string form
+        if bigframes.dtypes.is_array_like(func.output_dtype):
+            import bigframes.bigquery as bbq
+
+            result_dtype = bigframes.dtypes.arrow_dtype_to_bigframes_dtype(
+                func.output_dtype.pyarrow_dtype.value_type
+            )
+            result_series = bbq.json_extract_string_array(
+                result_series, value_dtype=result_dtype
+            )
```
TrevorBergeron, line 1514: would really rather we do this at a lower level

bigframes/session/__init__.py
shobsi: The place where it would be useful is reading back an remote function that was created using bigframes.  When they used it then their code didn't have to call astype, so we want to give them a way to have that code continue to work after reading back. The test `test_df_apply_axis_1_array_output` captures this usecase.

bigframes/functions/remote_function.py
```diff
@@ -198,9 +201,26 @@ def func(*bigframes_args, **bigframes_kwargs):
         )
     func.input_dtypes = tuple(function_input_dtypes)  # type: ignore
 
+    # set output bigframes data types
+    ibis_output_type = ibis_signature.output_type
+    if output_type is not None:
+        if not isinstance(ibis_signature.output_type, ibis.expr.datatypes.String):
+            raise TypeError(
```
shobsi, line 208: @tswast mentioned offline: "I would find that to be pretty annoying as a user. Better to make sure it matches and maybe warn that it's not necessary"

bigframes/session/__init__.py
shobsi: Remove this param in the latest change, PTAL

bigframes/functions/remote_function.py
shobsi: Removed output_type from read_gbq_function, PTAL

bigframes/dataframe.py
```diff
@@ -3956,6 +3956,19 @@ def apply(self, func, *, axis=0, args: typing.Tuple = (), **kwargs):
                     ops.NaryRemoteFunctionOp(func=func), series_list[1:]
                 )
             result_series.name = None
+
+            # if the output is an array, reconstruct it from the json serialized
+            # string form
+            if bigframes.dtypes.is_array_like(func.output_dtype):
```
TrevorBergeron, line 3962: Do we actually handle any array-like dtype?

bigframes/series.py
```diff
@@ -1513,6 +1513,18 @@ def apply(
             ops.RemoteFunctionOp(func=func, apply_on_null=True)
         )
 
+        # if the output is an array, reconstruct it from the json serialized
+        # string form
+        if bigframes.dtypes.is_array_like(func.output_dtype):
```
TrevorBergeron, line 1518: seems the code within this block assume not just array_like, but specifically that it is a pyarrow list_ type

bigframes/functions/_utils.py
```diff
@@ -202,13 +205,73 @@ def ibis_signature_from_python_signature(
...
+    except (TypeError, json.decoder.JSONDecodeError):
+        return None
+
+    try:
+        output_type = metadata_dict["value"]["python_output_type"]
+    except KeyError:
+        return None
+
+    try:
+        python_output_type = eval(output_type)
```
TrevorBergeron, line 245: `eval` always makes me a bit uncomfortable - can we do this in a more constrained way?

bigframes/functions/_utils.py
```diff
@@ -202,13 +205,73 @@ def ibis_signature_from_python_signature(
...
+    # Let's keep the actual metadata inside one level of nesting so that in
+    # future we can use a top level key "version" (parallel to "value"), so that
+    # "value" can be interpreted according to the "version". Absence of
+    # "version" should be interpreted as default version.
+    inner_metadata = {}
+    if python_output_type:
+        if typing.get_origin(python_output_type) is list:
+            python_output_type_ser = repr(python_output_type)
+        else:
+            python_output_type_ser = python_output_type.__name__
```
TrevorBergeron, line 262: shoudl we bother with non-list types right now?

bigframes/series.py
shobsi: That's exactly what the `array_like` implementation checks? https://github.com/googleapis/python-bigquery-dataframes/blob/5a2731bda8b2e9ea54bf582f823acdb6153dbb8f/bigframes/dtypes.py#L301-L304

bigframes/functions/_utils.py
shobsi: removed eval in the latest patch, PTAL

bigframes/functions/_utils.py
shobsi: throwing error for non-array and not-supported-array types in the latest patch, PTAL

bigframes/series.py
TrevorBergeron: eh, probably fine then, I don't really see array_like definition expanding anytime soon

TrevorBergeron (approved):

PR #1062, opened 2024-10-07
fix: correct zero row count in DataFrame from table view

Genesis929:

### DIFF
--- a/bigframes/core/nodes.py
+++ b/bigframes/core/nodes.py
@@ -508,6 +508,7 @@ class GbqTable:
     table_id: str = field()
     physical_schema: Tuple[bq.SchemaField, ...] = field()
     n_rows: int = field()
+    is_view: bool = field()
     cluster_cols: typing.Optional[Tuple[str, ...]]
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


bigframes/core/nodes.py
```diff
@@ -523,6 +524,7 @@ def from_table(table: bq.Table, columns: Sequence[str] = ()) -> GbqTable:
             table_id=table.table_id,
             physical_schema=schema,
             n_rows=table.num_rows,
+            is_view=(table.table_type == "VIEW"),
```
TrevorBergeron, line 527: I wonder if EXTERNAL and MATERIALIZED_VIEW have the same problem. https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.table.Table#google_cloud_bigquery_table_Table_table_type

TrevorBergeron (approved):

bigframes/core/nodes.py
Genesis929: As chatted offline, MATERIALIZED_VIEW will be skipped for now. 

And external table seems also shows row count zero if we read num_rows.

Logic updated.

PR #1066, opened 2024-10-08
chore: implement semantic similarity join

sycai:


bigframes/operations/semantics.py
```diff
@@ -527,6 +527,109 @@ def search(
...
+        else:
+            del join_result["distance"]
+
+        return join_result
+
+    @staticmethod
+    def _attach_embedding(dataframe, column: str, model):
+        result_df = dataframe.copy()
+        embeddings = model.predict(dataframe[column])["ml_generate_embedding_result"]
+        result_df["embedding"] = embeddings
```
chelsea-lin, line 630: To prevent naming conflicts in the agg method, if the dataframe already contains a column named embedding, the aggregation_group_id column should be assigned a unique identifier. This can be achieved by using a UUID generator like `bigframes.core.guid.generate_guid("agg")`.

bigframes/operations/semantics.py
```diff
@@ -527,6 +527,109 @@ def search(
...
+        other,
+        left_on: str,
+        right_on: str,
+        model,
+        top_k: int = 3,
+        score_column: Optional[str] = None,
+        max_rows: int = 1000,
+    ):
+        """
+        Joins two dataframes based on the similarity of the specified columns.
```
chelsea-lin, line 541: To provide more context and clarity for users, let's enhance the description of the`sim_join` method with details about its implementation and scalability. Please refer to `agg` method. Thanks

chelsea-lin (approved):
LGTM with two nit comments.

bigframes/operations/semantics.py
sycai: Good call. Though for the query table I have to hard code the "embedding" column.

bigframes/operations/semantics.py
sycai: SG. Done!

PR #1070, opened 2024-10-09
fix: access MATERIALIZED_VIEW with read_gbq

Genesis929:

### DIFF
--- a/bigframes/session/_io/bigquery/read_gbq_table.py
+++ b/bigframes/session/_io/bigquery/read_gbq_table.py
@@ -102,6 +102,7 @@ def validate_table(
     table_ref: bigquery.table.TableReference,
     columns: Optional[Sequence[str]],
     snapshot_time: datetime.datetime,
+    table_type: str,
     filter_str: Optional[str] = None,
 ) -> bool:
     """Validates that the table can be read, returns True iff snapshot is supported."""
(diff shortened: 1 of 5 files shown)


bigframes/session/_io/bigquery/read_gbq_table.py
```diff
@@ -124,6 +125,17 @@ def validate_table(
     if table_ref.dataset_id.startswith("_"):
         return False
 
+    # Materialized views，does not support snapshot
+    if table_type == "MATERIALIZED_VIEW":
+        warnings.warn(
```
shobsi, line 130: we should add a test to test this warning

Genesis929: Updated.

shobsi (approved):

tests/system/small/test_session.py
```diff
@@ -390,9 +390,16 @@ def test_read_gbq_twice_with_same_timestamp(session, penguins_table_id):
     assert df3 is not None
 
 
-def test_read_gbq_on_linked_dataset_warns(session):
+@pytest.mark.parametrize(
+    "source_table",
+    [
+        "bigframes-dev.thelook_ecommerce.orders",
+        "bigframes-dev.bigframes_tests_sys.base_table_mat_view",
```
tswast, line 397: I'm a concerned about hard coding this table. Could we at least include something in the `scripts/` directory that can recreate this table if needed?

Genesis929: Got it, will add that.

PR #1082, opened 2024-10-14
feat: Add warning when user tries to access struct series fields with `__getitem__`

sycai:


tests/system/small/core/test_indexers.py
```diff
@@ -0,0 +1,83 @@
...
+            pa.struct([("project", pa.string()), ("version", pa.int64())])
+        ),
+        session=session,
+    )
+
+    if should_warn:
+        with pytest.warns(UserWarning, match=r"Series\.struct\.field\(.+\)"):
+            s[key]
+    else:
+        s[key]
```
tswast, line 43: Should we have a check that a warning is _not_ raised in this case?

I would prefer two separate tests to make this easier to follow. See go/tott/661

tests/system/small/core/test_indexers.py
```diff
@@ -0,0 +1,83 @@
...
+            {"project": "pandas", "version": 1},
+        ],
+        dtype=bpd.ArrowDtype(
+            pa.struct([("project", pa.string()), ("version", pa.int64())])
+        ),
+        index=["p1"],
+        session=session,
+    )
+
+    s[key]
```
tswast, line 65: Please make this test fail if a warning is raised.

tests/system/small/core/test_indexers.py
```diff
@@ -0,0 +1,83 @@
...
+)
+def test_string_indexed_series_non_struct_accessor_no_warning(session, key):
+    s = bpd.Series(
+        [1],
+        dtype=bpd.Int64Dtype,
+        index=["a"],
+        session=session,
+    )
+
+    s[key]
```
tswast, line 83: Please make this test fail if a warning is raised.

tests/system/small/core/test_indexers.py
sycai: I separated the tests into two parts: 1 for warning and 1 for no warning. Hopefully this makes it easier to read and understand.

tests/system/small/core/test_indexers.py
sycai: Done

tests/system/small/core/test_indexers.py
sycai: Done

bigframes/core/indexers.py
```diff
@@ -391,6 +394,23 @@ def _perform_loc_list_join(
...
+        # No need to check series that do not have struct values
+        return
+
+    if not bigframes.dtypes.is_string_like(index.dtype):
+        # No need to check indexing with non-string values.
+        return
+
+    if not bigframes.dtypes.is_string_like(series.index.dtype):
+        warnings.warn(
+            "Are you trying to access struct fields? If so, please use Series.struct.field(...) method instead."
```
tswast, line 410: Could we do a custom category (defined in bigframes.exceptions) here to make it a bit easier to filter out if needed for some reason?

Also, I recommend setting stacklevel so that the error shows up in user code, not as appearing deep in BigFrames.

tests/system/small/core/test_indexers.py
tswast: Thanks! Maybe worth a comment about the filterwarnings above, as it'n not 100% obvious to me.

Alternatively, could we use the solution suggested here, instead?

https://docs.pytest.org/en/stable/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests

```suggestion
    with warnings.catch_warnings():
        warnings.simplefilter("error", category=UserWarning)
        s[key]
```

Better yet, have a custom category so we know precisely what warning to fail on.

sycai: Yeah this approach is simpler. Defined a dedicated warning class instead. I also set a stack level as suggested.

bigframes/core/indexers.py
sycai: I have provided a stack level such that the warning call site is now at `__getitem__`

bigframes/exceptions.py
```diff
@@ -73,3 +73,7 @@ class UnknownDataTypeWarning(Warning):
 
 class ApiDeprecationWarning(FutureWarning):
     """The API has been deprecated."""
+
+
+class IncorrectApiUsageWarning(Warning):
```
tswast, line 78: Let's be even more specific.

```suggestion
class BadIndexerKeyWarning(Warning):
```

bigframes/core/indexers.py
```diff
@@ -391,6 +395,25 @@ def _perform_loc_list_join(
...
+        return
+
+    if not bigframes.dtypes.is_string_like(index.dtype):
+        # No need to check indexing with non-string values.
+        return
+
+    if not bigframes.dtypes.is_string_like(series.index.dtype):
+        warnings.warn(
+            "Are you trying to access struct fields? If so, please use Series.struct.field(...) method instead.",
+            category=bigframes.exceptions.IncorrectApiUsageWarning,
```
tswast, line 412:
```suggestion
            category=bigframes.exceptions.BadIndexerKeyWarning,
```

tests/system/small/core/test_indexers.py
```diff
@@ -0,0 +1,94 @@
...
+        0,
+        "a",
+    ],
+)
+def test_struct_series_indexers_should_not_warn(request, series, key):
+    s = request.getfixturevalue(series)
+
+    with warnings.catch_warnings():
+        warnings.simplefilter(
+            "error", category=bigframes.exceptions.IncorrectApiUsageWarning
```
tswast, line 92:
```suggestion
            "error", category=bigframes.exceptions.BadIndexerKeyWarning
```

tests/system/small/core/test_indexers.py
```diff
@@ -0,0 +1,94 @@
...
+        [
+            {"project": "pandas", "version": 1},
+        ],
+        dtype=bpd.ArrowDtype(
+            pa.struct([("project", pa.string()), ("version", pa.int64())])
+        ),
+        session=session,
+    )
+
+    with pytest.warns(bigframes.exceptions.IncorrectApiUsageWarning):
```
tswast, line 68:
```suggestion
    with pytest.warns(bigframes.exceptions.BadIndexerKeyWarning):
```

tswast (approved):

PR #1084, opened 2024-10-14
fix: update session metrics in `read_gbq_query`

Genesis929:
This change makes sure the session metrics correctly reflect slot milliseconds after ML transform.

Fixes internal issue 372547905 🦕

### DIFF
--- a/bigframes/session/loader.py
+++ b/bigframes/session/loader.py
@@ -527,6 +527,9 @@ class GbqDataLoader:
             configuration=configuration,
         )
 
+        if self._metrics is not None:
+            self._metrics.count_job_stats(query_job)
+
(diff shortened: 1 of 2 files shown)


sycai (approved):

shobsi (approved):

bigframes/session/loader.py
```diff
@@ -529,6 +529,9 @@ def read_gbq_query(
             configuration=configuration,
         )
 
+        if self._metrics is not None:
```
shobsi, line 532: I missed earlier, is the query_job guaranteed to have finished at this point? We should wait for the job to finish, like we do in other places before calling `count_job_stats`.

PR #1085, opened 2024-10-15
fix: model.fit metric not collected issue.

Genesis929:

### DIFF
--- a/bigframes/ml/core.py
+++ b/bigframes/ml/core.py
@@ -273,8 +273,6 @@ class BqmlModelFactory:
     def _create_model_with_sql(self, session: bigframes.Session, sql: str) -> BqmlModel:
         # fit the model, synchronously
         _, job = session._start_query_ml_ddl(sql)
-        if session._metrics is not None:
-            session._metrics.count_job_stats(job)
 
         # real model path in the session specific hidden dataset and table prefix
         model_name_full = f"{job.destination.project}.{job.destination.dataset_id}.{job.destination.table_id}"
--- a/bigframes/ml/core.py
+++ b/bigframes/ml/core.py
@@ -273,7 +273,8 @@ class BqmlModelFactory:
     def _create_model_with_sql(self, session: bigframes.Session, sql: str) -> BqmlModel:
         # fit the model, synchronously
         _, job = session._start_query_ml_ddl(sql)
-        session._metrics.count_job_stats(job)
+        if session._metrics is not None:
+            session._metrics.count_job_stats(job)
 
         # real model path in the session specific hidden dataset and table prefix
         model_name_full = f"{job.destination.project}.{job.destination.dataset_id}.{job.destination.table_id}"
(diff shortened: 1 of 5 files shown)


bigframes/ml/core.py
```diff
@@ -273,6 +273,8 @@ def _create_model_ref(
     def _create_model_with_sql(self, session: bigframes.Session, sql: str) -> BqmlModel:
         # fit the model, synchronously
         _, job = session._start_query_ml_ddl(sql)
+        if session._metrics is not None:
+            session._metrics.count_job_stats(job)
```
shobsi, line 277: Should we not do this inside _start_query_ml_ddl? That way all ML DDLs will be accounted in the session metrics

tests/system/large/ml/test_linear_model.py
```diff
@@ -31,8 +31,14 @@ def test_linear_regression_configure_fit_score(penguins_df_default_index, datase
         ]
     ]
     y_train = df[["body_mass_g"]]
+
+    start_execution_count = df._block._expr.session._metrics.execution_count
+
     model.fit(X_train, y_train)
```
shobsi, line 37: I think we have the opportunity to make this work work all ML APIs - fit, score, predict, transform, .... We should write such tests for all APIs by writing the metrics collection logic in more central place in the code.

Genesis929: score, predict works, seems internally is calling other functions.

bigframes/ml/core.py
Genesis929: Updated, and added an assertion in test for register.

bigframes/session/__init__.py
```diff
@@ -1373,7 +1373,14 @@ def _start_query_ml_ddl(
         # https://cloud.google.com/bigquery/docs/customer-managed-encryption#encrypt-model
         job_config.destination_encryption_configuration = None
 
-        return bf_io_bigquery.start_query_with_client(self.bqclient, sql, job_config)
+        results_iterator, query_job = bf_io_bigquery.start_query_with_client(
```
shobsi, line 1376: I see `start_query_with_client` already has the stats update

https://github.com/googleapis/python-bigquery-dataframes/blob/fd06d31f5e7edab3ce3af9a4e3e4415a4463ee5d/bigframes/session/_io/bigquery/__init__.py#L245-L246

so feels like we could be adding double counting

tests/system/large/ml/test_linear_model.py
```diff
@@ -31,8 +31,14 @@ def test_linear_regression_configure_fit_score(penguins_df_default_index, datase
...
 
+    end_execution_count = df._block._expr.session._metrics.execution_count
+    assert end_execution_count - start_execution_count == 2
```
shobsi, line 40: Worth adding a comment why fit does 2 queries

tests/system/large/ml/test_linear_model.py
shobsi: Good to know, we could just pick one or two of the existing tests that does fit, score, predict, register and transform and assert everywhere that we are counting stats for those operations.

bigframes/session/__init__.py
Genesis929: It's not double counting, but yes here we can send the metric into the start_query_with_client, thanks for point that out.

tests/system/large/ml/test_linear_model.py
Genesis929: Test Added.

shobsi (approved):

PR #1088, opened 2024-10-15
feat: add support for pandas series & data frames as inputs for ml models. 

sycai:


bigframes/ml/utils.py
```diff
@@ -17,33 +17,41 @@
...
 import bigframes.pandas as bpd
 
 # Internal type alias
-ArrayType = Union[bpd.DataFrame, bpd.Series]
+InputArrayType = Union[bpd.DataFrame, bpd.Series, pd.DataFrame, pd.Series]
+OutputArrayType = Union[bpd.DataFrame, bpd.Series]
 
 
-def convert_to_dataframe(*input: ArrayType) -> Generator[bpd.DataFrame, None, None]:
+def convert_to_dataframe(
```
TrevorBergeron, line 30: Do you think we can merge the logic in this file into the core/convert module logic? Ideally we don't have pandas->bigframes logic in two places.

bigframes/ml/pipeline.py
```diff
@@ -101,8 +102,8 @@ def _from_bq(cls, session: bigframes.Session, bq_model: bigquery.Model) -> Pipel
 
     def fit(
         self,
-        X: Union[bpd.DataFrame, bpd.Series],
-        y: Optional[Union[bpd.DataFrame, bpd.Series]] = None,
+        X: Union[bpd.DataFrame, bpd.Series, pd.DataFrame, pd.Series],
+        y: Optional[Union[bpd.DataFrame, bpd.Series, pd.DataFrame, pd.Series]] = None,
```
TrevorBergeron, line 106: Do you think we can define a single annotation representing this set of types and use it everywhere. This will make it easier to accomodate additional types in the future

bigframes/ml/utils.py
```diff
@@ -17,33 +17,41 @@
...
+    *input: InputArrayType,
+) -> Generator[bpd.DataFrame, None, None]:
     return (_convert_to_dataframe(frame) for frame in input)
 
 
-def _convert_to_dataframe(frame: ArrayType) -> bpd.DataFrame:
+def _convert_to_dataframe(frame: InputArrayType) -> bpd.DataFrame:
```
TrevorBergeron, line 36: Should we also handle array-like data like numpy arrays or even plain python list/tuples?

bigframes/ml/utils.py
sycai: Good point! Though I think it's not very straightforward to do so. This is mainly because this function returns a Generator, while the one in the `core` package returns a single value. We will need some extra effort to make everything consistent (function names, parameter types, return types, etc). 

Considering that this RP is already not trivial, I think we can for now only focus on the feature delivery. I will migrate the conversion logic in another PR. Does it sound good to you?

bigframes/ml/pipeline.py
sycai: My main concern is that type aliases may not be expanded/resolved when generating docs, and thus confuse our end users.

bigframes/ml/utils.py
sycai: Yeah, the effort should be trivial if we are to consolidate the conversion functions from the ml package to the core package. I can handle this in another CL as what is proposed above. Let me know your thoughts.

bigframes/ml/utils.py
sycai: b/373716095 for reference

bigframes/ml/pipeline.py
TrevorBergeron: It seems that sklearn doesn't try to define a set of types in its docs: https://scikit-learn.org/1.5/modules/generated/sklearn.linear_model.LinearRegression.html#sklearn.linear_model.LinearRegression.fit ?

bigframes/ml/utils.py
```diff
@@ -55,44 +63,16 @@ def _convert_to_series(frame: ArrayType) -> bpd.Series:
         return typing.cast(bpd.Series, frame[label])
     if isinstance(frame, bpd.Series):
         return frame
+    if isinstance(frame, pd.DataFrame):
+        # Recursively call this method to re-use the length-checking logic
+        return _convert_to_series(bpd.read_pandas(frame))
```
TrevorBergeron, line 68: we might not always want the default session, if the other argument is a bigframes object with a non-default session. the core version uses the session from the co-argument

bigframes/ml/pipeline.py
sycai: Updated all to use type alias because Sphinx is able to resolve them

bigframes/ml/utils.py
sycai: Good point! I decided to use the sessions provided from the bqml_model whenever possible. The global session acts as a default.

bigframes/ml/llm.py
```diff
@@ -694,19 +694,19 @@ def _from_bq(
         model._bqml_model = core.BqmlModel(session, bq_model)
         return model
 
-    def predict(self, X: Union[bpd.DataFrame, bpd.Series]) -> bpd.DataFrame:
+    def predict(self, X: utils.ArrayType) -> bpd.DataFrame:
         """Predict the result from input DataFrame.
 
         Args:
-            X (bigframes.dataframe.DataFrame or bigframes.series.Series):
+            X (bigframes.dataframe.DataFrame or bigframes.series.Series or pandas.core.frame.DataFrame or pandas.core.series.Series):
```
TrevorBergeron, line 701: I worry that fully enumerating the accepted types will be too much once we further extend

TrevorBergeron (approved):

bigframes/ml/llm.py
sycai: I agree with you. The Google style prefers type hints over type documents, and it makes more sense. Here I'm just keeping the style consistent.

PR #1089, opened 2024-10-16
feat: support series items method

jialuoo:
See b/372343841.


bigframes/series.py
```diff
@@ -1117,6 +1117,15 @@ def ne(self, other: object) -> Series:
         # TODO: enforce stricter alignment
         return self._apply_binary_op(other, ops.ne_op)
 
+    def items(self):
+        column_ids = self._block.value_columns
+        assert (
+            len(column_ids) == 1
+        ), f"Expected lenght of column ids to be 1, but got {len(column_ids)}."
+        bpd_series = bigframes.series.Series(self._block.select_column(column_ids[0]))
+        for index, value in zip(bpd_series.index, bpd_series.values):
```
TrevorBergeron, line 1126: I would expect this to generate two separate queries, each for index and values. This is not quite ideal. Maybe take a look at dataframe.iterrows and copy that pattern.

TrevorBergeron (changes requested):

bigframes/series.py
jialuoo: Per offline discussion, the query will be based on the to_pandas_batches method in the block.

tests/system/small/test_series.py
```diff
@@ -213,6 +213,30 @@ def test_series_construct_from_list_escaped_strings():
...
+)
+def test_series_items(data, index):
+    bf_series = series.Series(data, index=index)
+    pd_series = pd.Series(data, index=index)
+
+    for (bf_index, bf_value), (pd_index, pd_value) in zip(
+        bf_series.items(), pd_series.items()
+    ):
+        # TODO(jialuo): Remove the if conditions after b/373699458 is addressed.
+        if not pd.isna(bf_index) or not pd.isna(pd_index):
```
jialuoo, line 234: We still need if clause here to handle edge cases that include types like None and numpy.nan.

TrevorBergeron (approved):

PR #1090, opened 2024-10-17
chore: debug delete table permission issue in bigframes-load-testing

shobsi:


shobsi:
The issue was identified as a potential scenario where the project contained test datasets created by other users. Since the kokoro service account did not have necesasry IAM permissions to clean up *any* dataset, it could only clean up self created dataset, and it was working so far because the project did not contain test datasets created by other users. We solved the issue by giving the kokoro service account IAM role for it to be able to delete any dataset.

PR #1097, opened 2024-10-21
feat: show possible correct key(s) in .__getitem__ KeyError message

arwas11:

Fixes #328113227 🦕


bigframes/core/groupby/__init__.py
```diff
@@ -91,8 +92,31 @@ def __getitem__(
 
         bad_keys = [key for key in keys if key not in self._block.column_labels]
 
-        if len(bad_keys) > 0:
-            raise KeyError(f"Columns not found: {str(bad_keys)[1:-1]}")
+        # Raise a KeyError message with the possible correct key(s)
+        if len(bad_keys) == 1:
```
tswast, line 96: I think you might be able to combine these two cases and avoid the if statement.

tests/system/small/test_groupby.py
```diff
@@ -464,6 +452,23 @@ def test_dataframe_groupby_getitem_list(
     pd.testing.assert_frame_equal(pd_result, bf_result, check_dtype=False)
 
 
+def test_dataframe_groupby_getitem_list_error(
```
tswast, line 455: Love it! Thanks for adding this test case.

tswast (approved):
Love it!

PR #1104, opened 2024-10-22
chore: add exec_seconds attribute to ExecutionMetrics

Genesis929:


bigframes/session/metrics.py
```diff
@@ -64,13 +66,12 @@ def get_performance_stats(
...
 
-    exec_seconds = (
-        (query_job.ended - query_job.created).total_seconds()
-        if query_job.created is not None and query_job.ended is not None
-        else None
-    )
+    if query_job.created is not None and query_job.ended is not None:
+        execution_secs = (query_job.ended - query_job.created).total_seconds()
+    else:
+        return None
```
shobsi, line 72: there are multiple `return None` statements which makes it a bit hard to follow. Can we do it only once in the beginning `if query_job.state != "DONE"`, otherwise always return a legit tuple?

Genesis929: Seems query_job.state != "DONE" does not guarantee the metrics exist for case like mocks and dry_run, so maybe this is more reliable.

shobsi (approved):

PR #1114, opened 2024-10-24
perf: Reduce CURRENT_TIMESTAMP queries

TrevorBergeron:


bigframes/session/time.py
```diff
@@ -0,0 +1,59 @@
...
+import threading
+import time
+from typing import cast, Optional
+
+import google.cloud.bigquery as bigquery
+
+MIN_RESYNC_SECONDS = 100
+
+
+class BigquerySyncedClock:
```
tswast, line 25: Nit: BigQuery (capital Q)

tswast: Also, could you make some unit tests for this functionality using [freezegun](https://github.com/spulec/freezegun) to mock out the system monotonic time? Thanks!

TrevorBergeron: added unit tests in new revision

tswast (approved):
Clever! This should really help things out.

Note: I think the benchmarks use a separate process per query if I remember correctly. If so, we might need to update those to take full advantage of this. Might even be worth some sort of "warm up" query before running the benchmarks proper.

PR #1121, opened 2024-10-29
chore: test fix.

Genesis929:


chelsea-lin (approved):

tests/system/small/test_session.py
```diff
@@ -399,7 +399,7 @@ def test_read_gbq_twice_with_same_timestamp(session, penguins_table_id):
 )
 def test_read_gbq_on_linked_dataset_warns(session, source_table):
     with warnings.catch_warnings(record=True) as warned:
-        session.read_gbq(source_table)
+        session.read_gbq(source_table, use_cache=False)
```
shobsi, line 402: Do we know why a cached read will trigger one extra warning?

Genesis929: If a table is read multiple times in the same session, there will be warning messages except for the first read. And there are two tests that involves reading this table, most likely in the same session. That I think is the cause.

PR #1132, opened 2024-11-05
fix: dataframe fillna with scalar.

Genesis929:

### DIFF
--- a/bigframes/dataframe.py
+++ b/bigframes/dataframe.py
@@ -752,7 +752,10 @@ class DataFrame(vendored_pandas_frame.DataFrame):
         )
 
     def _apply_scalar_binop(
-        self, other: float | int | bool | str, op: ops.BinaryOp, reverse: bool = False
+        self,
+        other: bigframes.dtypes.LOCAL_SCALAR_TYPE,
+        op: ops.BinaryOp,
+        reverse: bool = False,
     ) -> DataFrame:
         if reverse:
             expr = op.as_expr(
--- a/bigframes/dataframe.py
+++ b/bigframes/dataframe.py
@@ -734,7 +734,7 @@ class DataFrame(vendored_pandas_frame.DataFrame):
         how: str = "outer",
         reverse: bool = False,
     ):
-        if isinstance(other, (float, int, bool)):
+        if isinstance(other, (float, int, bool, str)):
             return self._apply_scalar_binop(other, op, reverse=reverse)
         elif isinstance(other, DataFrame):
             return self._apply_dataframe_binop(other, op, how=how, reverse=reverse)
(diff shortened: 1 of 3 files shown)


chelsea-lin (approved):
LGTM overall with a nit question:
If I understand correctly, before this Pull Request, using `fillna` with a string scalar would result in a `NotImplemented` error. If that's the case, perhaps the title should start with `feat:` instead of `fix:`?

bigframes/dataframe.py
```diff
@@ -734,7 +734,7 @@ def _apply_binop(
         how: str = "outer",
         reverse: bool = False,
     ):
-        if isinstance(other, (float, int, bool)):
```
TrevorBergeron, line 737: I feel like just adding another type here is a narrow fix - what we really want to do is determine - is this type interpretable as a supported scalar. This could include datatime objects, decimal, etc as well. We should probably have a single definition of this somewhere

TrevorBergeron (changes requested):

bigframes/dataframe.py
Genesis929: Added a LOCAL_SCALAR_TYPES constant, all types are from infer_literal_type function, should be supported.

bigframes/dataframe.py
```diff
@@ -752,7 +752,7 @@ def _apply_binop(
         )
 
     def _apply_scalar_binop(
-        self, other: float | int, op: ops.BinaryOp, reverse: bool = False
+        self, other: float | int | bool | str, op: ops.BinaryOp, reverse: bool = False
```
shobsi, line 755: SHould we use `bigframes.dtypes.LOCAL_SCALAR_TYPES` here too?

tests/system/small/test_dataframe.py
```diff
@@ -1020,13 +1020,20 @@ def test_df_interpolate(scalars_dfs):
     )
 
 
-def test_df_fillna(scalars_dfs):
+@pytest.mark.parametrize(
```
shobsi, line 1023: Let's add the customer usecase of filling "None" as a test

Genesis929: That should be the same as filling "A" I assume?

shobsi: fair enough!

bigframes/dataframe.py
Genesis929: Updated, added a new constant for annotation.

bigframes/dtypes.py
```diff
@@ -59,6 +59,25 @@
...
+    str,
+    np.str_,
+    bytes,
+    np.bytes_,
+    datetime.datetime,
+    pd.Timestamp,
+    datetime.date,
+    datetime.time,
+]
+LOCAL_SCALAR_TYPES = typing.get_args(LOCAL_SCALAR_TYPE)
```
TrevorBergeron, line 79: I have a feeling we will just give up at a certain point as we add more types

TrevorBergeron (approved):

PR #1138, opened 2024-11-07
fix: exclude index columns from model fitting processes.

Genesis929:

### DIFF
--- a/bigframes/ml/core.py
+++ b/bigframes/ml/core.py
@@ -307,11 +307,12 @@ class BqmlModelFactory:
         # Cache dataframes to make sure base table is not a snapshot
         # cached dataframe creates a full copy, never uses snapshot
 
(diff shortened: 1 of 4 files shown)


bigframes/ml/core.py
```diff
@@ -307,9 +307,11 @@ def create_model(
         # Cache dataframes to make sure base table is not a snapshot
         # cached dataframe creates a full copy, never uses snapshot
         if y_train is None:
-            input_data = X_train.cache()
+            input_data = X_train.reset_index(drop=True).cache()
```
GarrettWu, line 310: To be consistent, we may also want to drop index for other "create_model" methods. Even though those don't matter (as they only select specific column(s) to fit.

GarrettWu (approved):

PR #1139, opened 2024-11-07
feat!: Enable reading JSON data with `dbjson` extension dtype

chelsea-lin:
feat!: Enable reading JSON data with dbjson extension dtype (#1139)

This change updates how we handle JSON data types read from BigQuery.

Previously, BigQuery JSON types were treated as generic large strings within our system.  To improve accuracy and functionality, we now map them to a dedicated JSON data type (`db_dtypes.JSONType` or `db_dtypes.JSONArrowType` for pyarrow).

While this provides a more appropriate representation of JSON data, it's important to note that this feature is still in preview and may evolve.

Release-As: 1.34.0

- [X] Fixes internal issue 377764399 

Fixes internal issue 377764399 🦕


tswast (commented):
Very cool!

Let's make sure we mark this as a "breaking change" in our release notes (https://github.com/googleapis/release-please/blob/main/README.md#how-should-i-write-my-commits)

Since it's a breaking change for a preview feature, we shouldn't bump to 2.0 though. Let's use the `Release-As` footer in the commit message to make sure we do a 1.x release. https://github.com/googleapis/release-please/blob/main/README.md#how-do-i-change-the-version-number

tests/system/small/bigquery/test_json.py
```diff
@@ -203,14 +239,6 @@ def test_json_extract_string_array_w_invalid_series_type():
         bbq.json_extract_string_array(bpd.Series([1, 2]))
 
 
-# b/381148539
-def test_json_in_struct():
-    df = bpd.read_gbq(
-        "SELECT STRUCT(JSON '{\\\"a\\\": 1}' AS data, 1 AS number) as struct_col"
-    )
-    assert df["struct_col"].struct.field("data")[0] == '{"a":1}'
```
tswast, line 211: Should we keep / update this test, instead? I'd like to make sure we avoid regressions since I believe this was added to make sure we can work with some AI/ML/ObjectRef features.

tests/system/small/test_dataframe_io.py
```diff
@@ -247,23 +249,145 @@ def test_to_pandas_array_struct_correct_result(session):
...
+    pd_df = pd.DataFrame(
+        {"json_col": [{"bar": True, "foo": 10}]},
+        dtype=db_dtypes.JSONDtype(),
+    )
+    pd_df.index = pd_df.index.astype("Int64")
+    pd.testing.assert_series_equal(result.dtypes, pd_df.dtypes)
+    pd.testing.assert_series_equal(result["json_col"], pd_df["json_col"])
+
+
+def test_load_json_in_struct(session):
```
tswast, line 308: Perhaps this replaces the above test? If so, please add a docstring indicating this is a regression test for issue 381148539.

```suggestion
def test_load_json_in_struct(session):
    """Avoid regressions for internal issue 381148539."""
```

tswast (approved):

tests/system/small/bigquery/test_json.py
chelsea-lin: Yes, I moved this test to [test_dataframe_io.py](https://github.com/googleapis/python-bigquery-dataframes/pull/1139/files#diff-96483e6374e43da926825afbb4d140a7e74cc2188cf02f0a3205bb1c2d3e99cf). Also add similar tests for both `struct` and `array`

PR #1142, opened 2024-11-11
docs: add snippet for creating boosted tree model

rey-esp:


samples/snippets/classification_boosted_tree_model_test.py
```diff
@@ -39,4 +39,27 @@ def test_boosted_tree_model(random_model_id: str) -> None:
     )
     del input_data["functional_weight"]
     # [END bigquery_dataframes_bqml_boosted_tree_prepare]
+    # [START bigquery_dataframes_bqml_boosted_tree_create]
+    # from sklearn.ensemble import GradientBoostingClassifier
+    import bigframes.ml.linear_model
+
+    # input_data is defined in an earlier step.
+    training_data = input_data[input_data["dataframe"] == "training"]
```
tswast, line 47: No action needed, but something to consider for future: it would be nice to update the `prepare` section above to work without referencing an index (e.g. when ordering mode = "partial").

We have a few options, but the easiest will be to start with a string column and add `(True, "training")` as the last in the list of cases.

Aside: we have an issue open (349926559) to allow selecting _any_ column in the dataframe (such as `functional_weight`, which would be a natural choice in this example) even if its a different type, so long as a `True` (default) case is provided.

samples/snippets/classification_boosted_tree_model_test.py
```diff
@@ -39,4 +39,27 @@ def test_boosted_tree_model(random_model_id: str) -> None:
...
+    X = training_data.drop(columns=["income_bracket", "dataframe"])
+    y = training_data["income_bracket"]
```
tswast, line 49: Presumably you ran this code sample and it worked OK? I remember we had some bugs where `y` had to be a DataFrame not a Series in past, so just double-checking.

samples/snippets/classification_boosted_tree_model_test.py
```diff
@@ -39,4 +39,27 @@ def test_boosted_tree_model(random_model_id: str) -> None:
...
+
+    # create and train the model
+    census_model = bigframes.ml.linear_model.LogisticRegression(
+        # model_type="BOOSTED_TREE_CLASSIFIER",
+        # booster_type="gbtree",
+        max_iterations=50,
+    )
```
tswast, line 56: I don't think we should be doing `LogisticRegression` here. In the SQL we do use `model_type='BOOSTED_TREE_CLASSIFIER'`, but in BigQuery DataFrames we normally use separate Python classes to represent the different model types.

A few ways to discover which class to use:

* Search our [code for `BOOSTED_TREE_CLASSIFIER`](https://github.com/search?q=repo%3Agoogleapis%2Fpython-bigquery-dataframes%20BOOSTED_TREE_CLASSIFIER&type=code)
* Google search for [boosted trees BigFrames](https://www.google.com/search?q=boosted+trees+bigframes)

These should give you some strong hints as to which class to use instead.

tswast (changes requested):

samples/snippets/classification_boosted_tree_model_test.py
rey-esp: The code sample seems to run! Not sure if I did it right so here's the colab: https://colab.sandbox.google.com/drive/10jA6zSRiptXWrTkCcmyCT_sYBjLqGJx0?resourcekey=0-0TrIkmDzAJw_F6ONFikwaA#scrollTo=wU367u1SAj3Y

samples/snippets/classification_boosted_tree_model_test.py
tswast: Copying this from an internal comment I made for visibility:

Just like scikit-learn, it's one of the "ensemble" methods: [https://cloud.google.com/python/docs/reference/bigframes/latest/bigframes.ml.ensemble](https://www.google.com/url?q=https://cloud.google.com/python/docs/reference/bigframes/latest/bigframes.ml.ensemble&sa=D&source=docs&ust=1731968708166884&usg=AOvVaw2y30ZNkl2x1T5ip52wSzNr)

Normally we try to use the scikit-learn class names too, but I think we may have added this class before GradientBoostingClassifier as in scikit-learn

tswast (approved):

PR #1148, opened 2024-11-12
fix: unordered mode too many labels issue.

Genesis929:

### DIFF
--- a/bigframes/session/executor.py
+++ b/bigframes/session/executor.py
@@ -379,11 +379,12 @@ class BigQueryCachingExecutor:
             job_config.maximum_bytes_billed = (
                 bigframes.options.compute.maximum_bytes_billed
             )
-        # Note: add_labels is global scope which may have unexpected effects
-        bq_io.add_labels(job_config, api_name=api_name)
 
         if not self.strictly_ordered:
             job_config.labels["bigframes-mode"] = "unordered"
+
+        # Note: add_labels is global scope which may have unexpected effects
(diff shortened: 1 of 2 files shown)


bigframes/session/executor.py
```diff
@@ -379,11 +379,12 @@ def _run_execute_query(
...
                 bigframes.options.compute.maximum_bytes_billed
             )
-        # Note: add_labels is global scope which may have unexpected effects
-        bq_io.add_labels(job_config, api_name=api_name)
 
         if not self.strictly_ordered:
             job_config.labels["bigframes-mode"] = "unordered"
+
+        # Note: add_labels is global scope which may have unexpected effects
+        bq_io.add_labels(job_config, api_name=api_name)
```
shobsi, line 387: Highly recommend adding a test using the customer's use case

Genesis929: Added

tests/system/small/test_series.py
```diff
@@ -2766,22 +2766,17 @@ def test_series_case_when(scalars_dfs_maybe_ordered):
     bf_series = scalars_df["int64_col"]
     pd_series = scalars_pandas_df["int64_col"]
 
-    # TODO(tswast): pandas case_when appears to assume True when a value is
```
shobsi, line 2769: I would keep this comment

Genesis929: Updated to have all cases in the original test. Keep only one test can be overall faster I think.

shobsi (approved):

PR #1152, opened 2024-11-14
feat: (Series | DataFrame).plot.bar

chelsea-lin:
Fixes internal issue 375905303🦕

### DIFF
--- a/bigframes/operations/_matplotlib/core.py
+++ b/bigframes/operations/_matplotlib/core.py
@@ -52,6 +52,10 @@ class SamplingPlot(MPLPlot):
     def _kind(self):
         pass
 
+    @property
+    def _sampling_warning_msg(self) -> typing.Optional[str]:
+        return None
+
     def __init__(self, data, **kwargs) -> None:
         self.kwargs = kwargs
         self.data = data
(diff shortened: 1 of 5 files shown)


bigframes/operations/_matplotlib/core.py
```diff
@@ -74,18 +75,24 @@ def _compute_plot_data(self):
...
 class AreaPlot(SamplingPlot):
     @property
     def _kind(self) -> typing.Literal["area"]:
         return "area"
 
 
+class BarPlot(SamplingPlot):
+    @property
+    def _kind(self) -> typing.Literal["bar"]:
+        return "bar"
```
TrevorBergeron, line 87: Does bar plot make sense as a sampling plot? It seems that bar plots are really meant for small-cardinalities and don't down-sample well.

chelsea-lin: The bar plots are discrete while the area/line/scatter plots are continuous. Hence, the downsampling in the bar is not straightforward compared to others. Thanks for suggestions here. I just added a warning message to suggest users to consider pre-processing data (aggregations or select top categories).

TrevorBergeron (approved):

PR #1153, opened 2024-11-14
fix: Fix null index join with 'on' arg

TrevorBergeron:

### DIFF
--- a/tests/system/small/test_null_index.py
+++ b/tests/system/small/test_null_index.py
@@ -341,6 +341,25 @@ def test_null_index_df_concat(scalars_df_null_index, scalars_pandas_df_default_i
     )
 
 
+def test_null_index_map_dict_input(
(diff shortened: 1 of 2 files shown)


tests/system/small/test_null_index.py
```diff
@@ -268,6 +268,25 @@ def test_null_index_series_self_join(
     )
 
 
+def test_null_index_series_self_join_on(
```
shobsi, line 271: were we able to write a sample code that reproduces the issue in b/359602477?  It would be great to add that as a test case `.map()`

TrevorBergeron: added test in new revision

shobsi (approved):

tests/system/small/test_null_index.py
shobsi: Thank you!

PR #1154, opened 2024-11-14
docs: add snippet for evaluating a boosted tree model

rey-esp:

### DIFF
--- a/samples/snippets/classification_boosted_tree_model_test.py
+++ b/samples/snippets/classification_boosted_tree_model_test.py
@@ -38,7 +38,8 @@ def test_boosted_tree_model(random_model_id: str) -> None:
         ]
     )
     del input_data["functional_weight"]
-    # [END bigquery_dataframes_bqml_boosted_tree_prepare]   # [START bigquery_dataframes_bqml_boosted_tree_explain]
+    # [END bigquery_dataframes_bqml_boosted_tree_prepare]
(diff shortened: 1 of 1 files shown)


samples/snippets/classification_boosted_tree_model_test.py
```diff
@@ -62,5 +62,29 @@ def test_boosted_tree_model(random_model_id: str) -> None:
         replace=True,
     )
     # [END bigquery_dataframes_bqml_boosted_tree_create]
+    # [START bigquery_dataframes_bqml_boosted_tree_explain]
+    # Select model you'll use for predictions. `read_gbq_model` loads model
+    # data from BigQuery, but you could also use the `tree_model` object
+    # from the previous step.
+    tree_model = bpd.read_gbq_model(
+        your_model_id,  # For example: "your-project.bqml_tutorial.tree_model"
```
tswast, line 70: I see this is different from `to_gbq` above. Let's make sure the model ID here and there matches the one from the SQL sample.

tswast (commented):
Looks good! Let's just fix up the read_gbq and to_gbq comments

samples/snippets/classification_boosted_tree_model_test.py
rey-esp: Nice catch! I believe I added in the correct names for everything, but I'm not 100% certain

tswast (approved):
Awesome, thanks!

PR #1166, opened 2024-11-21
feat: support dataframe where method

jialuoo:
See details at https://buganizer.corp.google.com/issues/369895394#comment7


bigframes/dataframe.py
```diff
@@ -2039,6 +2039,55 @@ def itertuples(
...
+            other_id = other_col[x] if x in other_col else ex.const(None)
+            result_block, result_id = aligned_block.project_expr(
+                ops.where_op.as_expr(self_id, cond_id, other_id)
+            )
+            series = bigframes.series.Series(
+                result_block.select_column(result_id).with_column_labels([x])
+            )
+            result_series[x] = series
+
+        return DataFrame(result_series)
```
TrevorBergeron, line 2089: I think the only issue with this is we are rebuilding the columns index from the individual labels, so we may actually lose `df.columns.name` property from the original dataframe. Might be worth setting this property in one of the tests to see if we match pandas behavior

jialuoo: Nice catch! 
1. Fixed the `df.column.name` and `df.column.names` properties (and added a test for it) to match pandas behavior.
2. Per offline discussion, it will not support multi-index and/or multi-column for now (see details from https://buganizer.corp.google.com/issues/369895394#comment8)

bigframes/dataframe.py
```diff
@@ -2039,6 +2039,63 @@ def itertuples(
             for item in df.itertuples(index=index, name=name):
                 yield item
 
+    def where(self, cond, other=None):
+        if isinstance(other, bigframes.series.Series):
+            raise ValueError("Seires is not a supported replacement type!")
+
+        if self.columns.nlevels > 1 or self.index.nlevels > 1:
```
TrevorBergeron, line 2046: I understand `columns` multindex might cause some trouble, but what issues does the `index` mutliindex cause? Does default join alignment logic not work for some cases?

jialuoo: Very good question. I considered it and lean not to support it in this PR. This is because multi-index also includes some rare edge cases (similar to multi-columns) that need further investigation and analysis.

For pandas df.where(), it generally supports 2 major use cases:
-- Case1 (most commonly seen): when the condition/`cond` is a logical operation of the df itself, e.g. df.where(df>0), df.where(df[‘col’]>0).
-- Case2 (edge cases): when condition/`cond` is another user-defined dataframe (say df1), e.g. df.where(df1>0), df.where(df1[‘col’]>0), etc.
(please note, for replacement/`other`, it also supports similar use cases with some difference: e.g. it cannot be a Series)

The algorithm in this PR supports both use cases (same behavior as pandas, verified) when there is no multi-index and/or multi-columns involved. It actually supports Case1 for multi-index (verified too).

However, when supporting multi-index, Case2 are not well documented from pandas and the execution rules are ambiguous (e.g. the join step between df and df1 within the where method).

I suggest opening another task with more investigation and analysis for edge cases if we want to support it later.

TrevorBergeron (approved):

PR #1168, opened 2024-11-22
feat: Allow join-free alignment of analytic expressions

TrevorBergeron:


bigframes/core/node_align.py
```diff
@@ -0,0 +1,167 @@
...
+@dataclasses.dataclass(frozen=True)
+class ExpressionSpec:
+    expression: bigframes.core.expression.Expression
+    node: bigframes.core.nodes.BigFrameNode
+
+
+def get_expression_spec(
+    node: bigframes.core.nodes.BigFrameNode, id: bigframes.core.identifiers.ColumnId
+) -> ExpressionSpec:
+    """Normalizes column value by chaining expressions across multiple selection and projection nodes if possible"""
```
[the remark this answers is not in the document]
tswast: Also, could you describe a bit more about the purpose? I'm slightly confused as to why we'd need to do this normalization one column at a time? Maybe for Series?

bigframes/core/node_align.py
```diff
@@ -0,0 +1,167 @@
...
+        curr_node = curr_node.child
+
+
+def _linearize_trees(
+    base_tree: bigframes.core.nodes.BigFrameNode,
+    append_tree: bigframes.core.nodes.BigFrameNode,
+) -> bigframes.core.nodes.BigFrameNode:
+    """Linearize two divergent tree who only diverge through different additive nodes."""
+    assert append_tree.projection_base == base_tree.projection_base
+    if append_tree == append_tree.projection_base:
```
tswast, line 80: When would it be its own projection base? Could you give some examples in a comment?

bigframes/core/node_align.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def join_as_projection(
+    l_node: bigframes.core.nodes.BigFrameNode,
+    r_node: bigframes.core.nodes.BigFrameNode,
+    join_keys: Tuple[Tuple[str, str], ...],
+) -> Optional[bigframes.core.nodes.BigFrameNode]:
+    """Joins the two nodes"""
+    if l_node.projection_base != r_node.projection_base:
+        return None
```
tswast, line 108: Since there are `return None`s should the be `try_join_as_projection`?

bigframes/core/node_align.py
```diff
@@ -0,0 +1,167 @@
...
+    # check join key
+    for l_key, r_key in join_keys:
+        # Caller is block, so they still work with raw strings rather than ids
```
tswast, line 111: Should we fix that? e.g. make block use ColumnId?

bigframes/core/node_align.py
```diff
@@ -0,0 +1,167 @@
...
+
+import bigframes.core.expression
+import bigframes.core.guid
+import bigframes.core.identifiers
+import bigframes.core.join_def
+import bigframes.core.nodes
+import bigframes.core.window_spec
+import bigframes.operations.aggregations
+
+ADDITIVE_NODES = (
```
tswast, line 27: Would be helpful to have some comments about what these node types have in common. How would I decide if a node type should be added to this list?

bigframes/core/node_align.py
```diff
@@ -0,0 +1,167 @@
...
+        right_id = bigframes.core.identifiers.ColumnId(r_key)
+        if get_expression_spec(l_node, left_id) != get_expression_spec(
+            r_node, right_id
+        ):
+            return None
+    return combine_nodes(l_node, r_node)
+
+
+def pull_up_selection(
+    node: bigframes.core.nodes.BigFrameNode, rename_vars: bool = False
```
tswast, line 122: What's the purpose of the `rename_vars` parameter? A docstring would be helpful.

bigframes/core/node_align.py
```diff
@@ -0,0 +1,167 @@
...
+        ...,
+    ],
+]:
+    """Remove all selection nodes above the base node. Returns stripped tree."""
+    if node == node.projection_base:  # base case
+        return node, tuple(
+            (bigframes.core.expression.DerefOp(field.id), field.id)
+            for field in node.fields
+        )
+    assert isinstance(node, (bigframes.core.nodes.SelectionNode, *ADDITIVE_NODES))
```
tswast, line 136: Am I remembering correctly that `SelectionNode` represents a WHERE clause, based on [Selection from relational algebra terminology](https://en.wikipedia.org/wiki/Selection_(relational_algebra))?

Given the high-level that implicit joiner no longer handles these, I'm a little confused why we need to rewrite these? Is it so that they are all consolidated so that the nodes we can combine appear together in the tree? I would appreciate a bit more info.

TrevorBergeron: Its actually poorly named. SelectionNode just renames/reorders existing columns, without any scalar transforms

bigframes/core/node_align.py
TrevorBergeron: added docstring. its intended to make sure that when we combine columns from two sides, we don't get conflicts.

bigframes/core/node_align.py
TrevorBergeron: Yes, we should do that refactor, but its a bit of an undertaking. For now, block uses string ids, and ArrayValue is responsible for wrapping them up as ColumnIds.

bigframes/core/blocks.py
```diff
@@ -2341,7 +2341,9 @@ def join(
         # Handle null index, which only supports row join
         # This is the canonical way of aligning on null index, so always allow (ignore block_identity_join)
         if self.index.nlevels == other.index.nlevels == 0:
-            result = try_row_join(self, other, how=how)
+            result = try_legacy_row_join(self, other, how=how) or try_new_row_join(
```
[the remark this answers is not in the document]
TrevorBergeron: Not needed now, as now emulating more recent pandas alignment logic, where the sides need to match exactly, without any view-like interpretations of filtered objects.

bigframes/core/nodes.py
```diff
@@ -307,6 +311,14 @@ def transform_children(
             return self
         return transformed
 
+    def replace_child(
+        self, new_child: BigFrameNode, validate: bool = False
+    ) -> UnaryNode:
+        new_self = replace(self, child=new_child)  # type: ignore
+        if validate:
```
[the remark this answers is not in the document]
TrevorBergeron: eh, I think I"ll remove that, its not being used, and yeah, could run validations it as a separate invocation

bigframes/core/__init__.py
```diff
@@ -488,11 +514,4 @@ def _gen_namespaced_uid(self) -> str:
...
     def _gen_namespaced_uids(self, n: int) -> List[str]:
-        i = len(self.node.defined_variables)
-        genned_ids: List[str] = []
-        while len(genned_ids) < n:
-            attempted_id = f"col_{i}"
-            if attempted_id not in self.node.defined_variables:
-                genned_ids.append(attempted_id)
-            i = i + 1
-        return genned_ids
+        return [ids.ColumnId.unique().name for _ in range(n)]
```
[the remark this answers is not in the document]
TrevorBergeron: Not any less deterministic than we are now. If we want an isomorphism between query structure and output syntax, there is a bit more work that needs to be done to the system, which I think basically amounts to late binding identifiers serially through the tree.

bigframes/core/node_align.py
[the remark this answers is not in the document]
TrevorBergeron: Basically, we are trying to identify if the join keys are equal, so we want to compare the join key expression without evaluating. The same value can be achieve in different ways with different names, so the goal here is to normalize the key expression as much as possible to evaluate equivalence.

bigframes/core/node_align.py
TrevorBergeron: added comment explaining

bigframes/core/__init__.py
```diff
@@ -449,14 +449,40 @@ def relational_join(
         )
         return ArrayValue(join_node), (l_mapping, r_mapping)
 
-    def try_align_as_projection(
+    def try_new_row_join(
```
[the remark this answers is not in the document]
TrevorBergeron: done

bigframes/core/nodes.py
```diff
@@ -307,6 +311,14 @@ def transform_children(
             return self
         return transformed
 
+    def replace_child(
+        self, new_child: BigFrameNode, validate: bool = False
+    ) -> UnaryNode:
+        new_self = replace(self, child=new_child)  # type: ignore
```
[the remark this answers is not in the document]
TrevorBergeron: fixed this import

tswast (approved):

PR #1174, opened 2024-11-25
fix: fix read_gbq_function issue in dataframe apply method

jialuoo:
See the description/example from b/379750234.


bigframes/dataframe.py
```diff
@@ -3674,7 +3674,11 @@ def apply(self, func, *, axis=0, args: typing.Tuple = (), **kwargs):
             return result_series
 
         # Per-column apply
-        results = {name: func(col, *args, **kwargs) for name, col in self.items()}
+        if hasattr(func, "bigframes_remote_function"):
```
shobsi, line 3677: Let's add a test please, we could use the same example you used to report the issue.

bigframes/dataframe.py
```diff
@@ -3674,7 +3674,11 @@ def apply(self, func, *, axis=0, args: typing.Tuple = (), **kwargs):
...
+            results = {name: col.apply(func) for name, col in self.items()}
+        else:
+            results = {name: func(col, *args, **kwargs) for name, col in self.items()}
```
shobsi, line 3680: I believe this already has test coverage, let's double check

bigframes/dataframe.py
jialuoo: Added.

bigframes/dataframe.py
jialuoo: Yeah, we have several tests for it already. Confirmed!

tests/system/small/test_dataframe.py
```diff
@@ -1161,6 +1161,23 @@ def test_apply_series_scalar_callable(
     pandas.testing.assert_series_equal(bf_result, pd_result)
 
 
+def test_apply_from_read_gbq_function(dataset_id_permanent):
```
shobsi, line 1164: This should go in tests/system/small/test_remote_function.py

tests/system/small/test_dataframe.py
```diff
@@ -1161,6 +1161,23 @@ def test_apply_series_scalar_callable(
...
+    @bpd.remote_function()
```
shobsi, line 1165: We don't need to deploy a new remote function (it is costly, takes around 1 minute). We can write the test using existing functions screen/8yHKQ5KibqDMLvL

bigframes/dataframe.py
```diff
@@ -3674,7 +3674,11 @@ def apply(self, func, *, axis=0, args: typing.Tuple = (), **kwargs):
             return result_series
 
         # Per-column apply
-        results = {name: func(col, *args, **kwargs) for name, col in self.items()}
+        if hasattr(func, "bigframes_remote_function"):
+            results = {name: col.apply(func) for name, col in self.items()}
```
shobsi, line 3678: We should show a warning that args and kwargs are not supported in remote functions

bigframes/dataframe.py
shobsi, line 3677: @TrevorBergeron could you please review if this element-wise proxy for series-wise application doesn't have any blind sight?

tests/system/small/test_dataframe.py
jialuoo: Done (Created a test in test_remote_function.py and removed the old one).

tests/system/small/test_dataframe.py
jialuoo: Makes sense. Used an existing function instead.

bigframes/dataframe.py
jialuoo: Warning added.

bigframes/dataframe.py
TrevorBergeron: Isn't `dataframe.apply` for column or series-wise application? Why do we need to support applying as an element-wise operator? isn't this what `dataframe.map` is for?

tests/system/small/test_remote_function.py
```diff
@@ -839,6 +839,30 @@ def test_read_gbq_function_enforces_explicit_types(
...
+    bdf = bigframes.pandas.DataFrame(
+        {
+            "Column1": scalars_df["string_col"],
+            "Column2": scalars_df["string_col"],
+        }
+    )
+
+    func_ref = session.read_gbq_function("bqutil.fn.cw_lower_case_ascii_only")
+    bf_result = bdf.apply(func_ref).to_pandas()
+    # The str.lower() method has the same functionality as func_ref.
```
shobsi, line 854: This is true for ASCII only, we should say that in the comment

bigframes/dataframe.py
shobsi: For scalar ops the end result is the same. I think we don't lose anything by being lax for user convenience. Note that this API has other non-pandas-like extended behavior https://github.com/googleapis/python-bigquery-dataframes/blob/e0a82888cd34fa2404ac68229dc38496cb22c67b/bigframes/dataframe.py#L3933

jialuoo: After offline discussion, we'll direct users to use .map() method for element-wise application.

tests/system/small/test_remote_function.py
jialuoo: This test is not needed anymore.

bigframes/exceptions.py
```diff
@@ -81,3 +81,7 @@ class ApiDeprecationWarning(FutureWarning):
 
 class BadIndexerKeyWarning(Warning):
     """The indexer key is not used correctly."""
+
+
+class ArgsAndKwargsNotSupportedWarning(Warning):
```
shobsi, line 86: This is unnecessary, as now we are not supporting remote function at all in dataframe.apply

jialuoo: Removed.

bigframes/dataframe.py
```diff
@@ -3958,8 +3958,26 @@ def apply(self, func, *, axis=0, args: typing.Tuple = (), **kwargs):
             result_series.name = None
             return result_series
 
+        # This is when the func as a remote function is applied to each element of
```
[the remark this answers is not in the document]
jialuoo: Good point. Fixed.

bigframes/dataframe.py
```diff
@@ -4016,8 +4020,26 @@ def apply(self, func, *, axis=0, args: typing.Tuple = (), **kwargs):
...
+        # the dataframe (not supported).
+        if hasattr(func, "is_row_processor") and not func.is_row_processor:
```
shobsi, line 4025: We don't need this check anymore. Just the check on line 4032 is enough.

Just to clarify `is_row_processor` is set to indicate row-wise operation, not element-wise operation

bigframes/dataframe.py
```diff
@@ -4031,8 +4035,20 @@ def apply(self, func, *, axis=0, args: typing.Tuple = (), **kwargs):
...
 
+        # At this point column-wise or element-wise remote function operation will
+        # be performed (not supported).
+        if hasattr(func, "bigframes_remote_function"):
+            raise NotImplementedError(
+                "DataFrame '.apply()' does not support remote function for "
+                "element-wise or column-wise (i.e. with axis=0) application. "
+                "Please use '.map()' instead for element-wise application of the "
+                "remote function, or use regular python function for column-wise "
+                "application."
```
shobsi, line 4046: Hopefully this would be my comment :). elemrnt-wise should not even be a clarification for a pandas user, as DataFrame.apply only takes in row-wise or colum-wise functions. I would suggest rephrasing the error to the following:

"""
'.apply()' does not support remote function for column-wise (i.e. with axis=0) operations,
please use a regular python function instead.
For element-wise operation of the remote function, please use '.map()'.
"""

jialuoo: Done. No rush, take as long as we might need :)

shobsi (approved):

PR #1182, opened 2024-11-27
chore: update comments and tests for label counts.

Genesis929:
1. Renamed add_labels to add_and_trim_labels.
2. Added notes and docstring for future notification.
3. Added tests for add_and_trim_labels and the functions that calls add_and_trim_labels to ensure the total number of labels won't exceed 64.


bigframes/session/executor.py
```diff
@@ -348,7 +348,10 @@ def export_gcs(
             export_options=dict(export_options),
         )
         job_config = bigquery.QueryJobConfig()
-        bq_io.add_labels(job_config, api_name=f"dataframe-to_{format.lower()}")
+
+        # Note: Ensure no additional labels are added to job_config after this point,
+        # as `add_and_trim_labels` ensures the label count does not exceed 64.
```
shobsi, line 353: The repetition of this note at multiple places makes me think - can we not call `start_query_with_client` instead of `bqclient.query` from everywhere? That way any preprocessing (including labels) would be centralized and have only one note lilke this in our code.

Genesis929: The logic after bqclient.query in start_query_with_client don't seems to match other functions, so this may not work.

We can make a new function that add labels and execute to replace bqclient.query, what do you think? This is my original thought but I'm not sure if it's necessary to add a new function just for two lines of code.

shobsi: Not a great refactoring but better than leaving repetitive notes IMHO.

shobsi: Or we could make a hook in the `ClientsProvider._create_bigquery_client`, something like:

```python
original_query = bq_client.query

def query (job_config, *args, **kwargs, api_name):
    add_and_trim_labels(job_config, api_name=api_name)
    original_query(job_config, *args, **kwargs)
```

tswast: I would discourage us from monkey patching like this, though it should work in Python. I think your first instinct of using our central helper function makes the most sense to me.

I don't fully understand what you mean by

> The logic after bqclient.query in start_query_with_client don't seems to match other functions, so this may not work.

We should be able to put `start_query_with_client` in a try/except block. Also, it's probably a bug that we aren't showing that a query job is running here.

bigframes/session/executor.py
```diff
@@ -358,7 +361,9 @@ def dry_run(
     ) -> bigquery.QueryJob:
         sql = self.to_sql(array_value, ordered=ordered)
         job_config = bigquery.QueryJobConfig(dry_run=True)
-        bq_io.add_labels(job_config)
+        # Note: Ensure no additional labels are added to job_config after this point,
+        # as `add_and_trim_labels` ensures the label count does not exceed 64.
+        bq_io.add_and_trim_labels(job_config)
         query_job = self.bqclient.query(sql, job_config=job_config)
         _ = query_job.result()
```
tswast, line 368: I'm pretty sure `result()` is a no-op for dry-run.

bigframes/session/executor.py
```diff
@@ -358,7 +361,9 @@ def dry_run(
     ) -> bigquery.QueryJob:
         sql = self.to_sql(array_value, ordered=ordered)
         job_config = bigquery.QueryJobConfig(dry_run=True)
-        bq_io.add_labels(job_config)
+        # Note: Ensure no additional labels are added to job_config after this point,
+        # as `add_and_trim_labels` ensures the label count does not exceed 64.
+        bq_io.add_and_trim_labels(job_config)
```
tswast, line 366: I don't think we actually create a job resource in the backend when we do a dry-run, so this might actually lose analytics. I think probably better not to call `add_and_trim_labels` at all in this case.

bigframes/session/executor.py
Genesis929: Updated

bigframes/session/_io/bigquery/__init__.py
```diff
@@ -204,7 +204,12 @@ def format_option(key: str, value: Union[bool, str]) -> str:
     return f"{key}={repr(value)}"
 
 
-def add_labels(job_config, api_name: Optional[str] = None):
+def add_and_trim_labels(job_config, api_name: Optional[str] = None):
+    """
+    Add additional labels to the job configuration and trim the total number of labels
+    to ensure they do not exceed the maximum limit allowed by BigQuery, which is 64
```
tswast, line 210: FYI: I added a buffer to this, so the limit is slightly lower. See internal discussion we had in team meeting yesterday.

tests/unit/session/test_io_bigquery.py
```diff
@@ -148,6 +166,82 @@ def test_create_job_configs_labels_length_limit_met():
...
+    job_config = bigquery.job.QueryJobConfig()
+    job_config.labels = cur_labels
+
+    df.max()
+    for _ in range(60):
+        df.head()
+
+    io_bq.add_and_trim_labels(job_config=job_config)
+    assert job_config.labels is not None
+    assert len(job_config.labels) == 64
```
tswast, line 193: This should be failing now. The correct limit it 64 - 8 to give a buffer for internal labels. Let's use the max labels count constant here.

https://github.com/googleapis/python-bigquery-dataframes/blob/b86335874c288a16e67017829172e76819e4876e/bigframes/core/log_adapter.py#L24

tests/unit/session/test_io_bigquery.py
```diff
@@ -148,6 +166,82 @@ def test_create_job_configs_labels_length_limit_met():
...
+        mock_bq_client,
+        sql,
+        job_config,
+        max_results=max_results,
+        timeout=timeout,
+        api_name=api_name,
+    )
+
+    assert job_config.labels is not None
+    assert len(job_config.labels) == 64
```
tswast, line 238: Likewise, this limit should be lower.

https://github.com/googleapis/python-bigquery-dataframes/blob/b86335874c288a16e67017829172e76819e4876e/bigframes/core/log_adapter.py#L24

Genesis929: Thanks! In that case there is an additional value that need to be updated in bigframes/session/_io/bigquery/__init__.py, included now.

bigframes/session/_io/bigquery/__init__.py
Genesis929: Thanks! Good to know!

tswast (approved):

PR #1183, opened 2024-11-27
refactor: Define RowJoinNode and defer rewrite

TrevorBergeron:


bigframes/core/blocks.py
```diff
@@ -2335,8 +2335,8 @@ def join(
         # Handle null index, which only supports row join
         # This is the canonical way of aligning on null index, so always allow (ignore block_identity_join)
         if self.index.nlevels == other.index.nlevels == 0:
-            result = try_legacy_row_join(self, other, how=how) or try_new_row_join(
-                self, other
+            result = try_new_row_join(self, other) or try_legacy_row_join(
```
sycai, line 2338: We should probably rename "try_new_row_join" now or in the future. The reason is that I guess "try_legacy_row_join" will be eventually removed, and it would be very confusing if we have only a "new" version.

We can just call it "try_row_join".

bigframes/core/rewrite/implicit_align.py
```diff
@@ -131,9 +89,35 @@ def _linearize_trees(
...
+    """If root and child are projection nodes, merge them."""
+    if not isinstance(root, bigframes.core.nodes.ProjectionNode):
+        return root
+    if not isinstance(root.child, bigframes.core.nodes.ProjectionNode):
+        return root
+    mapping = {id: expr for expr, id in root.child.assignments}
+    new_exprs = (
+        *root.child.assignments,
+        *(
+            (expr.bind_refs(mapping, allow_partial_bindings=True), id)
```
sycai, line 114: This tuple comprehension is very long. Shall we use another local variable "root_assignments" to hold the value?

And 
"new_exprs = tuple(root.child.assignments) + root_assignments"

bigframes/core/rewrite/implicit_align.py
```diff
@@ -201,26 +185,24 @@ def pull_up_selection(
...
+    if len(roots) == 1:
+        return next(iter(roots))
 
-    for l_node, r_node in zip(l_path[-len(r_path) :], r_path[-len(l_path) :]):
-        if l_node == r_node:
-            return l_node
-    # should be impossible, as l_path[-1] == r_path[-1]
-    raise ValueError()
+    min_height = min(root.height for root in roots)
+    to_descend = set(root for root in roots if root.height > min_height)
```
sycai, line 197: It took me a while to realize that `to_descend` is a set of root notes.

Let's name it "descend_roots" ?

sycai (dismissed):

bigframes/core/rewrite/implicit_align.py
TrevorBergeron: renamed to `roots_to_descend`

bigframes/core/blocks.py
TrevorBergeron: renamed to `try_row_join`

bigframes/core/rewrite/implicit_align.py
TrevorBergeron: extracted out to variable as suggested

PR #371, opened 2024-02-06
feat: add `write_engine` parameter to `read_FORMATNAME` methods to control how data is written to BigQuery

tswast:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery-dataframes/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes internal issue 323176126
🦕

PR #1187, opened 2024-12-04
fix: json in struct destination type

GarrettWu:
Patches the json output as BQ JSON type instead of STRING. Which avoids JSON destination table error in b/381148539 and unblocks Multimodal integrations.

Our current implementation convert JSON to STRING completely when reading into BigFrames. Here it uses pd.ArrowDtype(pa.large_string()) as BigFrames Dtype and pa.large_string() as PA dtype, and pass the information though in BigFrames to determine the output type and parse STR back to JSON at the end. 

It is a workaround of the workaround of JSON as STR. We shall rework the JSON support after pyarrow brings in json type.


tswast:
Question: Does this change anything with regards to I/O? For example, I notice with `read_csv` with the pandas-y code path, the DataFrame from pandas puts string columns as the `large_string()` in my testing. Hopefully such columns don't end up as JSON in BigQuery now.

GarrettWu:
> Question: Does this change anything with regards to I/O? For example, I notice with `read_csv` with the pandas-y code path, the DataFrame from pandas puts string columns as the `large_string()` in my testing. Hopefully such columns don't end up as JSON in BigQuery now.

How did you test? I tried pandas df with pa.large_string dtype is interpreted as normal string. It shouldn't affect input as large_string, since we convert it to normal string in our I/O. 

screen/763yZzyywmEVWbR

tswast:
> How did you test?

It was part of my debugger step-through of https://github.com/googleapis/python-bigquery-dataframes/pull/371 but it's possible I accidentally changed something in the I/O path there.

TrevorBergeron (approved):

bigframes/core/__init__.py
```diff
@@ -107,7 +107,7 @@ def from_table(
             raise ValueError("must set at most one of 'offests', 'primary_key'")
         if any(i.field_type == "JSON" for i in table.schema if i.name in schema.names):
             warnings.warn(
-                "Interpreting JSON column(s) as StringDtype. This behavior may change in future versions.",
+                "Interpreting JSON column(s) as StringDtype and pyarrow.large_string. This behavior may change in future versions.",
```
shobsi, line 110: Shouldn't we just say "... as pyarrow.large_string"? StringDtype here is not relevant anymore, no?

bigframes/core/blocks.py
```diff
@@ -2564,13 +2564,13 @@ def _get_rows_as_json_values(self) -> Block:
 ),
 T1 AS (
     SELECT *,
-           JSON_OBJECT(
+           TO_JSON_STRING(JSON_OBJECT(
```
shobsi, line 2567: Do we need an explicit TO_JSON_STRING? Doesn't the change in ibis_types.py::ibis_dtype_to_bigframes_dtype take care of this post read?

```
    if isinstance(ibis_dtype, ibis_dtypes.JSON):
        ...
        return bigframes.dtypes.JSON_DTYPE
```

PR #1198, opened 2024-12-09
refactor: Create memoized tree traversal helpers

TrevorBergeron:


bigframes/core/nodes.py
```diff
@@ -1486,3 +1486,56 @@ def remap_refs(
...
+    validate=False,
+) -> BigFrameNode:
+    """
+    Perform a bottom-up transformation of the BigFrameNode tree.
+
+    If memoize=True, recursive calls are memoized within the scope of the traversal only.
+    """
+
+    def bottom_up_internal(root: BigFrameNode) -> BigFrameNode:
+        return transform(root.transform_children(bottom_up_internal))
```
GarrettWu, line 1532: Don't see a difference with top_down. Should bottom_up apply to parents?

TrevorBergeron: top down applies `transform_children()` after `transform(root)` that is the only difference.

GarrettWu:
```
def bottom_up_internal(root: BigFrameNode) -> BigFrameNode:
        return transform(root.transform_children(bottom_up_internal))
```
Then this is a mistake?

TrevorBergeron: nope, not a mistake. bottom_up transforms the children first. and then transforms the root last.

GarrettWu: Then bottom up doesn't apply to their parents? How is that "bottom up"?

TrevorBergeron: to which parents? its a classic post-order depth-first traversal, all nodes under the root are transformed, and then the root is the final node to be transformed

GarrettWu: OIC. I was thinking it to start from leaf, but it finds the leafs from the root. Cool.

GarrettWu (approved):

PR #1203, opened 2024-12-11
perf: Improve isin performance

TrevorBergeron:


bigframes/core/nodes.py
```diff
@@ -284,6 +290,24 @@ def prune(self, used_cols: COLUMN_SET) -> BigFrameNode:
         return self.transform_children(lambda x: x.prune(used_cols))
 
 
+class AdditiveNode:
+    """Definition of additive - if you drop added_fields, you end up with the descendent."""
```
tswast, line 294: A picture might help :-) 

```suggestion
    """Definition of additive - if you drop added_fields, you end up with the descendent.

    .. code-block:: text

        AdditiveNode (fields: a, b, c; added_fields: c)
            |
            |  additive_base
            V
        BigFrameNode (fields: a, b)

    """
```

See https://stackoverflow.com/a/50956831/101923 for creating a plain text block

bigframes/core/nodes.py
```diff
@@ -381,6 +405,106 @@ def remap_refs(self, mappings: Mapping[bfet_ids.ColumnId, bfet_ids.ColumnId]):
...
+    @property
+    def node_defined_ids(self) -> Tuple[bfet_ids.ColumnId, ...]:
+        return (self.indicator_col,)
+
+    @property
+    def additive_base(self) -> BigFrameNode:
+        return self.left_child
+
+    def replace_additive_base(self, node: BigFrameNode):
+        return dataclasses.replace(self, left_child=node)
```
tswast, line 481: Thinking aloud: I wonder if there's some way we can organize these sorts of tree transformations so that it's easier to reason about which can be applied in which order?

tswast (approved):

PR #1220, opened 2024-12-14
perf: Prevent deferring filters with remote functions

TrevorBergeron:


bigframes/core/compile/compiled.py
```diff
@@ -132,12 +133,22 @@ def projection(
         expression_id_pairs: typing.Tuple[typing.Tuple[ex.Expression, str], ...],
     ) -> T:
         """Apply an expression to the ArrayValue and assign the output to a column."""
+
+        # Remote ops are expensive so force reprojection before to ensure filters get applied first
+        any_remote_op = any(
+            ex.contains_op(expr, bigframes.operations.RemoteFunctionOp)
```
tswast, line 139: I'm a little bit concerned about hard coding remote function op here. Would it be better to flag each scalar operator as "expensive" / "cheap"? Certainly there's a scale there, but I guess a distinction could be if there's an associated cost (e.g. cloud run CPU in this case) per call. LLM-style ops or even BQML prediction (though that's not a scalar in BigQuery terms, though maybe could be in Bigframes eventually) could similarly have an "there's an extra cost associated here" flag.

TrevorBergeron:
Instead, removing all filter deferment in https://github.com/googleapis/python-bigquery-dataframes/pull/1336

PR #1222, opened 2024-12-16
feat: add `LogisticRegression.predict_explain()` to generate `ML.EXPLAIN_PREDICT` columns

arwas11:
  - `predict_explain`: screen/AGL7LuKUavHmYbn

Fixes #379743612 🦕

### DIFF
--- a/tests/system/small/ml/test_linear_model.py
+++ b/tests/system/small/ml/test_linear_model.py
@@ -307,48 +307,15 @@ def test_logistic_model_predict(penguins_logistic_model, new_penguins_df):
     )
 
 
-def test_logistic_model_predict_explain(
-    penguins_logistic_model: linear_model.LogisticRegression, new_penguins_df
-):
-    predictions = penguins_logistic_model.predict_explain(new_penguins_df).to_pandas()
-    assert predictions.shape == (3, 13)
-    result = predictions[["predicted_sex", "probability"]]
-    expected = pandas.DataFrame(
-        {
-            "predicted_sex": ["MALE", "MALE", "FEMALE"],
-            "probability": [
-                0.70692675811801065,
-                0.923363163640252,
-                0.99555295967825908,
-            ],
-        },
-        index=pandas.Index([1633, 1672, 1690], name="tag_number", dtype="Int64"),
-    )
-    expected["predicted_sex"] = expected["predicted_sex"].astype(
-        pandas.StringDtype(storage="pyarrow")
-    )
-    expected["probability"] = expected["probability"].astype(pandas.Float64Dtype())
-    pandas.testing.assert_frame_equal(
-        result.sort_index(),
-        expected,
-        check_exact=False,
-        rtol=0.1,
-    )
-
-
 def test_logistic_model_predict_params(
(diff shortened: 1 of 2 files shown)


tests/system/small/ml/test_linear_model.py
```diff
@@ -307,6 +307,83 @@ def test_logistic_model_predict(penguins_logistic_model, new_penguins_df):
     )
 
 
+def test_logistic_model_predict_explain(
+    penguins_logistic_model: linear_model.LogisticRegression, new_penguins_df
+):
+    predictions = penguins_logistic_model.predict_explain(new_penguins_df).to_pandas()
+    assert predictions.shape == (3, 13)
+    result = predictions[["predicted_sex", "probability"]]
```
tswast, line 315: It looks like these are the same columns you'd get with a regular `predict` call. Are there any additional columns we can add to make sure they are present?

tests/system/small/ml/test_linear_model.py
```diff
@@ -307,6 +307,83 @@ def test_logistic_model_predict(penguins_logistic_model, new_penguins_df):
     )
 
 
+def test_logistic_model_predict_explain(
```
tswast, line 310: Since this is so similar to the predict test and the other predict explain tests, I think we can remove it. The "params" test should be sufficient for what we need, which is just to make sure we called the right method.

tests/system/small/ml/test_linear_model.py
tswast: Alternatively, it might be best to remove this test, since we have others that test the same functionality.

tests/system/small/ml/test_linear_model.py
```diff
@@ -307,6 +307,83 @@ def test_logistic_model_predict(penguins_logistic_model, new_penguins_df):
...
+    expected["probability"] = expected["probability"].astype(pandas.Float64Dtype())
+    pandas.testing.assert_frame_equal(
+        result.sort_index(),
+        expected,
+        check_exact=False,
+        rtol=0.1,
+    )
+
+
+def test_logistic_model_predict_params(
```
tswast, line 339: Looks like this test is a duplicate of the one below (both are calling predict_explain). Let's remove this one.

tests/system/small/ml/test_linear_model.py
```diff
@@ -307,6 +307,83 @@ def test_logistic_model_predict(penguins_logistic_model, new_penguins_df):
...
+        "culmen_length_mm",
+        "culmen_depth_mm",
+        "flipper_length_mm",
+        "body_mass_g",
+        "sex",
+    }
+    assert expected_columns <= prediction_columns
+
+
+def test_logistic_model_predict_explain_params(
```
tswast, line 363: Keep this test.

tests/system/small/ml/test_linear_model.py
arwas11: SGTM! I'll go ahead and remove it.

tswast:
e2e failure `FAILED tests/system/large/ml/test_model_selection.py::test_cross_validate[4-4]` doesn't appear related.

tswast (approved):
Thanks!

PR #1232, opened 2024-12-19
chore: add experimental series.blob.display() function

GarrettWu:
Verified test env screen/38WRuQPuvkYP8PN

### DIFF
--- a/bigframes/operations/blob.py
+++ b/bigframes/operations/blob.py
@@ -14,9 +14,12 @@
 
 from __future__ import annotations
 
-import bigframes
+import IPython.display as ipy_display
+import requests
+
 from bigframes.operations import base
 import bigframes.operations as ops
+import bigframes.series
 
 
 class BlobAccessor(base.SeriesMethods):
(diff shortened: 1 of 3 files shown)


Genesis929 (approved):

bigframes/operations/blob.py
```diff
@@ -37,7 +40,29 @@ def metadata(self):
         details_json = self._apply_unary_op(ops.obj_fetch_metadata_op).struct.field(
             "details"
         )
-
         import bigframes.bigquery as bbq
 
         return bbq.json_extract(details_json, "$.gcs_metadata")
+
+    def display(self, n: int = 3):
```
shobsi, line 47: can we call it peek() to have some semantic resemblance to the dataframe/series peek()?

bigframes/operations/blob.py
```diff
@@ -37,7 +40,29 @@ def metadata(self):
...
+
+        .. note::
+            BigFrames Blob is still under experiments. It may not work and subject to change in the future.
+
+        Args:
+            n (int, default 3): number of sample blob objects to display.
+        """
+        import bigframes.bigquery as bbq
+
+        s = bigframes.series.Series(self._block).head(n)
```
shobsi, line 58: relying on `peek` instead of `head` would make it relevant in partial ordering mode as well

PR #1237, opened 2024-12-20
fix: ARIMAPlus series input

Genesis929:

### DIFF
--- a/bigframes/ml/forecasting.py
+++ b/bigframes/ml/forecasting.py
@@ -199,14 +199,14 @@ class ARIMAPlus(base.SupervisedTrainablePredictor):
         Returns:
             ARIMAPlus: Fitted estimator.
         """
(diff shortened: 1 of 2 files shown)


bigframes/ml/forecasting.py
```diff
@@ -199,11 +200,12 @@ def _fit(
         Returns:
             ARIMAPlus: Fitted estimator.
         """
-        if X.columns.size != 1:
+
+        if not isinstance(X, (pd.Series, bpd.Series)) and X.columns.size != 1:
             raise ValueError(
                 "Time series timestamp input X must only contain 1 column."
             )
-        if y.columns.size != 1:
```
GarrettWu, line 206: to generalize, convert to DataFrame first, then check columns. Then you don't need different logic for Series and DF.

GarrettWu (approved):

PR #1247, opened 2025-01-02
docs: update `bigframes.pandas.pandas` docstrings

arwas11:

Fixes #296390934 🦕

### DIFF
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -641,7 +641,7 @@ class Session(
                 to load from the default project.
 
         Returns:
-            bigframes.ml model:
+            model:
                 A bigframes.ml Model, Transformer or Pipeline wrapping the model.
         """
         import bigframes.ml.loader
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -633,7 +633,8 @@ class Session(
                 to load from the default project.
 
         Returns:
-            A bigframes.ml Model, Transformer or Pipeline wrapping the model.
+            bigframes.ml model:
+                A bigframes.ml Model, Transformer or Pipeline wrapping the model.
         """
         import bigframes.ml.loader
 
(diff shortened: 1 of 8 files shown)


bigframes/session/__init__.py
```diff
@@ -641,7 +641,8 @@ def read_gbq_model(self, model_name: str):
                 to load from the default project.
 
         Returns:
-            A bigframes.ml Model, Transformer or Pipeline wrapping the model.
+            bigframes.ml model:
```
shobsi, line 644: just "model" should suffice

shobsi (approved):

bigframes/session/__init__.py
```diff
@@ -641,7 +641,8 @@ def read_gbq_model(self, model_name: str):
                 to load from the default project.
 
         Returns:
-            A bigframes.ml Model, Transformer or Pipeline wrapping the model.
+            model:
```
tswast, line 644: This isn't the actual type name, so better to either omit it or change it to the actual type, which would be `Union[all the possible types]` or a shared superclass.

The `from_bq` method uses a big `Union`:

https://github.com/googleapis/python-bigquery-dataframes/blob/796fc3ed59bc0bd79f566ce2bc604f4cf2079fb3/bigframes/ml/loader.py#L84-L105

bigframes/session/__init__.py
```diff
@@ -1545,12 +1546,13 @@ def read_gbq_function(
                 a pandas Series.
 
         Returns:
-            callable: A function object pointing to the BigQuery function read
-            from BigQuery.
+            callable:
```
tswast, line 1549: I think this should be a capital `C`. It's a reference to https://docs.python.org/3/library/collections.abc.html#collections.abc.Callable

Might even want to use `collections.abc.Callable` for clarity.

bigframes/session/__init__.py
arwas11: I'll omit and revert back to just `A bigframes.ml Model, Transformer or Pipeline wrapping the model.` (similar  format to the `from_bq` return section).

tswast (approved):
LGTM. Let's figure out why these returns sections aren't rendering, though.

PR #1251, opened 2025-01-03
feat: implement confirmation threshold for semantic operators

sycai:
If the amount of row to be processed exceeds this threshold, we will ask for keyboard input from user to confirm.

### DIFF
--- a/bigframes/_config/compute_options.py
+++ b/bigframes/_config/compute_options.py
@@ -71,7 +71,7 @@ class ComputeOptions:
             If the number of rows exceeds the threshold, the user will be asked to confirm
             their operations to resume. The default value is 0. Set the value to None
             to turn off the guard.
-        semmantic_ops_confirmation_threshold (bool):
+        semantic_ops_threshold_autofail (bool):
             Guards against unexepcted processing of large amount of rows by semantic operators.
             When set to True, the operation automatically fails without asking for user inputs.
     """
--- a/bigframes/_config/compute_options.py
+++ b/bigframes/_config/compute_options.py
@@ -66,6 +66,11 @@ class ComputeOptions:
             engine to handle. However this comes at the cost of increase cost and latency.
         extra_query_labels (Dict[str, Any], Options):
             Stores additional custom labels for query configuration.
+        sem_ops_confirmation_threshold (int, Options):
+            Guards against unexepcted processing of large amount of rows by semantic operators.
+            If the number of rows exceeds the threshold, the user will be asked to confirm
+            their operations to resume. The default value is 25. Set the value to None
+            to turn off the guard.
     """
 
     maximum_bytes_billed: Optional[int] = None
@@ -73,6 +78,7 @@ class ComputeOptions:
     extra_query_labels: Dict[str, Any] = dataclasses.field(
         default_factory=dict, init=False
     )
+    sem_ops_confirmation_threshold: Optional[int] = 25
 
     def assign_extra_query_labels(self, **kwargs: Any) -> None:
         """
--- a/bigframes/_config/compute_options.py
+++ b/bigframes/_config/compute_options.py
@@ -71,6 +71,9 @@ class ComputeOptions:
             If the number of rows exceeds the threshold, the user will be asked to confirm
             their operations to resume. The default value is 0. Set the value to None
             to turn off the guard.
+        semmantic_ops_confirmation_threshold (bool):
+            Guards against unexepcted processing of large amount of rows by semantic operators.
+            When set to True, the operation automatically fails without asking for user inputs.
     """
 
     maximum_bytes_billed: Optional[int] = None
@@ -79,6 +82,7 @@ class ComputeOptions:
         default_factory=dict, init=False
     )
     semantic_ops_confirmation_threshold: Optional[int] = 0
+    semantic_ops_threshold_autofail = False
 
     def assign_extra_query_labels(self, **kwargs: Any) -> None:
         """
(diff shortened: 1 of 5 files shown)


bigframes/_config/compute_options.py
```diff
@@ -66,13 +66,19 @@ class ComputeOptions:
             engine to handle. However this comes at the cost of increase cost and latency.
         extra_query_labels (Dict[str, Any], Options):
             Stores additional custom labels for query configuration.
+        sem_ops_confirmation_threshold (int, Options):
```
tswast, line 69: Do you mean "optional"? I'm not sure what "options" means in this context.

Also, let's spell out `semantic`. I don't think `sem` is a common enough abbreviation to be self documenting.

```suggestion
        semantic_ops_confirmation_threshold (int, optional):
```

Also, consider adding an option to fail instead of asking for confirmation. When someone automates their script, they won't want it to hang waiting for input from standard in.

bigframes/operations/semantics.py
```diff
@@ -296,6 +300,9 @@ def cluster_by(
                 "It must be greater than 1."
             )
 
+        if not self._confirm_operation(len(self._df)):
+            return None
```
tswast, line 304: I'm a bit confused by the change of return type. This would make type checking much more difficult. Let's raise an exception saying the operation was cancelled instead.

bigframes/operations/semantics.py
```diff
@@ -1072,3 +1092,22 @@ def _validate_model(model):
...
+
+        threshold = bigframes.options.compute.sem_ops_confirmation_threshold
+
+        if threshold is None or row_count <= threshold:
+            return True
+
+        # Separate the prompt out. In IDE such VS Code, leaving prompt in the
+        # input function makes it less visible to the end user.
+        print(f"This operation will process about {row_count} rows. Proceed? [Y/n]")
+        reply = input().lower()
```
tswast, line 1109: Nit: `casefold()` is preferred for case insensitive checks (though doesn't matter much for ASCII characters). https://docs.python.org/3/library/stdtypes.html#str.casefold

bigframes/operations/semantics.py
```diff
@@ -1072,3 +1092,22 @@ def _validate_model(model):
 
         if not isinstance(model, GeminiTextGenerator):
             raise TypeError("Model is not GeminiText Generator")
+
+    @staticmethod
+    def _confirm_operation(row_count: int) -> bool:
```
tswast, line 1097: Let's have this raise instead of return. Otherwise, you'll have to change all the methods to return `Optional[Series]` / `Optional[DataFrame]` for type checking, which would be annoying to deal with for users in non-interactive use cases.

bigframes/operations/semantics.py
```diff
@@ -1072,3 +1092,22 @@ def _validate_model(model):
...
+        import bigframes
+
+        threshold = bigframes.options.compute.sem_ops_confirmation_threshold
+
+        if threshold is None or row_count <= threshold:
+            return True
+
+        # Separate the prompt out. In IDE such VS Code, leaving prompt in the
+        # input function makes it less visible to the end user.
+        print(f"This operation will process about {row_count} rows. Proceed? [Y/n]")
```
tswast, line 1108: Please print a line with a code sample explaining how to increase the threshold (or turn it off).

bigframes/_config/compute_options.py
```diff
@@ -66,13 +66,19 @@ class ComputeOptions:
...
+            their operations to resume. The default value is 25. Set the value to None
+            to turn off the guard.
     """
 
     maximum_bytes_billed: Optional[int] = None
     enable_multi_query_execution: bool = False
     extra_query_labels: Dict[str, Any] = dataclasses.field(
         default_factory=dict, init=False
     )
+    sem_ops_confirmation_threshold: Optional[int] = 25
```
tswast, line 81: I worry if someone tries this on small data and then tries to automate it, their script will hang waiting for input. Please set to 1 / 0 so that we always show this message and folks will know to turn it off for automated use cases.

tswast (changes requested):

bigframes/_config/compute_options.py
sycai: I added a "semantic_ops_threshold_autofail" option

bigframes/operations/semantics.py
sycai: SG!

tests/system/large/operations/test_semantics.py
```diff
@@ -447,8 +596,40 @@ def test_join(instruction, session, gemini_flash_model):
...
+    [
+        pytest.param("y"),
+        pytest.param(
+            "n", marks=pytest.mark.xfail(raises=exceptions.OperationAbortedError)
+        ),
+    ],
+)
+def test_join_with_confirmation(session, gemini_flash_model, reply, monkeypatch):
+    bigframes.options.experiments.semantic_operators = True
+    bigframes.options.compute.semantic_ops_confirmation_threshold = 0
```
tswast, line 610: Because our tests are multi-threaded, this has the possibility of introducing flakiness to our test suite.

Please use a `with bpd.option_context(...)` block to set these values. That will ensure thread-local values are used.

See: https://github.com/googleapis/python-bigquery-dataframes/blob/10f08daec6034aafe48096be56683c953accc79a/tests/system/small/test_progress_bar.py#L34-L35

sycai: Thanks for the example. Code updated

tswast (approved):

PR #1269, opened 2025-01-07
chore: unimplemented API tracking

Genesis929:

Currently this covers Series class and DataFrame class


bigframes/core/log_adapter.py
```diff
@@ -30,6 +37,78 @@
...
+        self.method_name = method_name
+
+    def __call__(self, *args, **kwargs):
+        submit_pandas_labels(
+            self.bq_client,
+            self.class_name,
+            self.method_name,
+            args,
+            kwargs,
+            task=PANDAS_API_TRACKING_TASK,
```
tswast, line 57: I'm a bit confused by this. Why the extra `task` argument?

bigframes/core/log_adapter.py
```diff
@@ -30,6 +37,78 @@
...
+        )
+        raise AttributeError(
+            "BigQuery DataFrames has not yet implemented an equivalent to"
+            f"'pandas.{self.class_name}.{self.method_name}'.\n{constants.FEEDBACK_LINK}"
+        )
+
+
+def submit_pandas_labels(
```
tswast, line 65: Please add a docstring explaining the purpose of this method.

bigframes/core/log_adapter.py
```diff
@@ -30,6 +37,78 @@
 _call_stack: List = []
 
 
+class UnimplementedMethodLogger:
```
tswast, line 40: Pease add a docstring.

bigframes/core/log_adapter.py
```diff
@@ -58,7 +137,18 @@ def wrapper(*args, **kwargs):
         _call_stack.append(full_method_name)
 
         try:
-            return method(*args, **kwargs)
+            return method(self, *args, **kwargs)
+        except (NotImplementedError, TypeError) as e:
+            if hasattr(self, "_block") and len(_call_stack) == 1:
```
tswast, line 142: Please add a comment explaining why we have these requirements in the if statement.

bigframes/dataframe.py
```diff
@@ -645,13 +644,8 @@ def __getattr__(self, key: str):
...
 
         if hasattr(pandas.DataFrame, key):
-            raise AttributeError(
-                textwrap.dedent(
-                    f"""
-                    BigQuery DataFrames has not yet implemented an equivalent to
-                    'pandas.DataFrame.{key}'. {constants.FEEDBACK_LINK}
-                    """
-                )
+            return log_adapter.UnimplementedMethodLogger(
```
tswast, line 647: I'm confused. Why would we ever return this class to the user?

bigframes/series.py
```diff
@@ -1299,13 +1298,8 @@ def __getattr__(self, key: str):
...
             raise AttributeError(key)
         elif hasattr(pandas.Series, key):
-            raise AttributeError(
-                textwrap.dedent(
-                    f"""
-                    BigQuery DataFrames has not yet implemented an equivalent to
-                    'pandas.Series.{key}'. {constants.FEEDBACK_LINK}
-                    """
-                )
+            return log_adapter.UnimplementedMethodLogger(
```
tswast, line 1301: Likewise, returning here seems wrong.

tswast (changes requested):

bigframes/dataframe.py
Genesis929: Return this is because gettattr won't get the args and kwargs. By returning the class, the class will be called as a function and receive the args and kwargs, and then we can set the labels.

bigframes/series.py
Genesis929: Return this is because gettattr won't get the args and kwargs. By returning the class, the class will be called as a function and receive the args and kwargs, and then we can set the labels.

bigframes/core/log_adapter.py
Genesis929: Task is to identify why the labels are collected. The  PANDAS_API_TRACKING_TASK means the method is missing(like df.resample). The PANDAS_PARAM_TRACKING_TASK means args or kwargs are missing, but the method has already being implemented. Later in dashboard this may help.

bigframes/core/log_adapter.py
Genesis929: Updated

bigframes/core/log_adapter.py
Genesis929: Updated

bigframes/core/log_adapter.py
Genesis929: Updated

bigframes/core/log_adapter.py
```diff
@@ -30,6 +37,126 @@
...
+        bq_client (bigquery.Client): The BigQuery client used to submit the log.
+        class_name (str): The name of the class in which the unimplemented method belongs.
+        method_name (str): The name of the unimplemented method being called.
+    """
+
+    def __init__(self, bq_client: bigquery.Client, class_name: str, method_name: str):
+        # Workaround for issue with notebook, this need to raise early so
+        # it can find an alternative _repr_ method.
+        if method_name in ["_repr_latex_"]:
+            raise
```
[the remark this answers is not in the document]
Genesis929: Updated.

bigframes/dataframe.py
Genesis929: Tested and descriptor won't work, __get__, __set__ seems never called, I'm trying another workaround for property now.

tswast: What if they are trying to access a property instead of method? Then we would just be giving them an object they can't use.

How important is it that we capture the args?

tswast: Please drop the args tracking and instead raise. This could cause incorrect customer code that is difficult to debug if they try to access a property we haven't implemented.

Genesis929: Removed, for property a workaround is to read the pandas type in __init__, since we always has a corresponding pandas_api if we want to submit the information. Tried that and works, but it's not included for now.

bigframes/core/log_adapter.py
```diff
@@ -13,15 +13,21 @@
...
+import pandas
+
 _lock = threading.Lock()
 
 # The limit is 64 (https://cloud.google.com/bigquery/docs/labels-intro#requirements),
 # but leave a few spare for internal labels to be added.
 # See internal issue 386825477.
 MAX_LABELS_COUNT = 64 - 8
+PANDAS_API_TRACKING_TASK = "pandas_api_tracking"
+PANDAS_PARAM_TRACKING_TASK = "pandas_param_tracking"
```
tswast, line 30: We don't need this anymore now that we are only tracking API names, right?

tests/unit/core/test_log_adapter.py
```diff
@@ -61,3 +70,112 @@ def test_get_and_reset_api_methods(test_instance):
...
+
+
+@pytest.mark.parametrize(
+    ("class_name", "method_name", "args", "kwargs"),
+    (
+        (
+            "DataFrame",
+            "resample",
+            [],
+            {"aa": "bb"},
```
tswast, line 120: Same here: It's not clear from this test definition that these should _not_ be included as labels.

tests/unit/core/test_log_adapter.py
```diff
@@ -61,3 +70,112 @@ def test_get_and_reset_api_methods(test_instance):
...
+
+
+@pytest.mark.parametrize(
+    ("class_name", "method_name", "args", "kwargs", "task"),
+    (
+        (
+            "DataFrame",
+            "resample",
+            ["a", "b", "c"],
+            {"aa": "bb", "rule": "1s"},
```
tswast, line 82: It's not clear from this test definition that `aa` should _not_ be included in the job labels.

Please add a parameter: "expected_labels" to these test parametrizations.

tests/unit/core/test_log_adapter.py
```diff
@@ -61,3 +70,112 @@ def test_get_and_reset_api_methods(test_instance):
...
+
+
+@pytest.mark.parametrize(
+    ("class_name", "method_name", "args", "kwargs"),
+    (
+        (
+            "DataFrame",
+            "resample",
+            [],
+            {"aa": "bb"},
```
tswast, line 160: Same here: I think an "expected_labels" parameter would be helpful.

bigframes/core/log_adapter.py
```diff
@@ -30,6 +36,75 @@
...
+    else:
+        return
+
+    if hasattr(cls, method_name):
+        method = getattr(cls, method_name)
+    else:
+        return
+
+    if kwargs:
+        signature = inspect.signature(method)
```
tswast, line 84: Needs comments explaining why we are inspecting (e.g. to only include labels for kwargs that are present in pandas)

Genesis929: Updated.

tests/unit/core/test_log_adapter.py
Genesis929: Updated

tests/unit/core/test_log_adapter.py
Genesis929: Updated, and merged with test_submit_pandas_labels test.

tests/unit/core/test_log_adapter.py
Genesis929: In this case, there won't be labels, as we won't call bqclient.query.

Simplified the test a little.

bigframes/core/log_adapter.py
Genesis929: It's not exactly necessary but still can have some use, we are still collecting args in case of missing args.

bigframes/core/log_adapter.py
```diff
@@ -30,6 +36,75 @@
...
+        return
+
+    if hasattr(cls, method_name):
+        method = getattr(cls, method_name)
+    else:
+        return
+
+    if kwargs:
+        # Iterate through the keyword arguments and add them to the labels dictionary if they
+        # are valid parameters and the maximum label count has not been reached.
```
tswast, line 85:
```suggestion
        # are parameters that are implemented in pandas and the maximum label count has not been reached.
```

tswast (approved):

PR #1278, opened 2025-01-09
deps: remove sqlglot upper bound

### LINKED ISSUE #942: Optionally relax sqlglot / ibis pin via env var for use in SQLMesh
https://github.com/googleapis/python-bigquery-dataframes/blob/04139644af0c30c10519397fb3be7bebf6ec813c/setup.py#L63

There is very strong synergy between bigframes and https://github.com/TobikoData/sqlmesh
However because of the sqlglot pin and the transitive sqlglot pin in the ibis pin, this library is incompatible with sqlmesh. (sqlmesh is built and maintained by creators of sqlglot and therefore usually has latest pin, and a strict one at that)

Since you are using setup.py, we should be able to use an env var or something similar to have an `opt-in` way to remove the version constraints on those two deps such that they can be defined by the consumer. We would accept that it opens one up to undefined behavior but by contrast it frees you from needing to constantly update your pins and allows more experimentation and adoption in the community. 

You can see an example sqlmesh model using bigframes here https://github.com/TobikoData/sqlmesh/pull/2667/files

z3z1ma:

Fixes #942  🦕

Additionally this allows use of bigframes with SQLMesh. If this is merged, I will build first class support into SQLMesh which opens up BigFrames usage to the audience of folks building transformations for their org with that tool. 

The upper bound is removed in this PR. This helps libraries include bigframes as a package. SQLMesh team are also the stewards of SQLGlot and hence they ride the bleeding edge. An upper bound would eventually introduce dependency hell (unresolvable). I checked the API surface area of what is being used in sqlglot. It seems quite minor outside what `ibis/expr/sql.py` touches.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,14 +57,11 @@ dependencies = [
     "requests >=2.27.1",
     "scikit-learn >=1.2.2",
     "sqlalchemy >=1.4,<3.0dev",
-    # Keep sqlglot versions in sync with ibis-framework. This avoids problems
-    # where the incorrect version of sqlglot is installed, such as
-    # https://github.com/googleapis/python-bigquery-dataframes/issues/315
-    "sqlglot >=23.6.3,<25.2",
-    "tabulate >= 0.9",
+    "sqlglot >=23.6.3",
+    "tabulate >=0.9",
     "ipywidgets >=7.7.1",
-    "humanize >= 4.6.0",
-    "matplotlib >= 3.7.1",
+    "humanize >=4.6.0",
+    "matplotlib >=3.7.1",
     # For vendored ibis-framework.
     "atpublic>=2.3,<6",
     "parsy>=2,<3",


testing/constraints-3.9.txt
```diff
@@ -21,7 +21,7 @@ pydata-google-auth==1.8.2
 requests==2.27.1
 scikit-learn==1.2.2
 sqlalchemy==1.4
-sqlglot==23.6.3
+sqlglot==26.1.3
```
tswast, line 24: We try to keep this and our lower-bound in sync so that we have at least one test session that runs with our advertised lower-bound.

Please either:

1. Revert this change
2. Update the lower bound in `setup.py`.

tswast (commented):
Thanks for the contribution!

I'm curious to see the results of our test suite with this change.

tswast:
Test failure is `FAILED tests/system/small/ml/test_llm.py::test_create_load_gemini_text_generator_model[gemini-1.5-pro-0` which appears to be a quota-related flake. I can retry in the morning.

tswast (approved):

PR #1280, opened 2025-01-10
fix: avoid global mutation in `BigQueryOptions.client_endpoints_override`

tswast:
This fixes a common "gotcha" in Python. Before this fix, mutations to this dictionary will affect all instances of the BigQueryOptions class, not just the current instance.

See: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

Fixes internal issue 389087778 🦕


GarrettWu (approved):

GarrettWu:
Thanks for the fix!

PR #1281, opened 2025-01-10
chore: format warning message

chelsea-lin:
This update improves the formatting of warning messages in BigFrames. For a detailed explanation of the decision-making process, please refer to the internal issue discussion (b/372728230).
- [X] Internal issue: 372728230

Fixes internal issue 372728230 🦕

### DIFF
--- a/bigframes/_config/bigquery_options.py
+++ b/bigframes/_config/bigquery_options.py
@@ -25,7 +25,7 @@ import jellyfish
 
 import bigframes.constants
 import bigframes.enums
-import bigframes.exceptions
+import bigframes.exceptions as bfe
 
 SESSION_STARTED_MESSAGE = (
     "Cannot change '{attribute}' once a session has started. "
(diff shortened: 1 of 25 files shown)


bigframes/_config/bigquery_options.py
```diff
@@ -271,10 +268,11 @@ def use_regional_endpoints(self, value: bool):
             )
 
         if value:
-            warnings.warn(
+            msg = (
                 "Use of regional endpoints is a feature in preview and "
                 "available only in selected regions and projects. "
             )
+            warnings.warn(msg)
```
tswast, line 275: Looks like this should raise preview category, too.

Also, we should almost always have at least `stacklevel=2` so that we have a chance of showing the warning in the user's code rather than pointing to this line, which isn't particularly helpful to the user.

tswast (approved):

bigframes/_config/bigquery_options.py
chelsea-lin: Good points. Fixed. Thanks!

PR #1291, opened 2025-01-15
chore: add experimental blob.display to support audio and video

GarrettWu:
Verified in test env screen/pEa4D8RzfveLjeq


bigframes/operations/blob.py
```diff
@@ -49,28 +49,66 @@ def metadata(self) -> bigframes.series.Series:
...
 
-    def display(self, n: int = 3):
+    def content_type(self) -> bigframes.series.Series:
+        """Retrive the content type of the Blob.
+
+        .. note::
+            BigFrames Blob is still under experiments. It may not work and subject to change in the future.
+
+        Returns:
+            JSON: json-string of the content type."""
```
Genesis929, line 59: Series of json

Genesis929 (approved):

bigframes/operations/blob.py
GarrettWu: Done.

PR #1294, opened 2025-01-16
fix: series sort_index and sort_values now raises when axis!=0

Genesis929:


bigframes/series.py
```diff
@@ -1347,6 +1347,8 @@ def value_counts(
     def sort_values(
         self, *, axis=0, ascending=True, kind: str = "quicksort", na_position="last"
     ) -> Series:
+        if axis != 0:
```
GarrettWu, line 1350: Is the behavior the same as pandas? Looks like pandas just ignores the param. https://pandas.pydata.org/docs/reference/api/pandas.Series.sort_index.html

Also it can take "index".

Genesis929: Updated

GarrettWu (approved):

PR #1296, opened 2025-01-16
deps: remove scikit-learn and sqlalchemy as required dependencies

Genesis929:


GarrettWu (approved):

noxfile.py
```diff
@@ -223,11 +223,17 @@ def run_unit(session, install_test_extra):
...
+        "scikit-learn>=1.2.2",
+    )
     run_unit(session, install_test_extra=True)
 
 
 @nox.session(python=UNIT_TEST_PYTHON_VERSIONS[-1])
 def unit_noextras(session):
+    session.install(
+        "scikit-learn>=1.2.2",
+    )
```
tswast, line 236: Do _not_ install optional dependencies in `unit_noextras`. The intention of such test sessions is to make sure that dependencies that are marked as optional are truly optional. This risks making a forced dependency on this package.

Also, prefer removing the package from other sessions as well.

noxfile.py
```diff
@@ -223,11 +223,17 @@ def run_unit(session, install_test_extra):
 
 @nox.session(python=UNIT_TEST_PYTHON_VERSIONS)
 def unit(session):
+    session.install(
+        "scikit-learn>=1.2.2",
+    )
```
tswast, line 228: If there is optional functionality that needs `scikit-learn`, please use the "extras" section in `setup.py`. Don't install packages this way.

tswast:
This is not a `chore: `. This is a user-visible change and should be labeled as such.

tswast (changes requested):
Please do not merge as is. We are opening our customers up to a big risk of having a broken installation by installing `scikit-learn` in so many of our test sessions.

noxfile.py
Genesis929: Updated, now test skipped is sklearn not installed.

For docs and docfx session, it's required to have sklearn, so changed the way of installation instead of remove it.

bigframes/ml/metrics/_metrics.py
```diff
@@ -176,9 +175,26 @@ def auc(
...
+    elif x_pandas.is_monotonic_increasing:
+        d = 1
+    else:
+        raise ValueError(f"x is neither increasing nor decreasing : {x_pandas}.")
+
+    if hasattr(np, "trapezoid"):
+        # new in numpy 2.0
+        return d * np.trapezoid(y_pandas, x_pandas)
+    # np.trapz has been deprecated in 2.0
+    return d * np.trapz(y_pandas, x_pandas)  # type: ignore
```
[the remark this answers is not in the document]
Genesis929: This is a simplified version from sklearn logic, because we need it to work for pandas series only. In this case should we put it in third-party?

noxfile.py
```diff
@@ -72,7 +72,13 @@
 UNIT_TEST_LOCAL_DEPENDENCIES: List[str] = []
 UNIT_TEST_DEPENDENCIES: List[str] = []
 UNIT_TEST_EXTRAS: List[str] = []
-UNIT_TEST_EXTRAS_BY_PYTHON: Dict[str, List[str]] = {"3.12": ["polars"]}
+UNIT_TEST_EXTRAS_BY_PYTHON: Dict[str, List[str]] = {
+    "3.9": ["scikit-learn"],
+    "3.10": ["scikit-learn"],
+    "3.11": ["scikit-learn"],
+    "3.12": ["polars", "scikit-learn"],
+    "3.13": ["scikit-learn"],
```
tswast, line 80: No.

We can't install it in _every_ session. We need to know that bigframes works without this package!

```suggestion
    "3.12": ["polars", "scikit-learn"],
```

noxfile.py
```diff
@@ -96,7 +102,7 @@
 ]
 SYSTEM_TEST_LOCAL_DEPENDENCIES: List[str] = []
 SYSTEM_TEST_DEPENDENCIES: List[str] = []
-SYSTEM_TEST_EXTRAS: List[str] = ["tests"]
+SYSTEM_TEST_EXTRAS: List[str] = ["tests", "scikit-learn"]
```
tswast, line 105: No.

We can't install it in _every_ session. We need to know that bigframes works without this package! Use `SYSTEM_TEST_EXTRAS_BY_PYTHON`, instead.

noxfile.py
```diff
@@ -510,7 +515,7 @@ def docs(session):
 def docfx(session):
     """Build the docfx yaml files for this library."""
 
-    session.install("-e", ".")
+    session.install("-e", ".[scikit-learn]")
```
tswast, line 518: I'm very concerned if we need to install `scikit-learn` just to build the docs. Please remove.

```suggestion
    session.install("-e", ".")
```

noxfile.py
```diff
@@ -468,8 +474,7 @@ def cover(session):
 @nox.session(python=DEFAULT_PYTHON_VERSION)
 def docs(session):
     """Build the docs for this library."""
-
-    session.install("-e", ".")
+    session.install("-e", ".[scikit-learn]")
```
tswast, line 477: I'm very concerned if we need to install `scikit-learn` just to build the docs. Please remove.

```suggestion
    session.install("-e", ".")
```

tswast (changes requested):
You did not address my feedback. You just changed how you install `scikit-learn`. You're still installing `scikit-learn` in way too many of our test sessions.

noxfile.py
Genesis929: Updated, but I assume in this case unit noextras sessions runs the same thing as it test?

noxfile.py
Genesis929: Updated.

noxfile.py
Genesis929: Not sure if it's possible to remove this as we are calling publish_api_coverage.py, and one of the things it do is go through sklearn APIs and generate a coverage report.

noxfile.py
Genesis929: Not sure if it's possible to remove this as we are calling publish_api_coverage.py, and one of the things it do is go through sklearn APIs and generate a coverage report.

bigframes/ml/metrics/_metrics.py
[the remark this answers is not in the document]
Genesis929: Moved to third party

noxfile.py
tswast: Good point. Thanks for clarifying.

tswast (approved):

PR #1301, opened 2025-01-17
refactor: ReadLocal and Explode nodes support offset outputs

TrevorBergeron:


tswast:
Notebook failure

```
FAILED notebooks/generative_ai/bq_dataframes_llm_kmeans.ipynb::bq_dataframes_llm_kmeans.ipynb - IndexError: single positional indexer is out-of-bounds
```

Likely unrelated?

tswast (approved):
LGTM.

A question: do we expect to add these `offsets_col` properties to all leaf-type nodes? I'm wondering if it there is a risk of us forgetting to add logic such as

```
        if self.offsets_col is not None:
            return itertools.chain(
                fields, (Field(self.offsets_col, bigframes.dtypes.INT_DTYPE),)
            )
```

in other node types?

TrevorBergeron:
> LGTM.
> 
> A question: do we expect to add these `offsets_col` properties to all leaf-type nodes? I'm wondering if it there is a risk of us forgetting to add logic such as
> 
> ```
>         if self.offsets_col is not None:
>             return itertools.chain(
>                 fields, (Field(self.offsets_col, bigframes.dtypes.INT_DTYPE),)
>             )
> ```
> 
> in other node types?

I don't think so. For ReadLocal, we want to leverage our ability to put offsets in the data locally before inlining/uploading to make explicit the local array ordering. This won't necessarily apply to other data sources, for which we may lack source-level offset-generating capabilities.

PR #1315, opened 2025-01-23
feat: Add DataFrame.corrwith method

TrevorBergeron:


bigframes/dataframe.py
```diff
@@ -1473,6 +1473,48 @@ def cov(self, *, numeric_only: bool = False) -> DataFrame:
...
         return result
 
+    def corrwith(
+        self,
+        other: typing.Union[DataFrame, bigframes.series.Series],
+        *,
+        numeric_only: bool = False,
+    ):
+        other_frame = other if isinstance(other, DataFrame) else other.to_frame()
+        if not numeric_only:
```
sycai, line 1483: Let's re-write this if-else block as

if numeric_only:
   ...
else
   ...

See go/tott/685

tests/system/small/test_dataframe.py
```diff
@@ -2246,6 +2246,41 @@ def test_cov_w_numeric_only(scalars_dfs_maybe_ordered, columns, numeric_only):
...
+    bf_result = scalars_df[l_cols].corrwith(scalars_df[r_cols]).to_pandas()
+    pd_result = scalars_pandas_df[l_cols].corrwith(scalars_pandas_df[r_cols])
+
+    # BigFrames and Pandas differ in their data type handling:
+    # - Column types: BigFrames uses Float64, Pandas uses float64.
+    # - Index types: BigFrames uses strign, Pandas uses object.
+    pd.testing.assert_series_equal(
+        bf_result, pd_result, check_dtype=False, check_index_type=False
+    )
+
```
sycai, line 2264: Perhaps we should add two more cases:

numeric_only = True, two dfs contain only numeric columns => computation proceeds successfully
numeric_only = True, one of the dfs has a non-numeric column => an error is raised.

TrevorBergeron: added

sycai (approved):

PR #1316, opened 2025-01-23
feat: Support python type as astype arg

TrevorBergeron:


bigframes/operations/generic_ops.py
```diff
@@ -56,17 +54,10 @@
 class AsTypeOp(base_ops.UnaryOp):
     name: typing.ClassVar[str] = "astype"
     # TODO: Convert strings to dtype earlier
-    to_type: typing.Union[dtypes.DtypeString, dtypes.Dtype]
```
shobsi, line 59: Isn't it cleaner to take care of str->dtype conversion centrally inside the op itself? Otherwise now we have to do the conversion separately in bunch of places - base, dataframe, series modules.

tests/unit/core/test_dtypes.py
```diff
@@ -197,45 +196,6 @@ def test_bigframes_dtype_converts(ibis_dtype, bigframes_dtype):
...
-            ibis_dtypes.Timestamp(timezone="UTC"),
-        ),
-        # Special case - "string" is acceptable for "string[pyarrow]"
-        ("string", ibis_dtypes.string),
-    ],
-)
-def test_bigframes_string_dtype_converts(ibis_dtype, bigframes_dtype_str):
-    """Test all the Ibis data types needed to read BigQuery tables"""
-    result = bigframes.core.compile.ibis_types.bigframes_dtype_to_ibis_dtype(
-        bigframes_dtype_str
```
shobsi, line 222: why not keep this test by passing `dtypes.bigframes_type(bigframes_dtype_str)` instead?

tests/unit/core/test_dtypes.py
```diff
@@ -197,45 +196,6 @@ def test_bigframes_dtype_converts(ibis_dtype, bigframes_dtype):
...
-    assert result == ibis_dtype
-
-
-def test_unsupported_dtype_raises_unexpected_datatype():
-    """Incompatible dtypes should fail when passed into BigQuery DataFrames"""
-    with pytest.raises(ValueError, match="Unexpected data type"):
-        bigframes.core.compile.ibis_types.bigframes_dtype_to_ibis_dtype(np.float32)
-
-
-def test_unsupported_dtype_str_raises_unexpected_datatype():
```
shobsi, line 233: This test still holds, no?

tests/unit/core/test_dtypes.py
```diff
@@ -197,45 +196,6 @@ def test_bigframes_dtype_converts(ibis_dtype, bigframes_dtype):
...
-    result = bigframes.core.compile.ibis_types.bigframes_dtype_to_ibis_dtype(
-        bigframes_dtype_str
-    )
-    assert result == ibis_dtype
-
-
-def test_unsupported_dtype_raises_unexpected_datatype():
-    """Incompatible dtypes should fail when passed into BigQuery DataFrames"""
-    with pytest.raises(ValueError, match="Unexpected data type"):
-        bigframes.core.compile.ibis_types.bigframes_dtype_to_ibis_dtype(np.float32)
```
shobsi, line 230: I believe this would pass since we support `np.floating` in `_infer_dtype_from_python_type`. Maybe repurpose and keep the test?

tests/system/small/test_index.py
```diff
@@ -123,6 +123,12 @@ def test_index_astype(scalars_df_index, scalars_pandas_df_index):
     pd.testing.assert_index_equal(bf_result, pd_result)
 
 
+def test_index_astype_python(scalars_df_index, scalars_pandas_df_index):
+    bf_result = scalars_df_index.set_index("int64_col").index.astype(float).to_pandas()
+    pd_result = scalars_pandas_df_index.set_index("int64_col").index.astype("Float64")
```
shobsi, line 128: Should be float-float parity?

TrevorBergeron: we use a different float type then pandas does, we use nullable Float64, while pandas uses non-nullable float64. So, we aren't quite matching the behavior.

bigframes/operations/generic_ops.py
TrevorBergeron: I think its best to canonicalize the type before putting it inside the op. We want to validate immediately anyways, so we may as well canonicalize at the same time.

tests/unit/core/test_dtypes.py
TrevorBergeron: test still works, brought it back and just amended error message

tests/unit/core/test_dtypes.py
TrevorBergeron: yeah, brought it back with new error message

tests/unit/core/test_dtypes.py
TrevorBergeron: ok, brought back test, now with mapping function

shobsi (approved):

PR #1318, opened 2025-01-23
feat: add `bigframes.bigquery.st_area` and suggest it from `GeoSeries.area`

arwas11:
  - `area`: screen/5SjGQsfX3K4mHpK
  -  `st_area`: screen/7X3xcQcsp968m4a, screen/3LrwWASNt26Whoa, screen/6VUmo2GL6BDjFKA

Fixes #393394363 🦕


tests/system/small/geopandas/test_geoseries.py
```diff
@@ -61,3 +66,24 @@ def test_geo_y(urban_areas_dfs):
...
+        dtype=GeometryDtype(),
+    )
+    bf_series: bigframes.geopandas.GeoSeries = s.geo
+    with pytest.raises(
+        NotImplementedError,
+        match=re.escape(
+            f"GeoSeries.area is not supported. Use bigframes.bigquery.st_area(series), instead. {constants.FEEDBACK_LINK}"
+        ),
+    ):
+        bf_series.area()
```
tswast, line 89: > mypy error: "Series" not callable  [operator]

Note: since `.area` is a "property", we don't call it as a function. We pretend it's just a normal attribute.

```suggestion
        bf_series.area
```

A bit more about properties: https://www.pythonmorsels.com/making-read-only-attribute/ and https://realpython.com/python-property/

It's very similar to what Java calls "getter" and "setter" functions, but in this case Python allows it to look like a normal attribute.

bigframes/geopandas/geoseries.py
```diff
@@ -39,3 +40,9 @@ def y(self) -> bigframes.series.Series:
         series = self._apply_unary_op(ops.geo_y_op)
         series.name = None
         return series
+
+    @property
+    def area(self, crs=None) -> bigframes.series.Series:
```
tswast, line 45: Let's try this to avoid the mypy failure:

```suggestion
    # GeoSeries.area overrides Series.area with something totally different.
    # Ignore this type error, as we are trying to be as close to geopandas as
    # we can.
    @property
    def area(self, crs=None) -> bigframes.series.Series:  # type: ignore
```

tests/system/small/bigquery/test_geo.py
```diff
@@ -0,0 +1,47 @@
...
+        Polygon([(0.10, 0.4), (0.9, 0.5), (0.10, 0.5)]),
+        Polygon([(0.1, 0.1), (0.2, 0.1), (0.2, 0.2)]),
+        LineString([(0, 0), (1, 1), (0, 1)]),
+        Point(0, 1),
+    ]
+
+    geopd_s = geopandas.GeoSeries(data=data, crs="EPSG:4326")
+    geobf_s = bigframes.geopandas.GeoSeries(data=data)
+
+    # Round both results to get an approximately similar output
```
arwas11, line 36: add comment about the number and projection site

tests/system/small/bigquery/test_geo.py
```diff
@@ -0,0 +1,47 @@
...
+from shapely.geometry import LineString, Point, Polygon  # type: ignore
+
+import bigframes.bigquery as bbq
+import bigframes.geopandas
+import bigframes.series
+
+
+def test_geo_st_area():
+    data = [
+        Polygon([(0.0, 0.0), (0.1, 0.1), (0.0, 0.1)]),
```
arwas11, line 26: make the coordinates smaller so the rounding is not as huge. Start with two zeros after the decimal point to round to the nearest thousands.

tests/system/small/bigquery/test_geo.py
```diff
@@ -0,0 +1,52 @@
...
+    geobf_s = bigframes.geopandas.GeoSeries(data=data)
+
+    # For `geopd_s`, the data was further projected with `geopandas.GeoSeries.to_crs`
+    # to `to_crs(26393)` to get the area in square meter. See: https://geopandas.org/en/stable/docs/user_guide/projections.html
+    # and https://spatialreference.org/ref/epsg/26393/. We then rounded both results
+    # to get them as close to each other as possible. Initially, the area results
+    # were +ten-millions. We added more zeros after the decimal point to round the
+    # area results to the nearest thousands.
+    geopd_s_result = geopd_s.to_crs(26393).area.round(-3)
+    geobf_s_result = bbq.st_area(geobf_s).to_pandas().round(-3)
```
tswast, line 43: Could we add a check that the values are at least 1000? I'd be slightly afraid that both round to 0, which would be a bit less useful of a test.

```suggestion
    assert geobf_s_result.iloc[0] >= 1000
```

tswast (approved):

tests/system/small/bigquery/test_geo.py
```diff
@@ -0,0 +1,52 @@
...
+    geobf_s = bigframes.geopandas.GeoSeries(data=data)
+
+    # For `geopd_s`, the data was further projected with `geopandas.GeoSeries.to_crs`
+    # to `to_crs(26393)` to get the area in square meter. See: https://geopandas.org/en/stable/docs/user_guide/projections.html
+    # and https://spatialreference.org/ref/epsg/26393/. We then rounded both results
+    # to get them as close to each other as possible. Initially, the area results
+    # were +ten-millions. We added more zeros after the decimal point to round the
+    # area results to the nearest thousands.
+    geopd_s_result = geopd_s.to_crs(26393).area.round(-3)
+    assert geobf_s_result.iloc[0] >= 1000
```
tswast, line 43: Sorry, I messed up the last suggested edit.

```suggestion
    geobf_s_result = bbq.st_area(geobf_s).to_pandas().round(-3)
    assert geobf_s_result.iloc[0] >= 1000
```

tswast:
Failed doctests are 

```
FAILED bigframes/session/__init__.py::bigframes.session.Session.read_gbq_function
FAILED bigframes/pandas/io/api.py::bigframes.pandas.io.api.read_gbq_function
```

which appears unrelated.

PR #1346, opened 2025-01-31
chore: support timestamp subtractions

sycai:
This PR enables subtraction operations for for Timestamp and datetime types.  

We don't support mix-match timestamp and datetime values in the same operations. It's not allowed in Ibis anyway.


bigframes/core/compile/compiler.py
```diff
@@ -109,7 +110,8 @@ def _preprocess(self, node: nodes.BigFrameNode):
     @functools.lru_cache(maxsize=5000)
     def compile_node(self, node: nodes.BigFrameNode) -> compiled.UnorderedIR:
         """Compile node into CompileArrayValue. Caches result."""
-        return self._compile_node(node)
+        # Need to dispatch op before compilation to keep it consistent with the compile_sql() call
+        return self._compile_node(nodes.bottom_up(node, rewrites.op_dynamic_dispatch))
```
TrevorBergeron, line 114: lets not run this on every node, instead, lets revive the dead `_preprocess` helper and apply all the pre-transforms there to the entire tree before running `compile_node` on the root

bigframes/core/rewrite/__init__.py
```diff
@@ -15,13 +15,15 @@
...
 from bigframes.core.rewrite.legacy_align import legacy_join_as_projection
+from bigframes.core.rewrite.operators import op_dynamic_dispatch
 from bigframes.core.rewrite.order import pull_up_order
 from bigframes.core.rewrite.slices import pullup_limit_from_slice, rewrite_slice
 
 __all__ = [
     "legacy_join_as_projection",
     "try_row_join",
     "rewrite_slice",
+    "op_dynamic_dispatch",
```
TrevorBergeron, line 26: I think something like "convert_duration_to_int" capture the high level intent best

bigframes/core/rewrite/operators.py
```diff
@@ -0,0 +1,78 @@
...
+def op_dynamic_dispatch(root: nodes.BigFrameNode) -> nodes.BigFrameNode:
+    if isinstance(root, nodes.ProjectionNode):
+        updated_assignments = tuple(
+            (_rewrite_expressions(expr, root.schema).expr, column_id)
+            for expr, column_id in root.assignments
+        )
+        root = nodes.ProjectionNode(root.child, updated_assignments)
+
+    # TODO(b/394354614): FilterByNode and OrderNode also contain expressions. Need to update them too.
+    return root
```
TrevorBergeron, line 40: as long as we get support those nodes before anybody starts using this!

bigframes/core/rewrite/operators.py
```diff
@@ -0,0 +1,78 @@
...
+        return _TypedExpr(expr, schema.get_type(expr.id.sql))
+
+    if isinstance(expr, ex.ScalarConstantExpression):
+        return _TypedExpr(expr, expr.dtype)
+
+    if isinstance(expr, ex.OpExpression):
+        updated_inputs = tuple(
+            map(lambda x: _rewrite_expressions(x, schema), expr.inputs)
+        )
+        return _rewrite_op_expr(expr, updated_inputs)
```
TrevorBergeron, line 55: I believe this will also need to be top-down rather than bottom-up.

bigframes/operations/datetime_ops.py
```diff
@@ -107,3 +107,22 @@ def output_type(self, *input_types: dtypes.ExpressionType) -> dtypes.ExpressionT
...
+    def output_type(self, *input_types: dtypes.ExpressionType) -> dtypes.ExpressionType:
+        if input_types[0] is not input_types[1]:
+            raise TypeError(
+                f"two inputs have different types. left: {input_types[0]}, right: {input_types[1]}"
+            )
+
+        if not dtypes.is_datetime_like(input_types[0]):
+            raise TypeError("expected timestamp input")
+
+        return dtypes.TIMEDETLA_DTYPE
```
TrevorBergeron, line 125: typo

bigframes/series.py
```diff
@@ -805,10 +805,14 @@ def __rsub__(self, other: float | int | Series) -> Series:
...
+    def sub(
+        self, other: float | int | pandas.Timestamp | datetime.datetime | Series
+    ) -> Series:
         return self._apply_binary_op(other, ops.sub_op)
 
-    def rsub(self, other: float | int | Series) -> Series:
+    def rsub(
+        self, other: float | int | pandas.Timestamp | datetime.datetime | Series
+    ) -> Series:
         return self._apply_binary_op(other, ops.sub_op, reverse=True)
```
TrevorBergeron, line 816: We might want to consider giving up on annotating `other` allowed dtypes

bigframes/series.py
```diff
@@ -2080,3 +2084,10 @@ def str(self) -> strings.StringMethods:
 
 def _is_list_like(obj: typing.Any) -> typing_extensions.TypeGuard[typing.Sequence]:
     return pandas.api.types.is_list_like(obj)
+
+
+def _has_timestamp_type(input: typing.Any) -> bool:
+    if isinstance(input, Series):
+        return bigframes.dtypes.is_datetime_like(input.dtype)
+
+    return isinstance(input, (pandas.Timestamp, datetime.datetime))
```
TrevorBergeron, line 2093: dead code?

bigframes/core/compile/compiler.py
```diff
@@ -58,6 +58,7 @@ def compile_sql(
         # TODO: get rid of output_ids arg
         assert len(output_ids) == len(list(node.fields))
         node = set_output_names(node, output_ids)
+        node = nodes.bottom_up(node, rewrites.op_dynamic_dispatch)
```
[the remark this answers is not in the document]
sycai: Sure, though I think it shouldn't matter much, because all node schemas are already stable at this point.

bigframes/core/compile/compiler.py
sycai: SG. Moved the code to `_preprocess`

bigframes/core/rewrite/__init__.py
sycai: I named it "rewrite_timedelta_ops" to better indicate that we are replacing the operators, not the values.

bigframes/core/rewrite/operators.py
sycai: PR soon to follow!

bigframes/core/rewrite/operators.py
sycai: I don't think it's possible to do this top-down, because we cannot get the input types by first processing the parent node. The parent output type can only be decided once we have rewrite all the subtrees.

bigframes/operations/datetime_ops.py
sycai: Nice catch. I'm glad we haven't officially announced this feature

bigframes/series.py
sycai: It makes sense. The operators themselves will perform type check for us anyway.

bigframes/series.py
sycai: removed

TrevorBergeron (approved):

PR #1347, opened 2025-01-31
perf: Prevent inlining of remote ops

TrevorBergeron:


tswast:
@TrevorBergeron  Can you please provide some more context around this change?

TrevorBergeron:
> @TrevorBergeron Can you please provide some more context around this change?

Basically, remote functions were being merged into subsequent selects, possibly inlining multiple times into different references. This was causing the SQL engine to call the remote function with duplicate calls. This change stops us from inlining remote function op SQL.

shobsi (approved):

PR #1349, opened 2025-02-01
chore: support timedeltas for read_pandas()

sycai:


bigframes/core/compile/compiler.py
```diff
@@ -173,6 +175,14 @@ def compile_readlocal(self, node: nodes.ReadLocalNode):
             io.BytesIO(node.feather_bytes),
             columns=[item.source_id for item in node.scan_list.items],
         )
+
+        # Convert timedeltas to microseconds for compatibility with BigQuery
+        for col in array_as_pd.columns:
+            if pdtypes.is_timedelta64_dtype(array_as_pd[col].dtype):
+                array_as_pd[col] = array_as_pd[col].map(utils.timedelta_to_micros)
+        if pdtypes.is_timedelta64_dtype(array_as_pd.index.dtype):
+            array_as_pd.index = array_as_pd.index.map(utils.timedelta_to_micros)
```
TrevorBergeron, line 184: I'm not sure we actually us the index, can probably ignore it

bigframes/core/schema.py
```diff
@@ -38,9 +38,13 @@ class ArraySchema:
     items: typing.Tuple[SchemaItem, ...]
 
     @classmethod
-    def from_bq_table(cls, table: google.cloud.bigquery.Table):
+    def from_bq_table(
+        cls,
+        table: google.cloud.bigquery.Table,
+        override_types: typing.Dict[str, bigframes.dtypes.Dtype] = {},
```
TrevorBergeron, line 44: I don't know about this override thing. If the caller has overrides in mind, they may as well just figure out the whole schema themself rather than call this method.

bigframes/core/utils.py
```diff
@@ -184,3 +184,7 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return decorator
+
+
+def timedelta_to_micros(td: pd.Timedelta) -> int:
+    return round(td.total_seconds() * 1_000_000)
```
TrevorBergeron, line 190: why not call `.microseconds`?

bigframes/session/_io/pandas.py
```diff
@@ -38,6 +39,7 @@ class DataFrameAndLabels:
     column_labels: Collection
     index_labels: Collection
     ordering_col: str
+    timedelta_cols: List[str]
```
TrevorBergeron, line 42: Do we truly need this?

bigframes/session/_io/pandas.py
```diff
@@ -163,9 +165,33 @@ def pandas_to_bq_compatible(pandas_dataframe: pandas.DataFrame) -> DataFrameAndL
...
     return DataFrameAndLabels(
         df=pandas_dataframe_copy,
         column_labels=col_labels,
         index_labels=idx_labels,
         ordering_col=ordering_col,
+        timedelta_cols=timedelta_cols,
     )
+
+
+def _replace_timedeltas_with_micros(dataframe: pandas.DataFrame) -> List[str]:
```
TrevorBergeron, line 179: Can we share logic between here and the compiler for ReadLocalNode

bigframes/core/utils.py
sycai: Because it only returns sub-second part:

expression `pd.Timedelta('1day2hour3ms').microseconds` evaluates to 3000

https://pandas.pydata.org/docs/dev/reference/api/pandas.Timedelta.microseconds.html#pandas-timedelta-microseconds

bigframes/core/compile/compiler.py
sycai: I guess it doesn't hurt if we support it, right?

bigframes/core/schema.py
sycai: We don't seem to have an API that is able to convert pandas dataframe directly to ArraySchema. We could define one, though I feel like it's out of the scope of this PR.

bigframes/session/_io/pandas.py
sycai: yeah.. this is for column type overrides.

bigframes/session/_io/pandas.py
sycai: Good call. Done.

bigframes/core/utils.py
TrevorBergeron: hmm, just wary of floats, would be nice if we could extra the microseconds directly

TrevorBergeron (approved):

PR #1365, opened 2025-02-05
perf: Prune unused operations from sql

TrevorBergeron:

### DIFF
--- a/bigframes/core/rewrite/pruning.py
+++ b/bigframes/core/rewrite/pruning.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import dataclasses
 import functools
-from typing import AbstractSet, Iterable, TypeVar
+from typing import AbstractSet
 
 import bigframes.core.identifiers
 import bigframes.core.nodes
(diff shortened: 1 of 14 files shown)


tswast:
I'll take a look, but one high-level thought: we should exclude operations from this that might have side effects. Specifically, remote function calls could do all sorts of things (send email, create files in GCS, etc.) without the return value being used.

TrevorBergeron:
> I'll take a look, but one high-level thought: we should exclude operations from this that might have side effects. Specifically, remote function calls could do all sorts of things (send email, create files in GCS, etc.) without the return value being used.

There is a lot that needs to be done to properly model intended side-effects. Lazy evaluation can result in side effects being repeated, while filter pushdown (client side or bigquery side) can remove calls. I think side-effects should probably be explicitly indicated through a special api (maybe with no-output consumer udfs?), and we can then model as its own node type.

bigframes/core/rewrite/pruning.py
```diff
@@ -0,0 +1,207 @@
...
+    if len(result.items) == 0:
+        # We need to select something, or stuff breaks
+        result = bigframes.core.nodes.ScanList(scanlist.items[:1])
+    return result
+
+
+T = TypeVar("T")
+
+
+def dedupe(items: Iterable[T]) -> Iterable[T]:
```
sycai, line 201: This function seems to have no caller. Remove it?

TrevorBergeron: removed

sycai (approved):

PR #1380, opened 2025-02-10
refactor: clean up bigframes/core/__init__.py

sycai:


bigframes/core/nodes.py
```diff
@@ -24,6 +24,7 @@
 
 import google.cloud.bigquery as bq
 
+from bigframes.core import identifiers as bf_ids
```
GarrettWu, line 27: Can those be consistent? "identifiers" vs "bf_ids".

Also look at line 30 there is another import bigframes.core.identifiers

GarrettWu (approved):

bigframes/core/nodes.py
sycai: Good call. Done

PR #1381, opened 2025-02-10
perf: Use simple null constraints to simplify queries

TrevorBergeron:


bigframes/core/compile/compiled.py
```diff
@@ -16,7 +16,7 @@
 import functools
 import itertools
 import typing
-from typing import Optional, Sequence
+from typing import Literal, Optional, Sequence, Tuple
```
GarrettWu, line 19: nit: if just use "tuple" for type annotation, then no need to import.

bigframes/core/compile/compiled.py
```diff
@@ -323,7 +323,97 @@ def from_pandas(
...
+            for left_index, right_index in conditions
+        ]
+
+        combined_table = bigframes_vendored.ibis.join(
+            left_table,
+            right_table,
+            predicates=join_conditions,
+            how=type,  # type: ignore
+        )
+        combined_table = bigframes_vendored.ibis.join(
```
GarrettWu, line 366: dup?

TrevorBergeron: fixed

bigframes/core/compile/compiled.py
TrevorBergeron: fixed all `Tuple` and `typing.Tuple` for file

GarrettWu (approved):

bigframes/core/compile/compiled.py
tswast: We need to be careful about this kind of feedback. `tuple` typing subscripting wasn't added until 3.9 (see: https://docs.python.org/3/library/typing.html#typing.Tuple). Since bigframes dropped 3.8 pretty early on, this is OK, but please include links to what versions of Python syntax like this is supported in in your feedback next time.

PR #1385, opened 2025-02-11
refactor: resolve identifier package names in nodes.py

sycai:


bigframes/core/nodes.py
```diff
@@ -887,7 +887,7 @@ def from_table(table: bq.Table, columns: Sequence[str] = ()) -> GbqTable:
             dataset_id=table.dataset_id,
             table_id=table.table_id,
             physical_schema=schema,
-            n_rows=table.num_rows,
+            n_rows=table.num_rows or 0,
```
GarrettWu, line 890: Is it redundant? "table.num_rows or 0" equals to table.num_rows

GarrettWu (approved):

PR #1386, opened 2025-02-11
refactor: Tree traversals now non-recursive

TrevorBergeron:

### DIFF
--- a/bigframes/core/plan.py
+++ b/bigframes/core/plan.py
@@ -1,0 +1,369 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import abc
+import collections
+import dataclasses
+import functools
+import itertools
+import typing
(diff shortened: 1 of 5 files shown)


bigframes/core/plan.py
```diff
@@ -0,0 +1,369 @@
...
+import abc
+import collections
+import dataclasses
+import functools
+import itertools
+import typing
+from typing import Callable, Dict, Generator, Iterable, Mapping, Set, Tuple
+
+import bigframes.core.guid
+import bigframes.core.identifiers
```
sycai, line 26: Here we have two imports for the same module.

Could you use `from bigframes.core import identifiers` and resolve all module references? go/pydev#imports

bigframes/core/plan.py
```diff
@@ -0,0 +1,369 @@
...
+                stack.extend(item.child_nodes)
+
+    def edges(
+        self: BigFrameNode,
+    ) -> Generator[Tuple[BigFrameNode, BigFrameNode], None, None]:
+        for item in self.unique_nodes():
+            for child in item.child_nodes:
+                yield (item, child)
+
+    def topo(self: BigFrameNode) -> Generator[BigFrameNode, None, None]:
```
sycai, line 303: name suggestion: `topo_ordered_nodes` or `get_nodes_in_topo_order` or something similar to reflect the fact that we are returning a collection of nodes.

bigframes/core/plan.py
sycai, line 1: Let's update the filename with anything that accurately describes the content, and preferably ends with "node"

bigframes/core/plan.py
TrevorBergeron: renamed to `iter_nodes_topo`

bigframes/core/plan.py
TrevorBergeron: ok, yeah, that is nicer, done

bigframes/core/plan.py
TrevorBergeron: renamed to bigframe_node.py

sycai (approved):

PR #1391, opened 2025-02-12
feat: JSON dtype support for read_pandas and Series constructor

chelsea-lin:
- [X] Fixes internal issue 395912450 
- [] Appropriate docs were updated (if necessary)

Fixes internal issue 395912450 🦕

### DIFF
--- a/bigframes/core/compile/compiler.py
+++ b/bigframes/core/compile/compiler.py
@@ -20,6 +20,7 @@ import typing
 
 import bigframes_vendored.ibis.backends.bigquery as ibis_bigquery
 import bigframes_vendored.ibis.expr.api as ibis_api
+import bigframes_vendored.ibis.expr.datatypes as ibis_dtypes
 import bigframes_vendored.ibis.expr.types as ibis_types
 import google.cloud.bigquery
 import pandas as pd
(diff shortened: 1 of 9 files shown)


bigframes/session/_io/pandas.py
```diff
@@ -165,11 +166,13 @@ def pandas_to_bq_compatible(pandas_dataframe: pandas.DataFrame) -> DataFrameAndL
...
     timedelta_cols = utils.replace_timedeltas_with_micros(pandas_dataframe_copy)
+    json_cols = utils.replace_json_with_string(pandas_dataframe_copy)
 
     return DataFrameAndLabels(
         df=pandas_dataframe_copy,
         column_labels=col_labels,
         index_labels=idx_labels,
         ordering_col=ordering_col,
         timedelta_cols=timedelta_cols,
+        json_cols=json_cols,
```
sycai, line 177: Shall we combine `timedelta_cols` and `json_cols` into a single field `col_type_override` for DataFrameAndLabels?

bigframes/core/compile/compiler.py
```diff
@@ -39,6 +39,7 @@
 import bigframes.core.nodes as nodes
 import bigframes.core.ordering as bf_ordering
 import bigframes.core.rewrite as rewrites
+import bigframes.dtypes
```
sycai, line 42: Another nit: let's follow Python readability guidance (go/pydev#imports) and use

`from bigframes import dtypes` instead.

bigframes/session/_io/pandas.py
chelsea-lin: Good points to make code cleaner.

bigframes/core/compile/compiler.py
chelsea-lin: Addressed. Thanks!

sycai (approved):

PR #1403, opened 2025-02-18
feat: Support DF.__array__(copy=True)

TrevorBergeron:


third_party/bigframes_vendored/pandas/core/frame.py
```diff
@@ -7179,7 +7179,7 @@ def __len__(self):
         """
         raise NotImplementedError(constants.ABSTRACT_METHOD_ERROR_MESSAGE)
 
-    def __array__(self):
+    def __array__(self, dtype=None, copy=None):
```
jialuoo, line 7182: If copy can be (bool or None, optional), better to use 'copy: Optional[bool] = None' too.

Same for series.

TrevorBergeron: added annotation

jialuoo (approved):

PR #1411, opened 2025-02-19
fix: calling to_timdelta() over timedeltas no longer changes their values

sycai:


tests/system/small/test_pandas.py
```diff
@@ -829,3 +829,18 @@ def test_to_timedelta_with_bf_series_invalid_unit(session, unit):
 @pytest.mark.parametrize("input", [1, 1.2, "1s"])
 def test_to_timedelta_non_bf_series(input):
     assert bpd.to_timedelta(input) == pd.to_timedelta(input)
+
+
+def test_to_timedelta_on_timedelta_series__should_be_no_op(scalars_dfs):
+    bf_df, pd_df = scalars_dfs
+    bf_series = bpd.to_timedelta(bf_df["int64_too"], unit="us")
+    pd_series = pd.to_timedelta(pd_df["int64_too"], unit="us")
```
chelsea-lin, line 837: nit: trying the float column too?

chelsea-lin (approved):

tests/system/small/test_pandas.py
sycai: SG!

PR #1419, opened 2025-02-21
feat: warn if default ingress_settings is used in remote_functions

chelsea-lin:
This change modifies the default ingress_settings to a sentinel object (None). While this initially defaults to 'all', a warning is now issued to users, indicating that the default will change and 'all' is less secure than 'internal-only'.

- [X] Fixes internal issue 397677177

Fixes internal issue 397677177🦕

### DIFF
--- a/bigframes/functions/_function_session.py
+++ b/bigframes/functions/_function_session.py
@@ -405,7 +405,8 @@ class FunctionSession:
             cloud_function_ingress_settings = "all"
             msg = (
                 "The `cloud_function_ingress_settings` are set to 'all' by default, "
-                "which may change. Consider using 'internal-only' for enhanced security. "
+                "which will change to 'internal-only' for enhanced security in future version 2.0 onwards. "
+                "However, you will be able to explicitly pass cloud_function_ingress_settings='all' if you need. "
                 "See https://cloud.google.com/functions/docs/networking/network-settings#ingress_settings for details."
             )
             warnings.warn(msg, category=FutureWarning, stacklevel=2)
--- a/bigframes/functions/_function_session.py
+++ b/bigframes/functions/_function_session.py
@@ -120,9 +120,9 @@ class FunctionSession:
         cloud_function_max_instances: Optional[int] = None,
         cloud_function_vpc_connector: Optional[str] = None,
         cloud_function_memory_mib: Optional[int] = 1024,
-        cloud_function_ingress_settings: Literal[
-            "all", "internal-only", "internal-and-gclb"
-        ] = "all",
+        cloud_function_ingress_settings: Optional[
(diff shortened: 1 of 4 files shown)


bigframes/functions/_function_session.py
```diff
@@ -400,6 +401,15 @@ def remote_function(
...
             )
 
+        if cloud_function_ingress_settings is None:
+            cloud_function_ingress_settings = "all"
+            msg = (
+                "The `cloud_function_ingress_settings` are set to 'all' by default, "
+                "which may change. Consider using 'internal-only' for enhanced security. "
+                "See https://cloud.google.com/functions/docs/networking/network-settings#ingress_settings for details."
+            )
+            warnings.warn(msg, category=UserWarning)
```
tswast, line 411: Let's use [`FutureWarning`](https://docs.python.org/3/library/exceptions.html#FutureWarning) for this, since it's for a breaking change coming in the future. Also, the `stacklevel=2` is better than the default for helping users find the problem code using this (https://github.com/python/cpython/issues/88462#issuecomment-1093917557).

```suggestion
            warnings.warn(msg, category=FutureWarning, stacklevel=2)
```

tests/system/large/functions/test_remote_function.py
tswast, line 1: Could we please add a test that makes sure the warning is raised, possibly a unit test since we don't really need to deploy a function for such a case?

bigframes/functions/_function_session.py
tswast: Also, TIL warn has added an option to be better than stacklevel which we can use in a few years when we drop Python 3.11 support:

> Changed in version 3.12: Added `skip_file_prefixes`.

bigframes/functions/_function_session.py
```diff
@@ -400,6 +401,15 @@ def remote_function(
                 " For more details see https://cloud.google.com/functions/docs/securing/cmek#before_you_begin"
             )
 
+        if cloud_function_ingress_settings is None:
+            cloud_function_ingress_settings = "all"
+            msg = (
+                "The `cloud_function_ingress_settings` are set to 'all' by default, "
+                "which may change. Consider using 'internal-only' for enhanced security. "
```
shobsi, line 408:
```suggestion
                "which will change to 'internal-only' for enhanced security in future version 2.0 onwards. "
                "However, you will be able to explicitly pass cloud_function_ingress_settings='all' if you need. "
```

tests/system/large/functions/test_remote_function.py
chelsea-lin: Good points. Added.

tswast (commented):
Thanks!

tswast (approved):

PR #1422, opened 2025-02-24
chore: add experimental blob tansform func tuning params

GarrettWu:
Verified in test env screen/BJL3xftb3nx39jk

modify default of max_batching_rows to align with Python UDF.

### DIFF
--- a/bigframes/operations/blob.py
+++ b/bigframes/operations/blob.py
@@ -278,7 +278,9 @@ class BlobAccessor(base.SeriesMethods):
         *,
         dst: Optional[Union[str, bigframes.series.Series]] = None,
         connection: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


bigframes/operations/blob.py
```diff
@@ -278,7 +278,9 @@ def image_blur(
         *,
         dst: Optional[Union[str, bigframes.series.Series]] = None,
         connection: Optional[str] = None,
-        max_batching_rows: int = 10000,
+        max_batching_rows: int = 8096,
```
Genesis929, line 281: Quick question, why change this to 8096?

Genesis929 (approved):

shuoweil (approved):

PR #1428, opened 2025-02-25
feat: add allow_large_results option

Genesis929:
Modified behavior:
1. New global option: bigframes.options.bigquery.allow_large_results
2. to_gbq keep the same way of creating temp table.
3. Anything to gcs(to_csv, to_json, to_parquet) always use explicit destination table.
4. Other IO methods and to_csv, to_json, to_parquet when save locally will use explicit destination table when bigframes.options.bigquery.allow_large_results is true.
5. The bigframes.options.bigquery.allow_large_results option will be override by local input allow_large_results.
6. If allow_large_results=True, we will read the table size(logical size, as we can't access physical size), and warn user that in bigframes 2.0, result > 10g may cause issue and they will need to set allow_large_results=True.


bigframes/session/executor.py
```diff
@@ -243,11 +244,13 @@ def execute(
         *,
         ordered: bool = True,
         col_id_overrides: Mapping[str, str] = {},
-        use_explicit_destination: bool = False,
+        use_explicit_destination: Optional[bool] = False,
```
TrevorBergeron, line 247: should probably add a check, that only one of ordered, use_explicit_destination is allowed at a time

bigframes/session/executor.py
```diff
@@ -333,11 +336,13 @@ def export_gcs(
         uri: str,
         format: Literal["json", "csv", "parquet"],
         export_options: Mapping[str, Union[bool, str]],
+        allow_large_results: Optional[bool] = None,
```
TrevorBergeron, line 339: I think export jobs should just assume a large result. Can't remember why we don't just export as a single job anyways?

bigframes/session/executor.py
Genesis929: Oh, forget about that conflict, thanks!

bigframes/session/executor.py
Genesis929: Updated, now it's always use a destination table.

bigframes/dataframe.py
```diff
@@ -3794,29 +3822,45 @@ def to_dict(
...
 
     def to_latex(
         self,
         buf=None,
         columns: Sequence | None = None,
         header: bool | Sequence[str] = True,
         index: bool = True,
+        allow_large_results=None,
         **kwargs,
     ) -> str | None:
```
tswast, line 3849: Aside: I'm a bit surprised this works. I guess we are type checking on Python 3.10 where this syntax was added (https://peps.python.org/pep-0604/), not 3.9?

bigframes/series.py
```diff
@@ -1780,26 +1798,40 @@ def to_json(
...
             )
 
     def to_latex(
-        self, buf=None, columns=None, header=True, index=True, **kwargs
+        self,
+        buf=None,
+        columns=None,
+        header=True,
+        index=True,
+        allow_large_results=None,
```
tswast, line 1823: Nit: we don't want people to access `allow_large_results` positionally.

```suggestion
        *,
        allow_large_results=None,
```

bigframes/series.py
```diff
@@ -1780,26 +1798,40 @@ def to_json(
...
-        return self.to_pandas().to_latex(
+        return self.to_pandas(allow_large_results=allow_large_results).to_latex(
             buf, columns=columns, header=header, index=index, **kwargs
         )
 
-    def tolist(self) -> _list:
-        return self.to_pandas().to_list()
+    def tolist(
+        self,
+        allow_large_results: Optional[bool] = None,
```
tswast, line 1832: Nit: we don't want people to access allow_large_results positionally.

```suggestion
        *,
        allow_large_results: Optional[bool] = None,
```

bigframes/series.py
```diff
@@ -1809,14 +1841,17 @@ def to_markdown(
         buf: typing.IO[str] | None = None,
         mode: str = "wt",
         index: bool = True,
+        allow_large_results=None,
```
tswast, line 1844: Nit: we don't want people to access allow_large_results positionally.

```suggestion
        *,
        allow_large_results=None,
```

bigframes/core/indexes/base.py
```diff
@@ -490,17 +490,28 @@ def __getitem__(self, key: int) -> typing.Any:
         else:
             raise NotImplementedError(f"Index key not supported {key}")
 
-    def to_pandas(self) -> pandas.Index:
+    def to_pandas(self, allow_large_results: Optional[bool] = None) -> pandas.Index:
```
tswast, line 493: It's less necessary in this context, since we aren't trying to mimic pandas, but I'd still like to avoid using this parameter positionally.

```suggestion
    def to_pandas(self, *, allow_large_results: Optional[bool] = None) -> pandas.Index:
```

bigframes/core/indexes/base.py
```diff
@@ -490,17 +490,28 @@ def __getitem__(self, key: int) -> typing.Any:
...
-        return self._block.index.to_pandas(ordered=True)
+        df, query_job = self._block.index.to_pandas(
+            ordered=True, allow_large_results=allow_large_results
+        )
+        self._query_job = query_job
+        return df
 
-    def to_numpy(self, dtype=None, **kwargs) -> np.ndarray:
-        return self.to_pandas().to_numpy(dtype, **kwargs)
+    def to_numpy(self, dtype=None, allow_large_results=None, **kwargs) -> np.ndarray:
```
tswast, line 511: In this case, we are trying to mimic pandas (https://pandas.pydata.org/pandas-docs/version/2.1.2/reference/api/pandas.Index.to_numpy.html), so it is very important to restrict use positionally.

```suggestion
    def to_numpy(self, dtype=None, *, allow_large_results=None, **kwargs) -> np.ndarray:
```

Otherwise, someone might have some pandas code that does `index.to_numpy("int64", True)` where in pandas that "True" means `copy=True`, but here it means something else.

tswast (changes requested):
One nit (repeated throughout): I'd like to make sure we only expose `allow_large_results` as a keyword argument, not allowing positional access. That'll make sure we prevent breakages if folks are coming from pandas.

bigframes/dataframe.py
Genesis929: Updated.

bigframes/series.py
Genesis929: Updated.

bigframes/series.py
Genesis929: Updated.

bigframes/core/indexes/base.py
Genesis929: Updated.

bigframes/core/indexes/base.py
Genesis929: Updated.

tswast (approved):
THanks!

PR #1429, opened 2025-02-25
feat: (Preview) Support automatic load of timedelta from BQ tables.

sycai:
This works only with `read_gbq_table()` under the hood, because `read_gbq_query()` erases column descriptions.


bigframes/dtypes.py
```diff
@@ -876,3 +882,6 @@ def lcd_type_or_throw(dtype1: Dtype, dtype2: Dtype) -> Dtype:
     "STRING",
     "ARRAY",
 }
+
+
+TIMEDELTA_DESCRIPTION_TAG = "#microsecond"
```
tswast, line 887: Very much a nit, but I think "microseconds" sounds better for a description of timedelta.

```suggestion
TIMEDELTA_DESCRIPTION_TAG = "#microseconds"
```

Alternatively, do we want to align with the JSON format used for remote functions descriptions? I think there are benefits to this implementation, but it'd probably be good to document that choice in the design if we haven't already.

bigframes/session/executor.py
```diff
@@ -323,8 +324,51 @@ def export_gbq(
...
+        table_ref: bigquery.TableReference | None,
+        array_value: bigframes.core.ArrayValue,
+        col_id_overrides: Mapping[str, str],
+    ):
+        """Updates the descriptions of columns that are holding timedeltas values with a special tag."""
+        assert table_ref is not None
+
+        table = self.bqclient.get_table(table_ref)
+        table_def = table.to_api_repr()
+        for field in table_def["schema"]["fields"]:
```
tswast, line 343: Do we need to be recursing into the sub fields if this is a RECORD / STRUCT so we can label those, too?

Or maybe leave that as a TODO?

tswast: Also, I'm wondering about the cost of this loop. Some tables might have 10,000 columns. This will likely be pretty slow in Python.

sycai: RECORD columns are complicated in the sense that it can contain both timedelta and integer columns, so our tag needs to have some kind of recursive structures too. Shall we wait for FRs from customers before evaluating the required effort?

Good call for the loop comment. I didn't realize it's a nested loop. Updated the code such that timedelta column labels are passed in as a set, and the lookup of each field costs O(1)

bigframes/dtypes.py
sycai: Could you share more about this JSON format description?

bigframes/session/executor.py
```diff
@@ -299,6 +302,7 @@ def export_gbq(
         destination: bigquery.TableReference,
         if_exists: Literal["fail", "replace", "append"] = "fail",
         cluster_cols: Sequence[str] = [],
+        timedelta_col_labels: Set[str] | None = None,
```
TrevorBergeron, line 305: The executor can figure out which columns are timedelta columns from the tree itself, lets remove this argument.

bigframes/dataframe.py
```diff
@@ -3709,12 +3709,25 @@ def to_gbq(
...
+        # they are represented as integers. Without this tag, we cannot tell a timedelta
+        # column from an integer column
+        timedelta_col_labels = set()
+        for label in self.columns:
+            col = self[label]
+            if not isinstance(col, bigframes.series.Series):
+                # Column name is duplicated, so ignore
+                continue
+            if col.dtype == bigframes.dtypes.TIMEDELTA_DTYPE:
+                timedelta_col_labels.add(label)
```
TrevorBergeron, line 3722: not needed

TrevorBergeron (changes requested):

bigframes/session/executor.py
sycai: The problem is that the materialized table uses labels as column names, while the tree node has only column ID. Therefore, we cannot rely on the tree alone to update the column descriptions. 

The `col_id_overrides` can be used as a lookup table, though, but it is a map from `col_id` to `col_label` , so we need to traverse the node schema for each table column field update, which could be slow.

TrevorBergeron: Hmm, we should probably actually push the labels into the tree first using the `override_ids` helper at the bottom of the file. Missed this in my recent refactor. Also, traversing output schema is an unavoidable part of exporting data, we aren't going to get sublinear scaling for that.

sycai:
Closing the PR for now to think about annotations for RECORD and ARRAY. I will prepare another PR once everything falls into place

bigframes/dataframe.py
sycai: Code deleted

bigframes/session/executor.py
sycai: Parameter removed.

bigframes/session/executor.py
```diff
@@ -320,8 +321,30 @@ def export_gbq(
             sql=sql,
             job_config=job_config,
         )
+        self._attach_timedelta_metadata(query_job.destination, array_value)
         return query_job
 
+    def _attach_timedelta_metadata(
```
TrevorBergeron, line 327: why not put this logic instead in `convert_to_schema_field`? it is the natural inverse of `convert_schema_field`

bigframes/dtypes.py
```diff
@@ -678,6 +678,12 @@ def convert_schema_field(
         pa_struct = pa.struct(fields)
         pa_type = pa.list_(pa_struct) if is_repeated else pa_struct
         return field.name, pd.ArrowDtype(pa_type)
+    elif (
+        field.field_type == "INTEGER"
+        and field.description is not None
+        and TIMEDELTA_DESCRIPTION_TAG in field.description
```
TrevorBergeron, line 684: nit: but I do wonder if once in a while, we might catch a user description that accidentally uses '#microseconds' somewhere? maybe restrict to beginning, or put some odds character sequence to ensure non-accidental triggering

bigframes/session/executor.py
sycai: Good call! That makes the code cleaner.

sycai: Though interestingly, in `export_gbq` we rely on the BigQuery engine to figure out the schema based on the provided SQL. I need to update the schema again afterwards to include the tag

bigframes/dtypes.py
sycai: SG. Changed the logic to check whether the description ends with the tag.

bigframes/session/executor.py
```diff
@@ -320,6 +320,14 @@ def export_gbq(
             sql=sql,
             job_config=job_config,
         )
+
+        if if_exists != "append":
```
TrevorBergeron, line 324: In the interest of saving 3 microseconds of latency, are we able to validate locally whether the update is needed? eg by comparing destination schema to intended schema?

TrevorBergeron (approved):

bigframes/session/executor.py
sycai: Hah sure. It should be fairly straightforward

PR #1432, opened 2025-02-27
chore: show `FutureWarning` for LEP endpoints usage with  `use_regional_endpoints=True`

shobsi:
```python
>>> import bigframes.pandas as bpd
>>> bpd.options.bigquery.use_regional_endpoints = True
<stdin>:1: PreviewWarning: Use of regional endpoints is a feature in preview and available only in selected regions and projects.
>>> bpd.options.bigquery.location = "europe-west1"
>>> s = bpd.get_global_session()
.../bigframes/bigframes/session/clients.py:106: FutureWarning: Support for regional endpoints is not yet available in the location
europe-west1 for BigQuery and BigQuery Storage APIs. For the supported
locations and APIs see https://cloud.google.com/bigquery/docs/regional-endpoints.
For other locations and APIs, currently an older, now deprecated locational
endpoints are being used, which requires your project to be allowlisted. In
future version 2.0 onwards the locational endpoints will no longer be
supported automatically when you enable regional endpoints. However, if you
still need them, you will be able to override the endpoints directly by
doing the following:
    bigframes.pandas.options.bigquery.client_endpoints_override = {
        "bqclient": "https://europe-west1-bigquery.googleapis.com",
        "bqconnectionclient": "europe-west1-bigqueryconnection.googleapis.com",
        "bqstoragereadclient": "europe-west1-bigquerystorage.googleapis.com"
    }
  warnings.warn(
Query job ... is DONE.0 Bytes processed.
https://console.cloud.google.com/bigquery?project=...
>>> 
```

Fixes internal issue 399101005 🦕


tests/system/large/test_location.py
```diff
@@ -118,12 +120,28 @@ def test_bq_location_non_canonical(set_location, resolved_location):
     sorted(bigframes.constants.REP_ENABLED_BIGQUERY_LOCATIONS),
 )
 def test_bq_rep_endpoints(bigquery_location):
-    session = bigframes.Session(
-        context=bigframes.BigQueryOptions(
-            location=bigquery_location, use_regional_endpoints=True
+    with warnings.catch_warnings(record=True) as record:
```
tswast, line 123: I recommend `pytest.warns` instead (https://docs.pytest.org/en/stable/reference/reference.html#pytest.warns). It takes a `match` argument that lets use check that LEP / REP was mentioned, to avoid false negatives.

bigframes/session/clients.py
```diff
@@ -94,8 +96,22 @@ def __init__(
...
+            use_regional_endpoints
+            and location is not None
+            and location.lower()
+            not in bigframes.constants.REP_ENABLED_BIGQUERY_LOCATIONS
+        ):
+            warnings.warn(
+                bigframes.constants.LEP_DEPRECATION_WARNING_MESSAGE.format(
+                    location=location
+                ),
+                category=bigframes.exceptions.DeprecationWarning,
```
tswast, line 110: Let's use the built-in `FutureWarning`.

```suggestion
                category=FutureWarning,
```

bigframes/session/clients.py
```diff
@@ -117,20 +133,22 @@ def __init__(
...
-                    and self._location.lower()
-                    in bigframes.constants.REP_ENABLED_BIGQUERY_LOCATIONS
-                    else _BIGQUERY_LOCATIONAL_ENDPOINT
-                ).format(location=self._location),
-            )
         if "bqclient" in self._client_endpoints_override:
             bq_options = google.api_core.client_options.ClientOptions(
                 api_endpoint=self._client_endpoints_override["bqclient"]
             )
+        elif self._use_regional_endpoints:
```
tswast, line 140: By creating an `elif` and preferring `_client_endpoints_override`, I think we are making some user-visible changes. I don't think "chore" is the right way to classify this issue. Perhaps a fix.

tests/system/large/test_location.py
shobsi: We are using `pytest.warns` in `test_bq_lep_endpoints` (line#168) where the LEP warning _is_ raised. Here we are verifying that that LEP warning _is not_ raised - by capturing all warnings and asserting that none of them is LEP warning.

`pytest.warns(None)` could have worked, but it is deprecated in pytest 7 and unsupported in pytest 8 https://docs.pytest.org/en/7.1.x/deprecations.html#using-pytest-warns-none, so decided to not use that.

Let me know if that makes sense.

bigframes/session/clients.py
shobsi: Done, thanks!

bigframes/session/clients.py
shobsi: Actually it is more for readability, achieving the same end result with one less check.

we are just rewriting

```python
if cond1:
    options = option1
if cond2:
    options = option2
```

to 

```python
if cond2:
    options = option2
elif cond1:
    options = option1
```

bigframes/exceptions.py
```diff
@@ -81,6 +81,10 @@ class UnknownDataTypeWarning(Warning):
     """Data type is unknown."""
 
 
+class DeprecationWarning(Warning):
+    """The functionality or support is being deprecated."""
+
+
```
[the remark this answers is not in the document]
shobsi: Removed, thanks!

tests/system/large/test_location.py
tswast: Ack thanks. I missed that.

bigframes/session/clients.py
tswast: You're right. If cond1 and cond2 are both true, both of these end up with option2. I think the `elif` version makes that clearer.

tswast (approved):

PR #1436, opened 2025-02-27
feat: Support dry_run in `to_pandas()`

sycai:


tswast:
Heads up: I suspect we might encounter some conflicts after https://github.com/googleapis/python-bigquery-dataframes/pull/1448/ merges. I think that `allow_large_results` change is more urgent, so let's refrain from merging this until that's in.

sycai:
I used a DataFrame for dry run stats, just like what `describe()` does. 

It looks like this: https://screenshot.googleplex.com/gEEBU7aMqUy93SB

Plus, I applied some "@overload" magic to make `mypy` happy.

Let me know your thoughts @tswast

tswast:
> I used a DataFrame for dry run stats, just like what describe() does.

I think that's a bit confusing too, since only one column will ever have a value.

tswast:
Here's what I had in mind:

```python
import copy

import pandas as pd
from google.colab import auth
from google.cloud import bigquery

# Authenticate the user
auth.authenticate_user()

# Initialize a BigQuery client
client = bigquery.Client(project='bigframes-dev') # Replace with your project ID

job_config.dry_run = True

query = """
SELECT
  name,
  SUM(number) AS total
FROM
  `bigquery-public-data.usa_names.usa_1910_2013`
GROUP BY
  name
ORDER BY
  total DESC
LIMIT
  10;
"""

job = client.query(query, job_config=job_config)

job_api_repr = copy.deepcopy(job._properties)
print(job_api_repr)
index = []
values = []
query_configuration = job_api_repr['configuration'].pop("query")
index.extend(query_configuration.keys())
values.extend(query_configuration.values())

query_statistics = job_api_repr['statistics'].pop("query")
index.extend(query_statistics.keys())
values.extend(query_statistics.values())

remaining_statistics = job_api_repr['statistics']
index.extend(remaining_statistics.keys())
values.extend(remaining_statistics.values())

series = pd.Series(values, index=index)
print(series)
series.to_frame()
```

tswast:
Re: https://github.com/googleapis/python-bigquery-dataframes/pull/1436#issuecomment-2704399530

It would need

* some careful checks in case some of those keys don't exist,
* maybe we want an allowlist of keys so we only expose the job stats we've validated are useful to a bigframes user,
* add dtypes and index_dtypes for the expected output types (possibly remove `schema`).

sycai:
Now the stats look like these: https://screenshot.googleplex.com/BhaUb8uTYrdz7iM

I hard-coded all the keys for value lookup in the dry run job.

bigframes/core/blocks.py
```diff
@@ -2691,20 +2779,49 @@ def column_ids(self) -> Sequence[str]:
...
         ordered = ordered if ordered is not None else True
+        if dry_run:
+            series, query_job = self._block.select_columns([]).to_pandas(
+                ordered=ordered,
+                allow_large_results=allow_large_results,
+                dry_run=dry_run,
+            )
+            return series, query_job
+
         df, query_job = self._block.select_columns([]).to_pandas(
```
tswast, line 2823: The `select_columns([])` confuses me, but I see that was here before. Please refactor these a bit so that `self._block.select_columns([])` is saved to a variable since it is in common with both.

bigframes/core/blocks.py
```diff
@@ -2691,20 +2779,49 @@ def column_ids(self) -> Sequence[str]:
...
         """Executes deferred operations and downloads the results."""
         if len(self.column_ids) == 0:
             raise bigframes.exceptions.NullIndexError(
                 "Cannot materialize index, as this object does not have an index. Set index column(s) using set_index."
             )
         ordered = ordered if ordered is not None else True
+        if dry_run:
+            series, query_job = self._block.select_columns([]).to_pandas(
+                ordered=ordered,
+                allow_large_results=allow_large_results,
```
tswast, line 2818: I think `allow_large_results` shouldn't have an effect on dry run queries, as that controls the destination table property.

bigframes/core/blocks.py
```diff
@@ -2691,20 +2779,49 @@ def column_ids(self) -> Sequence[str]:
...
+                dry_run=dry_run,
+            )
+            return series, query_job
+
         df, query_job = self._block.select_columns([]).to_pandas(
-            ordered=ordered, allow_large_results=allow_large_results
+            ordered=ordered, allow_large_results=allow_large_results, dry_run=dry_run
```
tswast, line 2824: Why include the `dry_run` argument here if we know it's false?

tswast, line 2824: Why include the `dry_run` argument here if we know it's false?

bigframes/core/indexes/base.py
```diff
@@ -490,20 +492,51 @@ def __getitem__(self, key: int) -> typing.Any:
...
         """
+        if dry_run:
+            series, query_job = self._block.index.to_pandas(
+                ordered=True, allow_large_results=allow_large_results, dry_run=dry_run
+            )
+            if query_job:
+                self._query_job = query_job
+            return series
+
+        # Repeat the to_pandas() call to make mypy deduce type correctly, because mypy cannot resolve
```
tswast, line 536: Why don't you just use `bool` consistently, then?

bigframes/core/indexes/base.py
```diff
@@ -490,20 +492,51 @@ def __getitem__(self, key: int) -> typing.Any:
...
+                a Pandas series containing dtype and the amount of bytes to be processed.
 
         Returns:
-            pandas.Index:
-                A pandas Index with all of the labels from this Index.
+            pandas.Index | pandas.Series:
+                A pandas Index with all of the labels from this Index. If dry run is set to True,
+                returns a Series containing dry run statistics.
         """
+        if dry_run:
```
tswast, line 528: This feels really redundant. Why not just name the variable `df_or_series`? The code is identical.

bigframes/core/blocks.py
tswast: Alternatively, we can get rid of this `if` statement and rename the variable `df_or_series`.

tswast (commented):
Please find another way to make the type checker happy. I dislike the repeated code.

sycai:
> Please find another way to make the type checker happy. I dislike the repeated code.

Yeah that's true...  I moved the `if dry_run then block._compute_dry_run` code from the `block.to_pandas()` to `DataFrames/Series/Index.to_pandas()`, that should clean up the code a bit. Plus, no more signature override to `block.to_pandas()`

tswast:
Presubmits failed:

```
FAILED tests/system/small/test_dataframe.py::test_df_peek[partial] - assert (...
FAILED tests/system/small/test_dataframe.py::test_df_peek[strict] - assert (6...
```

Hard to say if it's related to this PR.

tswast (approved):

PR #1438, opened 2025-02-27
refactor: Remove col_id_overrides from executor interfaces

TrevorBergeron:

### DIFF
--- a/bigframes/core/array_value.py
+++ b/bigframes/core/array_value.py
@@ -349,7 +349,7 @@ class ArrayValue:
             )
         )
 
-    def rename(self, col_id_overrides: Mapping[str, str]) -> ArrayValue:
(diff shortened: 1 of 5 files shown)


bigframes/core/array_value.py
```diff
@@ -349,6 +349,20 @@ def select_columns(self, column_ids: typing.Sequence[str]) -> ArrayValue:
             )
         )
 
+    def rename(self, col_id_overrides: Mapping[str, str]) -> ArrayValue:
```
sycai, line 352: nit: maybe call it "rename_col_ids" or something similar?

sycai (approved):
Thanks!

bigframes/core/array_value.py
TrevorBergeron: renamed rename to rename_columns

PR #1439, opened 2025-02-28
chore: impl experimental blob image output encodings

GarrettWu:
Verified in test env screen/VhtjnVggxkmjAYb

now image transform output encoding depends on the output(input) file extensions.

### DIFF
--- a/bigframes/blob/_functions.py
+++ b/bigframes/blob/_functions.py
@@ -105,13 +105,15 @@ AS r\"\"\"
 
 # Blur images. Takes ObjectRefRuntime as JSON string. Outputs ObjectRefRuntime JSON string.
 def image_blur_func(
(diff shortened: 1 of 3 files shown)


Genesis929 (approved):

bigframes/blob/_functions.py
```diff
@@ -395,7 +434,7 @@ def pdf_chunk_func(src_obj_ref_rt: str, chunk_size: int, overlap_size: int) -> s
     if curr_chunk:
         all_text_chunks.append(curr_chunk)
 
-    all_text_json_string = json.dumps(all_text_chunks)
+    all_text_json_string = json.dumps(["123"])
```
shuoweil, line 437: It will write dummy text instead of chunked results, please fix it

shuoweil (commented):
Please fix the bug

PR #1444, opened 2025-03-01
chore: write chunked text instead of dummy text for pdf chunk

shuoweil:
write chunked text instead of dummy text for pdf chunk, verified in test env: screen/7YbCgKQH8eWLNHr


GarrettWu:
Don't use "fix" title, it'll appear in release note. "chore" instead.

GarrettWu (approved):

PR #1448, opened 2025-03-04
feat: add allow_large_results to peek

Genesis929:
1. Added allow_large_results to peek
2. Switched allow_large_results to query_and_wait.
3. Added warning messages to metrics and allow_large_results option to warn user that we won't have metrics and won't sampling.
4. skip sampling when allow_large_results=False as we can't get the table size.

### DIFF
--- a/bigframes/_config/bigquery_options.py
+++ b/bigframes/_config/bigquery_options.py
@@ -35,6 +35,12 @@ SESSION_STARTED_MESSAGE = (
 
 UNKNOWN_LOCATION_MESSAGE = "The location '{location}' is set to an unknown value. Did you mean '{possibility}'?"
 
+LARGE_RESULTS_WARNING_MESSAGE = (
+    "Sampling is disabled because 'allow_large_results' is set to False. "
+    "To prevent downloading excessive data, it is recommended to limit the data "
+    "fetched with methods like .head() or .sample() before proceeding with downloads."
+)
+
 
 def _get_validated_location(value: Optional[str]) -> Optional[str]:
 
@@ -99,6 +105,9 @@ class BigQueryOptions:
         self._application_name = application_name
         self._kms_key_name = kms_key_name
         self._skip_bq_connection_check = skip_bq_connection_check
+
+        if not allow_large_results:
+            warnings.warn(LARGE_RESULTS_WARNING_MESSAGE, UserWarning)
         self._allow_large_results = allow_large_results
         self._session_started = False
         # Determines the ordering strictness for the session.
@@ -252,6 +261,8 @@ class BigQueryOptions:
 
     @allow_large_results.setter
     def allow_large_results(self, value: bool):
+        if not value:
+            warnings.warn(LARGE_RESULTS_WARNING_MESSAGE, UserWarning)
         self._allow_large_results = value
 
     @property
--- a/bigframes/_config/bigquery_options.py
+++ b/bigframes/_config/bigquery_options.py
@@ -35,12 +35,6 @@ SESSION_STARTED_MESSAGE = (
 
 UNKNOWN_LOCATION_MESSAGE = "The location '{location}' is set to an unknown value. Did you mean '{possibility}'?"
 
-LARGE_RESULTS_WARNING_MESSAGE = (
-    "Sampling is disabled because 'allow_large_results' is set to False. "
-    "To prevent downloading excessive data, it is recommended to limit the data "
-    "fetched with methods like .head() or .sample() before proceeding with downloads."
-)
-
 
 def _get_validated_location(value: Optional[str]) -> Optional[str]:
 
@@ -105,9 +99,6 @@ class BigQueryOptions:
         self._application_name = application_name
         self._kms_key_name = kms_key_name
         self._skip_bq_connection_check = skip_bq_connection_check
-
-        if not allow_large_results:
-            warnings.warn(LARGE_RESULTS_WARNING_MESSAGE, UserWarning)
         self._allow_large_results = allow_large_results
         self._session_started = False
         # Determines the ordering strictness for the session.
@@ -261,8 +252,6 @@ class BigQueryOptions:
 
     @allow_large_results.setter
     def allow_large_results(self, value: bool):
-        if not value:
-            warnings.warn(LARGE_RESULTS_WARNING_MESSAGE, UserWarning)
         self._allow_large_results = value
 
     @property
--- a/bigframes/core/blocks.py
+++ b/bigframes/core/blocks.py
@@ -629,7 +629,7 @@ class Block:
         else:
             # Since we cannot acquire the table size without a query_job,
             # we skip the sampling.
-            fraction = 1
+            fraction = 2
 
         # TODO: Maybe materialize before downsampling
         # Some downsampling methods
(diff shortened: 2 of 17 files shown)


bigframes/_config/bigquery_options.py
```diff
@@ -35,6 +35,12 @@
 
 UNKNOWN_LOCATION_MESSAGE = "The location '{location}' is set to an unknown value. Did you mean '{possibility}'?"
 
+LARGE_RESULTS_WARNING_MESSAGE = (
+    "Sampling is disabled because 'allow_large_results' is set to False. "
+    "To prevent downloading excessive data, it is recommended to limit the data "
+    "fetched with methods like .head() or .sample() before proceeding with downloads."
```
tswast, line 41: Might be better to point folks at peek().

```suggestion
    "To prevent downloading excessive data, it is recommended to use the peek() method, "
    "or limit the data with methods like .head() or .sample() before proceeding with downloads."
```

bigframes/_config/bigquery_options.py
```diff
@@ -252,6 +261,8 @@ def allow_large_results(self) -> bool:
 
     @allow_large_results.setter
     def allow_large_results(self, value: bool):
+        if not value:
+            warnings.warn(LARGE_RESULTS_WARNING_MESSAGE, UserWarning)
```
tswast, line 265: I don't think the setter makes the most sense for this warning. We want to encourage people to use this, not discourage them. I'd rather see it when they call a function that initiates `query_and_wait` and sampling has been requested.

bigframes/_config/bigquery_options.py
```diff
@@ -35,6 +35,12 @@
 
 UNKNOWN_LOCATION_MESSAGE = "The location '{location}' is set to an unknown value. Did you mean '{possibility}'?"
 
+LARGE_RESULTS_WARNING_MESSAGE = (
+    "Sampling is disabled because 'allow_large_results' is set to False. "
```
tswast, line 39: Isn't the "head" style sampling still supported, just not sampling by number of bytes?

bigframes/core/blocks.py
```diff
@@ -614,17 +616,20 @@ def _materialize_local(
...
+            max_download_size = sample_config.max_download_size
+            fraction = (
+                max_download_size / table_mb
+                if (max_download_size is not None) and (table_mb != 0)
+                else 2
+            )
+        else:
+            # Since we cannot acquire the table size without a query_job,
+            # we skip the sampling.
+            fraction = 2
```
tswast, line 632: Here seems like a good place for the warning.

bigframes/session/executor.py
```diff
@@ -329,7 +329,7 @@ def export_gbq(
         if if_exists != "append" and has_timedelta_col:
             # Only update schema if this is not modifying an existing table, and the
             # new table contains timedelta columns.
-            assert query_job.destination is not None
+            assert query_job is not None and query_job.destination is not None
             table = self.bqclient.get_table(query_job.destination)
```
tswast, line 333: I'm curious why we don't use the `destination` argument for this instead of fetching from the job.

bigframes/session/executor.py
```diff
@@ -385,12 +386,24 @@ def peek(
         if not tree_properties.can_fast_peek(plan):
             msg = "Peeking this value cannot be done efficiently."
             warnings.warn(msg)
+        if use_explicit_destination is None:
+            use_explicit_destination = bigframes.options.bigquery.allow_large_results
+
+        job_config = bigquery.QueryJobConfig()
+        # Use explicit destination to avoid 10GB limit of temporary table
+        if use_explicit_destination:
+            destination_table = self.storage_manager.create_temp_table(
```
tswast, line 395: I don't think it needs to be a full table, right? We should be able to avoid the `tables.create` call. `_random_table` seems more appropriate (https://github.com/googleapis/python-bigquery-dataframes/blob/a745290896e34a0bb25ae023ab17fb04d2c5887f/bigframes/session/temp_storage.py#L70)

tests/system/small/test_dataframe_io.py
```diff
@@ -253,27 +253,21 @@ def test_to_pandas_array_struct_correct_result(session):
 
 
 def test_to_pandas_override_global_option(scalars_df_index):
-    # Direct call to_pandas uses global default setting (allow_large_results=True),
-    # table has 'bqdf' prefix.
-    scalars_df_index.to_pandas()
-    assert scalars_df_index._query_job.destination.table_id.startswith("bqdf")
```
tswast, line 259: Seems like it would still be useful to have such a test. Why are we removing this part?

tests/system/small/test_dataframe.py
```diff
@@ -662,7 +662,25 @@ def test_rename(scalars_dfs):
...
+
+def test_df_peek_with_large_results_not_allowed(scalars_dfs_maybe_ordered):
+    scalars_df, scalars_pandas_df = scalars_dfs_maybe_ordered
+
+    session = scalars_df._block.session
+    execution_count = session._metrics.execution_count
+    peek_result = scalars_df.peek(n=3, force=False, allow_large_results=False)
+
+    # The metrics won't be updated when we call query_and_wait.
+    assert session._metrics.execution_count == execution_count
```
tswast, line 683: Seems like we could still track the number of queries we execute, right? Shouldn't need a job object for that.

Genesis929: Updated metric to count execution count if query_job is None.

Tests updated.

tests/system/small/test_dataframe_io.py
Genesis929: It's removed because it's not the main thing we test, added it back.

bigframes/session/executor.py
tswast: Ah, good point.

bigframes/session/__init__.py
```diff
@@ -344,11 +344,25 @@ def _project(self):
     @property
     def bytes_processed_sum(self):
         """The sum of all bytes processed by bigquery jobs using this session."""
+        warnings.warn(
+            "Queries executed with `allow_large_results` within the session will not "
```
tswast, line 348:
```suggestion
            "Queries executed with `allow_large_results=False` within the session will not "
```

bigframes/session/__init__.py
```diff
@@ -344,11 +344,25 @@ def _project(self):
...
+            "to get relevant metrics.",
+            UserWarning,
+        )
         return self._metrics.bytes_processed
 
     @property
     def slot_millis_sum(self):
         """The sum of all slot time used by bigquery jobs in this session."""
+        warnings.warn(
+            "Queries executed with `allow_large_results` within the session will not "
```
tswast, line 360:
```suggestion
            "Queries executed with `allow_large_results=False` within the session will not "
```

bigframes/session/_io/bigquery/__init__.py
```diff
@@ -228,14 +228,28 @@ def start_query_with_client(
...
+        if not query_with_job:
+            results_iterator = bq_client.query_and_wait(
+                sql,
+                job_config=job_config,
+                location=location,
+                project=project,
+                api_timeout=timeout,
+            )
+            if metrics is not None:
+                metrics.count_job_stats()
```
tswast, line 250: https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query response includes `totalBytesProcessed`. Let's create a follow-up issue to include those metrics, too.

tswast (approved):

bigframes/session/_io/bigquery/__init__.py
tswast: Filed internal issue 400961399

PR #1452, opened 2025-03-04
fix: do not compare schema description during schema validation

sycai:


chelsea-lin (approved):

chelsea-lin:
Thanks for quick fixing!

PR #1455, opened 2025-03-05
feat: warn when the BigFrames version is more than a year old

sycai:


tswast:
Doctest failures are

```
FAILED third_party/bigframes_vendored/pandas/core/frame.py::frame.DataFrame.apply
FAILED third_party/bigframes_vendored/pandas/core/series.py::series.Series.apply
2 failed, 630 passed, 261 warnings in 5732.11s (1:35:32)
```

These appear to be the usual remote functions flakiness.

bigframes/version.py
```diff
@@ -12,8 +12,17 @@
...
 
 # {x-release-please-start-date}
 __release_date__ = "2025-03-05"
 # {x-release-please-end}
+
+if datetime.datetime.today() - datetime.datetime.strptime(
+    __release_date__, "%Y-%m-%d"
+) > datetime.timedelta(days=365):
+    msg = f"Your BigFrames version {__version__} is beyond the support period. Please update to the lastest version."
+    warnings.warn(msg)
```
tswast, line 28: Let's use a custom category in `bigframes.exceptions`, so that folks can filter these out if they really want to.

Also, doing this at import time isn't the best, as sometimes the import might be done before the user actually sees it (e.g. on notebook startup if colab adds some bigframes-specific hooks, for example). Can we move this to session creation?

bigframes/version.py
```diff
@@ -12,8 +12,17 @@
...
 __version__ = "1.39.0"
 
 # {x-release-please-start-date}
 __release_date__ = "2025-03-05"
 # {x-release-please-end}
+
+if datetime.datetime.today() - datetime.datetime.strptime(
+    __release_date__, "%Y-%m-%d"
+) > datetime.timedelta(days=365):
+    msg = f"Your BigFrames version {__version__} is beyond the support period. Please update to the lastest version."
```
tswast, line 27: This feels a bit too precise. Support policies can change. Let's just say it's greater than 1 year old.

bigframes/version.py
sycai: Sure. Moved the check to session.__init__

tswast (approved):

PR #1458, opened 2025-03-05
feat!: reading JSON data as a custom arrow extension type

chelsea-lin:
We initially implemented a local pandas extension (`db_dtypes.JSONType`) for handling JSON data. Subsequently, the Arrow project introduced a native JSON data type in pyarrow after v19.0. We've opted to adopt this native type as our primary solution (see go/bf-json2 for internal design document). To ensure compatibility for users with older pyarrow versions, we've been using a custom Arrow extension as a fallback. This change transitions to using this custom Arrow extension as a stepping stone towards fully integrating the native pyarrow JSON type.

Release-As: 1.40.0

- [X] Fixes internal issue 401054811

Fixes internal issue 401054811🦕

### DIFF
--- a/bigframes/dtypes.py
+++ b/bigframes/dtypes.py
@@ -62,7 +62,7 @@ BIGNUMERIC_DTYPE = pd.ArrowDtype(pa.decimal256(76, 38))
 # No arrow equivalent
 GEO_DTYPE = gpd.array.GeometryDtype()
 # JSON
-JSON_DTYPE = db_dtypes.JSONDtype()
+JSON_DTYPE = pd.ArrowDtype(db_dtypes.JSONArrowType())
 OBJ_REF_DTYPE = pd.ArrowDtype(
     pa.struct(
         (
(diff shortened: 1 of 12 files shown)


bigframes/dtypes.py
```diff
@@ -62,7 +62,7 @@
 # No arrow equivalent
 GEO_DTYPE = gpd.array.GeometryDtype()
 # JSON
-JSON_DTYPE = db_dtypes.JSONDtype()
+JSON_DTYPE = pd.ArrowDtype(db_dtypes.JSONArrowType())
```
tswast, line 65: Should we switch to `pd.ArrowDtype(pyarrow.json_(pyarrow.string()))` if `pyarrow.json_` is available?

Also, would be good to make sure we align with `OBJ_REF_DTYPE` by creating a `JSON_ARROW_TYPE` variable to use here and there.

tswast (approved):
Thanks! One suggestion, but otherwise looks good.

bigframes/dtypes.py
chelsea-lin: I will follow up with b/401055693. Thanks for reviewing!

PR #1459, opened 2025-03-05
chore: Remove most unit tests, lint, doc testing from kokoro presubmit

TrevorBergeron:
b/Fixes: b/401037761

### DIFF
--- a/bigframes/blob/_functions.py
+++ b/bigframes/blob/_functions.py
@@ -393,7 +393,7 @@ def pdf_extract_func(src_obj_ref_rt: str) -> str:
     return all_text
 
 
-pdf_extract_def = FunctionDef(pdf_extract_func, ["pypdf", "requests"])
+pdf_extract_def = FunctionDef(pdf_extract_func, ["pypdf", "requests", "pypdf[crypto]"])
 
 
 # Extracts text from a PDF url and chunks it simultaneously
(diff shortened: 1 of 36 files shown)


bigframes/blob/_functions.py
```diff
@@ -393,7 +393,7 @@ def pdf_extract_func(src_obj_ref_rt: str) -> str:
     return all_text
 
 
-pdf_extract_def = FunctionDef(pdf_extract_func, ["pypdf", "requests"])
+pdf_extract_def = FunctionDef(pdf_extract_func, ["pypdf", "requests", "pypdf[crypto]"])
```
tswast, line 396: Was this a merge conflict error?

noxfile.py
```diff
@@ -112,16 +112,8 @@
...
 # ahead to fail fast at presubmit running.
-# 'docfx' is excluded since it only needs to run in 'docs-presubmit'
 nox.options.sessions = [
-    "lint",
-    "lint_setup_py",
-    "mypy",
-    "format",
-    "docs",
-    "docfx",
     "unit",
```
tswast, line 116: Why keep `unit` but remove `unit_noextras`?

bigframes/blob/_functions.py
TrevorBergeron: yeah, merge issue, re-updated branch, shouldn't show up anymore now

noxfile.py
TrevorBergeron: yeah, its for coverage, didn't want to mess too much with that, I doubt either unit, or integration tests individually would meet the coverage requirement. Maybe could just run unit tests as part of the integration suite though?

shobsi: In the PR#1460 we are adding back format and mypy, but what about the others?

TrevorBergeron: others are already covered by existing docs, unit test actions, so are already redundant

tswast (approved):

PR #1460, opened 2025-03-05
chore: Add mypy github action

TrevorBergeron:
Fixes: b/401037761


.github/sync-repo-settings.yaml
```diff
@@ -12,6 +12,8 @@ branchProtectionRules:
     - 'cla/google'
     - 'docs'
     - 'lint'
+    - 'format'
```
tswast, line 15: Do we actually want this one? As far as I know GitHub won't commit the changes back. `lint` also runs the formatter but will fail if any makes a change.

.github/workflows/mypy.yml
```diff
@@ -0,0 +1,22 @@
+on:
```
tswast, line 1: If we add this here, I think we can safely remove from https://github.com/googleapis/python-bigquery-dataframes/blob/7b0cab5c504ec2b24ea35b29ee32901da65681b6/noxfile.py#L119 which defines the list used in the kokoro presubmit. Or are you waiting to do that cleanup in a subsequent PR (once `mypy` is required check, perhaps?)

TrevorBergeron: Removing it (and others) in a following PR: https://github.com/googleapis/python-bigquery-dataframes/pull/1459.

.github/sync-repo-settings.yaml
TrevorBergeron: ok, so it looks like `lint` wasn't doing `isort`, so I integrated `isort --check` into the lint workflow, so no need for `format`

.github/workflows/format.yml
tswast, line 1: I think we can remove this file now that lint runs both black and isort.

TrevorBergeron: removed

tswast (approved):

PR #1461, opened 2025-03-05
fix: fix the merge issue between 1424 and 1373

jialuoo:

### DIFF
--- a/bigframes/functions/_function_session.py
+++ b/bigframes/functions/_function_session.py
@@ -455,7 +455,7 @@ class FunctionSession:
 
         # raise a UserWarning if user does not explicitly set cloud_function_service_account to a
         # user-managed cloud_function_service_account of to default
-        msg = (
+        msg = bfe.format_message(
             "You have not explicitly set a user-managed `cloud_function_service_account`. "
             "Using the default Compute Engine service account. "
             "To use Bigframes 2.0, please explicitly set `cloud_function_service_account` "
--- a/bigframes/functions/_function_session.py
+++ b/bigframes/functions/_function_session.py
@@ -47,6 +47,7 @@ from google.cloud import (
 )
 
 from bigframes import clients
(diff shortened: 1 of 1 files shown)


bigframes/functions/_function_session.py
```diff
@@ -445,6 +453,26 @@ def remote_function(
...
+            "Using the default Compute Engine service account. "
+            "To use Bigframes 2.0, please explicitly set `cloud_function_service_account` "
+            'either to a user-managed service account (preferred) or to `"default"` '
+            "to use the Compute Engine service account (discouraged). "
+            "See, https://cloud.google.com/functions/docs/securing/function-identity."
+        )
+
+        if (
+            bigframes_version.__version__.startswith("1.")
+            and cloud_function_service_account is None
```
chelsea-lin, line 469: Can you please check why the `test_remote_function_warns_default_cloud_function_service_account` didn't catch the issue?

jialuoo: This test's logic appears correct . However, the presubmit check is not triggering its execution. Thus bypassing the intended error detection during the merge process.

bigframes/functions/_function_session.py
```diff
@@ -445,6 +453,26 @@ def remote_function(
         # Some defaults may be used from the session if not provided otherwise.
         session = self._resolve_session(session)
 
+        # raise a UserWarning if user does not explicitly set cloud_function_service_account to a
+        # user-managed cloud_function_service_account of to default
+        msg = (
```
chelsea-lin, line 458: After #1281 merged, please call `bfe.format_message` to format the error message. Thanks!

chelsea-lin (approved):
Thanks for quick fixing!

bigframes/functions/_function_session.py
jialuoo: Done.

PR #1462, opened 2025-03-06
chore: Cleanup kokoro artifacts at end of build

TrevorBergeron:

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -50,3 +50,6 @@ if [[ -n "${NOX_SESSION:-}" ]]; then
 else
     python3 -m nox --stop-on-first-error
 fi
+
+# Prevent kokoro from trying to collect many mb of artifacts, wasting several minutes
+sudo rm -rf "${KOKORO_ARTIFACTS_DIR?}"/*


.kokoro/build.sh
```diff
@@ -50,3 +50,6 @@ if [[ -n "${NOX_SESSION:-}" ]]; then
 else
     python3 -m nox --stop-on-first-error
 fi
+
+# Prevent kokoro from trying to collect many mb of artifacts, wasting several minutes
+sudo rm -rf "${KOKORO_ARTIFACTS_DIR?}"/*
```
tswast, line 55: I worry that this might break the release build. Could you check?

Also, I recall there is a https://github.com/googleapis/repo-automation-bots/tree/main/packages/flakybot that could use these to automatically file issues for which specific tests are being flaky. It was in the list of TODOs when we first setup the repo, but I don't think we ever got around to setting it up.

TrevorBergeron: From what I can see, release build uses release.sh for its build file. Flakybot seems to be independent of post-build artifact collection, and is only invoked from samples tests it seems.

tswast (approved):

PR #1464, opened 2025-03-06
perf: Compilation no longer bounded by recursion

TrevorBergeron:

### DIFF
--- a/bigframes/core/bigframe_node.py
+++ b/bigframes/core/bigframe_node.py
@@ -386,6 +386,7 @@ class BigFrameNode:
         return results[self]
 
     def reduce_up(self, reduction: Callable[[BigFrameNode, Tuple[T, ...]], T]) -> T:
+        """Apply a bottom-up reduction to the tree."""
         results: dict[BigFrameNode, T] = {}
         for node in list(self.iter_nodes_topo()):
             # child nodes have already been transformed
--- a/bigframes/core/bigframe_node.py
+++ b/bigframes/core/bigframe_node.py
@@ -32,6 +32,8 @@ if typing.TYPE_CHECKING:
 
 COLUMN_SET = frozenset[identifiers.ColumnId]
 
+T = typing.TypeVar("T")
+
 
 @dataclasses.dataclass(frozen=True)
 class Field:
@@ -379,6 +381,16 @@ class BigFrameNode:
             # child nodes have already been transformed
             result = node.transform_children(lambda x: results[x])
             result = transform(result)
+            results[node] = result
+
+        return results[self]
+
+    def reduce_up(self, reduction: Callable[[BigFrameNode, Tuple[T, ...]], T]) -> T:
+        results: dict[BigFrameNode, T] = {}
+        for node in list(self.iter_nodes_topo()):
+            # child nodes have already been transformed
+            child_results = tuple(results[child] for child in node.child_nodes)
+            result = reduction(node, child_results)
             results[node] = result
 
         return results[self]
--- a/bigframes/core/compile/api.py
+++ b/bigframes/core/compile/api.py
@@ -27,10 +27,6 @@ if TYPE_CHECKING:
 
 
 class SQLCompiler:
-    def __init__(self, strict: bool = True):
-        # Not used, should maybe remove, if strictness is no longer compile-time
-        self._strict = strict
-
     def compile(
         self,
         node: bigframes.core.nodes.BigFrameNode,
--- a/bigframes/core/compile/api.py
+++ b/bigframes/core/compile/api.py
@@ -25,12 +25,11 @@ if TYPE_CHECKING:
     import bigframes.core.ordering
     import bigframes.core.schema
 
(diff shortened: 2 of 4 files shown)


bigframes/core/bigframe_node.py
```diff
@@ -382,3 +384,13 @@ def bottom_up(
             results[node] = result
 
         return results[self]
+
+    def reduce_up(self, reduction: Callable[[BigFrameNode, Tuple[T, ...]], T]) -> T:
+        results: dict[BigFrameNode, T] = {}
+        for node in list(self.iter_nodes_topo()):
+            # child nodes have already been transformed
+            child_results = tuple(results[child] for child in node.child_nodes)
+            result = reduction(node, child_results)
```
chelsea-lin, line 393: The `bottom_up` and `reduce_up` are similar but the transformed methods APIs are different. Do we have a way to merge them into one? If not, could you please add docstring to `reduce_up`

bigframes/core/compile/api.py
```diff
@@ -25,12 +25,11 @@
     import bigframes.core.ordering
     import bigframes.core.schema
 
-_STRICT_COMPILER = compiler.Compiler(strict=True)
-
 
 class SQLCompiler:
     def __init__(self, strict: bool = True):
-        self._compiler = compiler.Compiler(strict=strict)
+        # Not used, should maybe remove, if strictness is no longer compile-time
```
chelsea-lin, line 31: If it's not used, any reasons to keep it still?

bigframes/core/compile/compiler.py
```diff
@@ -41,250 +40,259 @@
...
+
+    offsets = node.offsets_col.sql if node.offsets_col else None
+    return compiled.UnorderedIR.from_pandas(
+        array_as_pd, node.scan_list, offsets=offsets
+    )
+
+
+@_compile_node.register
+def compile_readtable(node: nodes.ReadTableNode, *args):
+    return compile_read_table_unordered(node.source, node.scan_list)
```
chelsea-lin, line 182: nit: `compile_read_table_unordered` is only used within this `compile_readtable` function. Would it improve readability to inline its logic here, or is there a reason for keeping it separate?

bigframes/core/compile/compiler.py
```diff
@@ -41,250 +40,259 @@
...
-                ).name(scan_item.source_id)
-                ibis_table = ibis_table.mutate(json_column)
-
-        return compiled.UnorderedIR(
-            ibis_table,
-            tuple(
-                ibis_table[scan_item.source_id].name(scan_item.id.sql)
-                for scan_item in scan.items
-            ),
+scalar_op_compiler = compile_scalar.ScalarOpCompiler()
```
chelsea-lin, line 43: It's not used in this file. Could we remove it?

bigframes/core/compile/compiler.py
```diff
@@ -41,250 +40,259 @@
...
+        array_as_pd, node.scan_list, offsets=offsets
+    )
+
+
+@_compile_node.register
+def compile_readtable(node: nodes.ReadTableNode, *args):
+    return compile_read_table_unordered(node.source, node.scan_list)
+
+
+def read_table_as_unordered_ibis(
```
chelsea-lin, line 185: nit: `_read_table_as_unordered_ibis` since it's internal used only. Can we move all these helper methods in the bottom of this file?

chelsea-lin (changes requested):
Does the performance run on the presubmit? If so, do we have any data for performance gain from this PR?

bigframes/core/bigframe_node.py
TrevorBergeron: In theory, bottom_up could be implemented in terms of reduce_up, but will leave separate for now, maybe they'll have somewhat different optimizations or extra arguments later on. Added a brief description

bigframes/core/compile/api.py
TrevorBergeron: removed

bigframes/core/compile/compiler.py
TrevorBergeron: I think there was an ordered version at some point, inlined now

bigframes/core/compile/compiler.py
TrevorBergeron: removed

bigframes/core/compile/compiler.py
TrevorBergeron: function removed per other comment.

TrevorBergeron:
> Does the performance run on the presubmit? If so, do we have any data for performance gain from this PR?

Main gains are when you would otherwise hit stack overflow. Though, still a few other cases where you can hit stack overflow, so might not be that helpful by itself?

bigframes/core/compile/compiler.py
chelsea-lin: They're different functions. Could you double check?

chelsea-lin (approved):
Left one nit comment. Overall LGTM

PR #1467, opened 2025-03-06
chore: fix Multimodal Gemini modifies input DF

GarrettWu:
Think it is consistent for output after converting. That modifying converted object doesn't affect the original one.


sycai (commented):
Hmmm, do we really need to make a copy? DataFrames and Series are inherently immutable, right? 

Are you seeing errors when the converter returns the original object instead of the copy?

GarrettWu:
> Hmmm, do we really need to make a copy? DataFrames and Series are inherently immutable, right?
> 
> Are you seeing errors when the converter returns the original object instead of the copy?

No, they are mutable. E.g. assign df["a"] = 1

sycai (approved):

PR #1473, opened 2025-03-07
docs: split DataFrame and Series methods into their own pages

tswast:

Fixes internal issue 401261155 🦕


scripts/generate_docs_toc.py
```diff
@@ -0,0 +1,165 @@
...
+
+        # Avoid trying to call getattr if this is an attribute on a class, not the
+        # module.
+        if "." in attribute:
+            continue
+
+        value = getattr(module, attribute)
+        if inspect.isclass(value):
+            for class_attribute in find_class_attributes(value, attribute):
+                attributes.insert(0, class_attribute)
```
tswast, line 137: Tip from @arwas11 : https://docs.python.org/3/library/collections.html#collections.deque could be faster.

tswast:
Closing. I'd like to revisit this in future. It's possible this script will be useful, but also possibly not, depending on the plugin changes introduced in https://github.com/googleapis/sphinx-docfx-yaml/issues/415 to address this feature.

PR #1474, opened 2025-03-07
test: target e2e tests to python 3.11 for max coverage

shobsi:
BEGIN_COMMIT_OVERRIDE
test: target e2e tests to python 3.11 for max coverage (#1474)

chore: remove version 1.x guard to surface version 2.0 `FutureWarning` (#1474)
END_COMMIT_OVERRIDE

Cloud Run Functions supports Python versions up to 3.12  - https://cloud.google.com/run/docs/runtimes/python.
Managed Python UDF is supported only in Python 3.11. Let's set the E2E tests version to 3.11 to cover most code paths.

Fixes internal issue 401318988 🦕

### DIFF
--- a/tests/system/large/functions/test_managed_function.py
+++ b/tests/system/large/functions/test_managed_function.py
@@ -13,10 +13,7 @@
 # limitations under the License.
 
 import pandas
-import pytest
 
-from bigframes.functions import _function_session as bff_session
-from bigframes.functions._utils import get_python_version
 import bigframes.pandas as bpd
 from tests.system.utils import cleanup_function_assets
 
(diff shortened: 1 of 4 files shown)


Genesis929 (approved):

tests/system/large/functions/test_managed_function.py
```diff
@@ -78,10 +74,6 @@ def multiply(x, y):
         cleanup_function_assets(multiply, bigquery_client)
 
 
-@pytest.mark.skipif(
```
jialuoo, line 81: There is another test case. Please remove the `skipif` for it too.

jialuoo, line 81: There is another test left. Please remove the skipif for it too.

shobsi: ah, nice catch, thanks, done!

jialuoo (approved):

PR #1478, opened 2025-03-10
test: use jobless query api for faster tests

TrevorBergeron:


tests/system/small/test_dataframe.py
```diff
@@ -165,9 +165,10 @@ def test_df_construct_from_dict():
 
 def test_df_construct_inline_respects_location():
     # Note: This starts a thread-local session.
+    bpd.close_session()
```
shobsi, line 168: Is there not enough separation across tests? This is something we should be able to handle through fixtures.

tests/system/small/test_dataframe.py
```diff
@@ -165,9 +165,10 @@ def test_df_construct_from_dict():
...
     with bpd.option_context("bigquery.location", "europe-west1"):
         df = bpd.DataFrame([[1, 2, 3], [4, 5, 6]])
-        repr(df)
+        df.to_gbq()
```
shobsi, line 171: Does `repr` not work when `allow_large_results=False`, meaning the user won't be able to preview the dataframe without peak/head?

tests/system/small/test_pandas_options.py
```diff
@@ -59,6 +59,8 @@ def test_read_gbq_start_sets_session_location(
     read_method,
     query_prefix,
 ):
+    bpd.close_session()
```
shobsi, line 62: Again, this was not needed earlier but needed now suggests that we lost test isolation?

tests/system/small/test_dataframe.py
TrevorBergeron: it does work, it just doesn't generate a query job or results table because it will use the fast path. we are testing query result location here, so we want to force results table creation

tests/system/small/test_dataframe.py
TrevorBergeron: replaced with fixture

tests/system/conftest.py
```diff
@@ -1468,3 +1468,12 @@ def images_mm_df(
     return session.from_glob_path(
         images_gcs_path, name="blob_col", connection=bq_connection
     )
+
+
+@pytest.fixture()
+def reset_default_session_and_location():
+    bpd.close_session()
+    with bpd.option_context("bigquery.location", None):
```
shobsi, line 1476: [nit] probably redundant (default is None, which then defaults to US)?

tests/system/conftest.py
```diff
@@ -1460,3 +1468,12 @@ def images_mm_df(
     return session.from_glob_path(
         images_gcs_path, name="blob_col", connection=bq_connection
     )
+
+
+@pytest.fixture()
+def reset_default_session_and_location():
```
shobsi, line 1474: Looks like we can remove from here now that we added this in the top level conftest.py?

tests/system/conftest.py
TrevorBergeron: yes, redundant, thought I had removed the other one

tests/system/conftest.py
TrevorBergeron: yeah, not needed, removed

tests/system/conftest.py
```diff
@@ -1468,3 +1468,12 @@ def images_mm_df(
...
+    bpd.close_session()
+    with bpd.option_context("bigquery.location", None):
+        yield
+    bpd.close_session()
```
shobsi, line 1478: Looks like lines 1478-1479 are redundant, as the lines 1475-1476 are ensuring the same

shobsi (approved):
Left comment about setting location to None being redundant

PR #1487, opened 2025-03-12
chore: enable multi-model input for sem_map and sem_filter

sycai:


bigframes/operations/semantics.py
```diff
@@ -489,14 +516,29 @@ def map(
...
 
-        results = typing.cast(
-            bigframes.series.Series,
-            model.predict(
-                self._make_prompt(df, columns, user_instruction, output_instruction),
-                temperature=0.0,
-                ground_with_google_search=ground_with_google_search,
-            )["ml_generate_text_llm_result"],
-        )
+        if has_blob_column:
```
chelsea-lin, line 519: Does this code handle scenarios with both 'blob' and 'text' columns? If so, please add a test case to demonstrate it.

bigframes/operations/semantics.py
```diff
@@ -489,14 +516,29 @@ def map(
...
+                        df, columns, user_instruction, output_instruction
+                    ),
+                    temperature=0.0,
+                    ground_with_google_search=ground_with_google_search,
+                )["ml_generate_text_llm_result"],
+            )
+        else:
+            results = typing.cast(
+                bigframes.series.Series,
+                model.predict(
```
chelsea-lin, line 534: It looks `model.predict` has an uniform API for both text and multimodel prompt. Can we please try to reduce the duplicates code here?

bigframes/operations/semantics.py
sycai: Sounds good. Test updated.

bigframes/operations/semantics.py
sycai: Unfortunately it's less desirable at the moment. 

The model predict API relies on the present of "prompt" parameter to tell wether there we should use multi-model, and if yes, we need to turn on the multi-model experiment flag.

Unifying both API calls basically means we need the users to turn on the multi-model experiment even when their inputs are pure texts, which I think don't make much sense.

We won't have this problem anymore once I unify all the experiment flags into one.

chelsea-lin (approved):
LGTM overall. Left two nit comments.

GarrettWu:
MM isn't published yet. You may want to change the title to "chore".

tests/system/large/operations/test_semantics.py
```diff
@@ -388,6 +389,27 @@ def test_filter(session, gemini_flash_model):
...
+def test_filter_multi_model(session, gemini_flash_model):
+    with bigframes.option_context(
+        SEM_OP_EXP_OPTION,
+        True,
+        BLOB_EXP_OPTION,
+        True,
+        THRESHOLD_OPTION,
+        10,
+    ):
+        df = session.from_glob_path("gs://sycai_bucket/images/*", name="image")
```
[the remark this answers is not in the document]
sycai: Done!

tests/system/small/operations/test_semantics.py
```diff
@@ -0,0 +1,124 @@
...
+
+class FakeGeminiTextGenerator(llm.GeminiTextGenerator):
+    def __init__(self, prediction):
+        self.prediction = prediction
+
+    def predict(self, *args, **kwargs):
+        return self.prediction
+
+
+def test_semantics_experiment_off_raise_error(session):
```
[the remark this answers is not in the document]
sycai: Sure. I will circle back once we have more time.

PR #1489, opened 2025-03-12
perf: eliminate count queries in llm retry

GarrettWu:
screen/Ak8x7ZQBUkL2CPx


sycai (approved):
Thanks! Looks like there are some flakiness in the presubmit. Might need to double check before merging.

PR #1490, opened 2025-03-13
chore: add error handle for pdf extract and pdf chunk

shuoweil:
chore: add error handle for pdf extract and pdf chunk
add docs and testcase for pdf extract and pdf chunk
testcases are verified: screen/vFUmM6VeAFwhYv9

### DIFF
--- a/bigframes/blob/_functions.py
+++ b/bigframes/blob/_functions.py
@@ -425,10 +425,10 @@ def pdf_extract_func(src_obj_ref_rt: str) -> str:
             if page_extract_text:
                 all_text += page_extract_text
 
-        result_dict = {"status": None, "content": all_text}
+        result_dict = {"status": "", "content": all_text}
 
     except Exception as e:
-        result_dict = {"status": str(e), "content": None}
+        result_dict = {"status": str(e), "content": ""}
 
     result_json = json.dumps(result_dict)
     return result_json
--- a/bigframes/blob/_functions.py
+++ b/bigframes/blob/_functions.py
@@ -434,45 +434,57 @@ def pdf_chunk_func(src_obj_ref_rt: str, chunk_size: int, overlap_size: int) -> s
     import io
     import json
 
-    from pypdf import PdfReader  # type: ignore
+    from pypdf import errors, PdfReader  # type: ignore
     import requests
     from requests import adapters
 
-    session = requests.Session()
-    session.mount("https://", adapters.HTTPAdapter(max_retries=3))
+    try:
+        session = requests.Session()
+        session.mount("https://", adapters.HTTPAdapter(max_retries=3))
 
-    src_obj_ref_rt_json = json.loads(src_obj_ref_rt)
-    src_url = src_obj_ref_rt_json["access_urls"]["read_url"]
+        src_obj_ref_rt_json = json.loads(src_obj_ref_rt)
+        src_url = src_obj_ref_rt_json["access_urls"]["read_url"]
 
-    response = session.get(src_url, timeout=30, stream=True)
-    response.raise_for_status()
-    pdf_bytes = response.content
+        response = session.get(src_url, timeout=30, stream=True)
+        response.raise_for_status()
+        pdf_bytes = response.content
 
-    pdf_file = io.BytesIO(pdf_bytes)
-    reader = PdfReader(pdf_file, strict=False)
+        pdf_file = io.BytesIO(pdf_bytes)
+        reader = PdfReader(pdf_file, strict=False)
 
-    # extract and chunk text simultaneously
-    all_text_chunks = []
-    curr_chunk = ""
-    for page in reader.pages:
-        page_text = page.extract_text()
-        if page_text:
-            curr_chunk += page_text
-            # split the accumulated text into chunks of a specific size with overlaop
-            # this loop implements a sliding window approach to create chunks
-            while len(curr_chunk) >= chunk_size:
-                split_idx = curr_chunk.rfind(" ", 0, chunk_size)
-                if split_idx == -1:
-                    split_idx = chunk_size
-                actual_chunk = curr_chunk[:split_idx]
-                all_text_chunks.append(actual_chunk)
-                overlap = curr_chunk[split_idx + 1 : split_idx + 1 + overlap_size]
-                curr_chunk = overlap + curr_chunk[split_idx + 1 + overlap_size :]
-    if curr_chunk:
-        all_text_chunks.append(curr_chunk)
+        # extract and chunk text simultaneously
+        all_text_chunks = []
+        curr_chunk = ""
+        for page in reader.pages:
+            page_text = page.extract_text()
+            if page_text:
+                curr_chunk += page_text
+                # split the accumulated text into chunks of a specific size with overlaop
+                # this loop implements a sliding window approach to create chunks
+                while len(curr_chunk) >= chunk_size:
+                    split_idx = curr_chunk.rfind(" ", 0, chunk_size)
+                    if split_idx == -1:
+                        split_idx = chunk_size
+                    actual_chunk = curr_chunk[:split_idx]
+                    all_text_chunks.append(actual_chunk)
+                    overlap = curr_chunk[split_idx + 1 : split_idx + 1 + overlap_size]
+                    curr_chunk = overlap + curr_chunk[split_idx + 1 + overlap_size :]
+        if curr_chunk:
+            all_text_chunks.append(curr_chunk)
 
-    all_text_json_string = json.dumps(all_text_chunks)
-    return all_text_json_string
+        all_text_json_string = json.dumps(all_text_chunks)
+        return {"status": None, "content": all_text_json_string}
+
+    except requests.exceptions.RequestException as e:
+        return {"status": f"Request error: {str(e)}", "content": None}
+    except json.JSONDecodeError as e:
+        return {"status": f"JSON decode error: {str(e)}", "content": None}
+    except KeyError as e:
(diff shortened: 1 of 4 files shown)


bigframes/blob/_functions.py
```diff
@@ -436,41 +436,45 @@ def pdf_chunk_func(src_obj_ref_rt: str, chunk_size: int, overlap_size: int) -> s
...
-                actual_chunk = curr_chunk[:split_idx]
-                all_text_chunks.append(actual_chunk)
-                overlap = curr_chunk[split_idx + 1 : split_idx + 1 + overlap_size]
-                curr_chunk = overlap + curr_chunk[split_idx + 1 + overlap_size :]
-    if curr_chunk:
-        all_text_chunks.append(curr_chunk)
-
-    all_text_json_string = json.dumps(all_text_chunks)
-    return all_text_json_string
+    try:
```
GarrettWu, line 439: try can move to the top of the function, so it also captures import errors.

bigframes/operations/blob.py
```diff
@@ -663,5 +664,16 @@ def pdf_chunk(
 
         res = df.apply(pdf_chunk_udf, axis=1)
 
-        res_array = bbq.json_extract_string_array(res)
-        return res_array
+        status_series = bbq.json_extract_string_array(res, "$.status")
```
GarrettWu, line 667: Lets add a parameter "verbose". 

If True, return struct<str, array<str>>.

If False, just return array<str>. (Or None if error)

bigframes/operations/blob.py
```diff
@@ -663,5 +664,16 @@ def pdf_chunk(
...
+        content_series = bbq.json_extract_string_array(res, "$.content")
+        result_series = bpd.Series(
+            [
+                content
```
GarrettWu, line 671: make a struct. see https://github.com/googleapis/python-bigquery-dataframes/blob/main/bigframes/bigquery/_operations/struct.py

GarrettWu:
Can we also add tests as those are already available in PROD now. like https://github.com/googleapis/python-bigquery-dataframes/pull/1482

shuoweil:
> Can we also add tests as those are already available in PROD now. like #1482
done

bigframes/blob/_functions.py
```diff
@@ -398,81 +398,92 @@ def image_normalize_to_bytes_func(
...
-        page_extract_text = page.extract_text()
-        if page_extract_text:
-            all_text += page_extract_text
-    return all_text
+        result_dict = {"status": "", "content": all_text}
+        return json.dumps(result_dict)
+
+    except Exception as e:
+        result_dict = {"status": str(e), "content": ""}
+        return json.dumps(result_dict)
```
GarrettWu, line 433: You can put the return outside of try/except block, then you only need one return, also it catches the failures in json.dumps().

bigframes/operations/blob.py
```diff
@@ -597,8 +605,17 @@ def pdf_extract(
         ).udf()
 
         src_rt = self._get_runtime_json_str(mode="R")
+
         res = src_rt.apply(pdf_extract_udf)
-        return res
+
+        content_series = bbq.json_extract(res, "$.content")
```
GarrettWu, line 611: This returns a JSON string instead of string. Still need JSON_VALUE.

bigframes/operations/blob.py
```diff
@@ -609,6 +626,7 @@ def pdf_chunk(
         max_batching_rows: int = 1,
         container_cpu: Union[float, int] = 2,
         container_memory: str = "1Gi",
+        verbose: bool = True,
```
GarrettWu, line 629: Default to false.

bigframes/operations/blob.py
```diff
@@ -629,14 +647,20 @@ def pdf_chunk(
...
             container_memory (str, default "1Gi"): container memory size. String of the format <number><unit>. Possible values are from 512Mi to 32Gi.
+            verbose (bool, default "True"): controls the verbosity of the output.
+                When set to True, both error messages and the extracted content
+                are displayed. Conversely, when set to False, only the extracted
+                content is presented, suppressing error messages.
 
         Returns:
-            bigframe.series.Series: Series of array[str], where each string is a
-                chunk of text extracted from PDF.
+            bigframe.series.Series: Series of array[str] | str,
```
GarrettWu, line 656: array[str] or struct[str, array[str]], depend on the "verbose" parameter.

tests/system/large/blob/test_function.py
```diff
@@ -285,3 +285,93 @@ def test_blob_image_normalize_to_bq(images_mm_df: bpd.DataFrame, bq_connection:
...
+    actual_exploded = (
+        pdf_mm_df["pdf"]
+        .blob.pdf_extract(connection=bq_connection)
+        .struct.explode()
+        .to_pandas()
+    )
+    extract_data = [
+        {"status": '"File has not been decrypted"', "content": '""'},
+        {
+            "status": '""',
```
GarrettWu, line 310: Does it feel awkward to have 2 sets of quotes? That is using JSON_EXTRACT, which returns JSON string.

tests/system/large/blob/test_function.py
```diff
@@ -285,3 +285,93 @@ def test_blob_image_normalize_to_bq(images_mm_df: bpd.DataFrame, bq_connection:
...
+
+@pytest.fixture(scope="session")
+def pdf_uris() -> list[str]:
+    return [
+        "uri: gs://bigframes_blob_test/pdfs/test-protected.pdf",
+        "uri: gs://bigframes_blob_test/pdfs/sample-local-pdf.pdf",
+    ]
+
+
+def test_blob_pdf_extract(pdf_mm_df: bpd.DataFrame, bq_connection: str):
```
GarrettWu, line 298: Test for both verbose=True or False.

tests/system/large/blob/test_function.py
```diff
@@ -285,3 +285,93 @@ def test_blob_image_normalize_to_bq(images_mm_df: bpd.DataFrame, bq_connection:
...
+            "content": [
+                "Sample  PDF    This  is  a  testing  file.  Some ",
+                "dummy  messages  are  used  for  testing ",
+                "purposes.   ",
+            ],
+        },
+    ]
+    expected_df = pd.DataFrame(
+        {
+            "pdf": pdf_uris,
```
GarrettWu, line 359: no need to compare "pdf" and "authorizer", inputs are exactly the same, only draws extra confusion.

tests/system/large/blob/test_function.py
```diff
@@ -285,3 +285,93 @@ def test_blob_image_normalize_to_bq(images_mm_df: bpd.DataFrame, bq_connection:
...
+            "authorizer": [bq_connection.casefold(), bq_connection.casefold()],
+            "values": extract_data,
+        }
+    )
+    actual = pd.DataFrame(
+        {
+            "pdf": pdf_uris,
+            "authorizer": [bq_connection.casefold(), bq_connection.casefold()],
+            "values": actual_exploded.to_dict("records"),
```
GarrettWu, line 325: Here it created a pd.DF, convert to a dict, pass in to pd.DF constructor again, doesn't make sense.
 
Just take the simple path, do sth in bigframes, after all the operations, convert to pd.DF or Series, then compare the result with the expected pd.DF or pd.Series

bigframes/blob/_functions.py
```diff
@@ -398,81 +398,92 @@ def image_normalize_to_bytes_func(
...
+                    overlap = curr_chunk[split_idx + 1 : split_idx + 1 + overlap_size]
+                    curr_chunk = overlap + curr_chunk[split_idx + 1 + overlap_size :]
+        if curr_chunk:
+            all_text_chunks.append(curr_chunk)
+
+        result_dict = {"status": "", "content": all_text_chunks}
+        return json.dumps(result_dict)
+
+    except Exception as e:
+        result_dict = {"status": str(e), "content": []}
```
GarrettWu, line 485: Is it possible to return None in content or it has to be empty str/array?

tests/system/large/blob/test_function.py
```diff
@@ -285,3 +285,82 @@ def test_blob_image_normalize_to_bq(images_mm_df: bpd.DataFrame, bq_connection:
...
+    )
+
+    extract_data = [
+        {"status": "File has not been decrypted", "content": ""},
+        {
+            "status": "",
+            "content": "Sample  PDF    This  is  a  testing  file.  Some  dummy  messages  are  used  for  testing  purposes.   ",
+        },
+    ]
+    if verbose:
```
GarrettWu, line 320: Put the output also in @pytest.mark.parametrize, instead of a control block. To improve readability. 

Generally we want tests to be as easy as possible. go/tott/643

tests/system/large/blob/test_function.py
```diff
@@ -285,3 +285,82 @@ def test_blob_image_normalize_to_bq(images_mm_df: bpd.DataFrame, bq_connection:
...
+        expected = pd.Series(chunk_data)
+    else:
+        content_values = [item["content"] for item in chunk_data]
+        expected = pd.Series(content_values)
+    pd.testing.assert_series_equal(
+        actual.to_pandas(),
+        expected,
+        check_dtype=False,
+        check_index=False,
+        check_names=False,
```
GarrettWu, line 365: It can check names, if you pass in the series name in creation.

GarrettWu:
Generally don't enable auto-merge unless the PR is just trivial. In case the reviewer wants to approve with minor comments to fix.

GarrettWu (approved):

PR #1492, opened 2025-03-13
chore: Finish factoring unit tests from presubmit

TrevorBergeron:


noxfile.py
```diff
@@ -119,10 +119,10 @@
 # Sessions are executed in the order so putting the smaller sessions
 # ahead to fail fast at presubmit running.
 nox.options.sessions = [
-    "unit",
```
tswast, line 122: This is because it's redundant with GitHub Actions, right?

Doesn't look like https://github.com/googleapis/python-bigquery-dataframes/blob/bd07e05d26820313c052eaf41c267a1ab20b4fc6/.github/workflows/unittest.yml#L61 actually uses nox for the coverage step, so that's probably okay to just be system tests.

tswast (approved):

PR #1497, opened 2025-03-17
feat: Allow iloc to support lists of negative indices

Genesis929:


tests/system/small/test_dataframe.py
```diff
@@ -4409,7 +4409,7 @@ def test_loc_list_multiindex(scalars_dfs_maybe_ordered):
 
 
 def test_iloc_list(scalars_df_index, scalars_pandas_df_index):
-    index_list = [0, 0, 0, 5, 4, 7]
+    index_list = [0, 0, 0, 5, 4, 7, -2, -5, 3]
```
GarrettWu, line 4412: Do we support base case of iloc[neg_number]?

GarrettWu (approved):

tests/system/small/test_dataframe.py
Genesis929: Seems works.

bigframes/core/indexers.py
```diff
@@ -477,6 +478,24 @@ def _iloc_getitem_series_or_dataframe(
...
+
+        # Check if both positive index and negative index are necessary
+        if isinstance(key, bigframes.series.Series):
+            # Avoid data download
+            is_key_unisigned = False
+        elif "shape" in series_or_dataframe._block.__dict__:
+            # If there is a cache, we convert all indices to positive.
+            row_count = series_or_dataframe._block.shape[0]
+            key = [k if k >= 0 else row_count + k for k in key]
+            is_key_unisigned = True
```
TrevorBergeron, line 490: This seems a bit fragile. We can use `block.expr.node.row_count`, but going though `shape` depends on some implementation details that might change. I don't know if we necessarily need this optimization at all?

bigframes/core/indexers.py
```diff
@@ -497,7 +516,23 @@ def _iloc_getitem_series_or_dataframe(
         block = df._block
         # explicitly set index to offsets, reset_index may not generate offsets in some modes
         block, offsets_id = block.promote_offsets("temp_iloc_offsets_")
-        block = block.set_index([offsets_id])
+        pos_block = block.set_index([offsets_id])
+
+        if not is_key_unisigned or key[0] < 0:
+            neg_block, _ = block.apply_window_op(
+                offsets_id,
+                ops.aggregations.ReverseRowNumberOp(),
```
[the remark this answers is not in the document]
Genesis929: Updated, using SizeUnaryOp and SubOp instead.

bigframes/core/indexers.py
Genesis929: Removed

shuoweil (approved):

bigframes/core/indexers.py
```diff
@@ -477,6 +478,19 @@ def _iloc_getitem_series_or_dataframe(
                 Union[bigframes.dataframe.DataFrame, bigframes.series.Series],
                 series_or_dataframe.iloc[0:0],
             )
+
+        # Check if both positive index and negative index are necessary
+        if isinstance(key, bigframes.series.Series):
```
TrevorBergeron, line 483: might need to check for bigframes.Index as well? Or maybe we should have a helper that helps identify an "remote" or "large" object we don't want to iterate over

TrevorBergeron (approved):

bigframes/core/indexers.py
Genesis929: Updated to also check index type. 

For large object it's may not be necessary, in cloudtop, I tried 1 million keys(which have the same sign), and this process took 0.03s.

PR #1499, opened 2025-03-17
fix: fix pandas.cut errors with empty bins

chelsea-lin:
Fixes internal issue 403638910 🦕

### DIFF
--- a/bigframes/core/reshape/tile.py
+++ b/bigframes/core/reshape/tile.py
@@ -41,14 +41,19 @@ def cut(
     right: typing.Optional[bool] = True,
     labels: typing.Union[typing.Iterable[str], bool, None] = None,
 ) -> bigframes.series.Series:
-    if isinstance(bins, int) and bins <= 0:
-        raise ValueError("`bins` should be a positive integer.")
+    if labels is not None and labels is not False:
+        raise NotImplementedError(
+            "The 'labels' parameter must be either False or None. "
+            "Please provide a valid value for 'labels'."
+        )
 
-    # TODO: Check `right` does not apply for IntervalIndex.
-
-    if isinstance(bins, typing.Iterable):
+    if isinstance(bins, int):
+        if bins <= 0:
+            raise ValueError("`bins` should be a positive integer.")
+        op = agg_ops.CutOp(bins, right=right, labels=labels)
+        return x._apply_window_op(op, window_spec=window_specs.unbound())
+    elif isinstance(bins, typing.Iterable):
         if isinstance(bins, pd.IntervalIndex):
-            # TODO: test an empty internval index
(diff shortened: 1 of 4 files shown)


tests/system/small/test_pandas.py
```diff
@@ -458,47 +467,36 @@ def test_cut_numeric_breaks(scalars_dfs, breaks, right):
...
+    pd_result_converted = _convert_pandas_category(pd_result)
 
     pd.testing.assert_series_equal(
         bf_result, pd_result_converted, check_index=False, check_dtype=False
     )
 
 
 @pytest.mark.parametrize(
-    ("bins",),
+    ("bins"),
```
sycai, line 478: nit: I think we could just remove the parentheses

bigframes/core/reshape/tile.py
```diff
@@ -88,10 +84,13 @@ def cut(
             "Please provide a valid value for 'labels'."
         )
 
-    return x._apply_window_op(
-        agg_ops.CutOp(bins, right=right, labels=labels),
-        window_spec=window_specs.unbound(),
-    )
+    op = agg_ops.CutOp(bins, right=right, labels=labels)
+    if isinstance(bins, typing.Iterable) and len(as_index) == 0:
```
sycai, line 88: Could we directly return the [pd.NA] * len(x) result from the elif branch `len(list(bins)) == 0`?

In that case we don't need the additional if-else branch at the bottom of the function, and the logic looks more straightforward. Plus, we don't need to alter the behavior of CutOp.output_type() in aggregations.py too

chelsea-lin: All of the `elif` branch can turn to the `[pd.NA] * len(x)` result, such as 
- 1st elif for `pd.IntervalIndex.from_tuples([])`. 
- 2nd elif for `[]`
- 4st elif for `[1]`
Because of that, `CutOp.output_type()` might return different result for each case above.

tests/system/small/test_pandas.py
chelsea-lin: Done.

bigframes/core/reshape/tile.py
sycai: My main concern is that the "if" check on line 88 has quite some overlap with line 49 + line 56. It seems we are doing the same check repeatedly.

Anyway, this shouldn't affect the overall functionality.

sycai (approved):

bigframes/core/reshape/tile.py
chelsea-lin: Just refactor these `if` branches for more readable. With this refactor, we also catches another edge cases bug. Thanks!

PR #1516, opened 2025-03-20
chore: throws NotImplementedError when read_pandas with nested JSON type

chelsea-lin:
Fixes internal issue 395912450 🦕


tswast (approved):
Thanks!

PR #1522, opened 2025-03-21
chore: assert test udfs active deletion, make pre-commit mypy dep consistent with nox

shobsi:

### DIFF
--- a/tests/system/large/functions/test_managed_function.py
+++ b/tests/system/large/functions/test_managed_function.py
@@ -72,7 +72,7 @@ def test_managed_function_multiply_with_ibis(
         )
     finally:
         # clean up the gcp assets created for the managed function.
-        cleanup_function_assets(multiply, bigquery_client)
+        cleanup_function_assets(multiply, bigquery_client, ignore_failures=False)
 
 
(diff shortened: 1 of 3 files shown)


jialuoo (approved):

tests/system/large/functions/test_managed_function.py
```diff
@@ -72,7 +72,7 @@ def multiply(x, y):
         )
     finally:
         # clean up the gcp assets created for the managed function.
-        cleanup_function_assets(multiply, bigquery_client)
+        cleanup_function_assets(multiply, bigquery_client, ignore_failures=False)
```
GarrettWu, line 75: Why ignore_failures=False? If the deletion doesn't success, no need to fail out tests

shobsi: Yes that is the behavior so far. But with the recent mystery around managed UDFs still growing in the load testing project, I am introducing this change to not accept any clean up failures.

shobsi:
Merging despite the e2e failures, which is due to the cloud run quota issue and being investigated.

PR #1530, opened 2025-03-24
chore: cleanup udf resources in blob transform functions

GarrettWu:

### DIFF
--- a/bigframes/operations/blob.py
+++ b/bigframes/operations/blob.py
@@ -238,6 +238,10 @@ class BlobAccessor(base.SeriesMethods):
         for _, row in pandas_df.iterrows():
             display_single_url(row["read_url"], row["content_type"])
 
+    @property
+    def session(self):
+        return self._block.session
+
     def _resolve_connection(self, connection: Optional[str] = None) -> str:
         """Resovle the BigQuery connection.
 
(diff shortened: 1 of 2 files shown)


shuoweil (dismissed):

bigframes/operations/blob.py
```diff
@@ -286,6 +290,13 @@ def _get_runtime_json_str(
         runtime = self._get_runtime(mode=mode, with_metadata=with_metadata)
         return runtime._apply_unary_op(ops.ToJSONString())
 
+    # TODO(b/404605969): remove cleanups when UDF fixes dataset deletion.
+    def _add_to_cleanup_set(self, udf):
```
shobsi, line 294: Recommend reusing existing mechanism:

```python
self.session._function_session._update_temp_artifacts(udf.bigframes_bigquery_function, "")
```

Example https://github.com/googleapis/python-bigquery-dataframes/blob/578081e978f2cca21ddae8b3ee371972ba723777/bigframes/functions/_function_session.py#L985

bigframes/session/__init__.py
```diff
@@ -398,6 +398,14 @@ def close(self):
                 self.bqclient, self.cloudfunctionsclient, self.session_id
             )
 
+        # TODO(b/404605969): remove cleanups when UDF fixes dataset deletion.
```
shobsi, line 401: After we accept the suggestion in the other comment, this will no longer be needed

bigframes/operations/blob.py
GarrettWu: Done. Thanks.

bigframes/session/__init__.py
GarrettWu: Done.

shobsi:
It seems not all tests are using the `session` fixture (which ensures `session.close()`), e.g. `test_blob_image_blur_to_folder` - this means some tests will still leave the BQ udfs behind after this change?

GarrettWu:
> It seems not all tests are using the `session` fixture (which ensures `session.close()`), e.g. `test_blob_image_blur_to_folder` - this means some tests will still leave the BQ udfs behind after this change?

It does, implicitly in the input dataframe.

shobsi (approved):

shuoweil (approved):

PR #1531, opened 2025-03-24
chore: include source columns in semantics.top_k

chelsea-lin:
BREAKING CHANGE: The experimental `semantics.top_k` function now returns all columns from the input DataFrame, instead of only the columns directly involved in the ranking. 

Fixes internal issue 404912585 🦕


sycai (approved):
Thanks!

PR #1532, opened 2025-03-24
chore: disable managed function tests temporarily

shobsi:
This is to give a breather to the BigQuery managed udf team to troubleshoot the quota issue.


GarrettWu (approved):

tests/system/large/blob/test_function.py
shobsi, line 1: It seems not all tests are using the session fixture (which does session.close() in the end), e.g. `test_blob_image_blur_to_folder` - this means some tests will leave the BQ udfs?

PR #1536, opened 2025-03-25
feat: upgrade BQ managed `udf` to preview

shobsi:

### DIFF
--- a/bigframes/functions/_function_client.py
+++ b/bigframes/functions/_function_client.py
@@ -53,6 +53,9 @@ _INGRESS_SETTINGS_MAP = types.MappingProxyType(
     }
 )
 
+# BQ managed functions (@udf) currently only support Python 3.11.
+_MANAGED_FUNC_PYTHON_VERSION = "python-3.11"
+
 
 class FunctionClient:
     # Wait time (in seconds) for an IAM binding to take effect after creation.
@@ -193,11 +196,20 @@ class FunctionClient:
         name,
         packages,
         is_row_processor,
+        *,
+        capture_references=False,
     ):
         """Create a BigQuery managed function."""
-        import cloudpickle
 
-        pickled = cloudpickle.dumps(func)
+        if capture_references:
+            # Check the Python version.
+            python_version = _utils.get_python_version()
+            if python_version != _MANAGED_FUNC_PYTHON_VERSION:
+                raise bf_formatting.create_exception_with_feedback_link(
+                    RuntimeError,
+                    f"Python version {python_version} is not supported yet for "
+                    "BigFrames managed function.",
+                )
 
         # Create BQ managed function.
         bq_function_args = []
@@ -209,13 +221,15 @@ class FunctionClient:
             bq_function_args.append(f"{name_} {type_}")
 
         managed_function_options = {
-            "runtime_version": _utils.get_python_version(),
+            "runtime_version": _MANAGED_FUNC_PYTHON_VERSION,
             "entry_point": "bigframes_handler",
         }
 
         # Augment user package requirements with any internal package
         # requirements.
-        packages = _utils._get_updated_package_requirements(packages, is_row_processor)
+        packages = _utils._get_updated_package_requirements(
+            packages, is_row_processor, capture_references
+        )
         if packages:
             managed_function_options["packages"] = packages
         managed_function_options_str = self._format_function_options(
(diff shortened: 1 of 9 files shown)


bigframes/functions/_function_client.py
```diff
@@ -193,11 +196,20 @@ def provision_bq_managed_function(
...
     ):
         """Create a BigQuery managed function."""
-        import cloudpickle
 
-        pickled = cloudpickle.dumps(func)
+        if capture_references:
+            # Check the Python version.
+            python_version = _utils.get_python_version()
+            if python_version != _MANAGED_FUNC_PYTHON_VERSION:
+                raise bf_formatting.create_exception_with_feedback_link(
```
jialuoo, line 208: In this error message, we should add more details about "capture_references=True cases has limitation on versions ...".

bigframes/functions/_function_client.py
```diff
@@ -193,11 +196,20 @@ def provision_bq_managed_function(
         name,
         packages,
         is_row_processor,
+        *,
+        capture_references=False,
```
jialuoo, line 200: nit: add comments about why we need capture_references.

bigframes/functions/_function_session.py
```diff
@@ -809,17 +806,8 @@ def udf(
...
 
-        # Check the Python version.
-        python_version = _utils.get_python_version()
-        if python_version not in _MANAGED_FUNC_PYTHON_VERSIONS:
-            raise bf_formatting.create_exception_with_feedback_link(
-                RuntimeError,
-                f"Python version {python_version} is not supported yet for "
-                "BigFrames managed function.",
-            )
+        warnings.warn("udf is in preview.", category=bfe.PreviewWarning)
```
jialuoo, line 810: This is a user-facing warning. To maintain consistency, I suggest we replace 'udf' with 'managed function' (I feel I used latter one in some other user-facing messages).

I'm flexible on which term we use, but it should be consistent.

tests/system/large/functions/test_managed_function.py
```diff
@@ -121,21 +120,12 @@ def stringify(x):
         cleanup_function_assets(stringify, bigquery_client, ignore_failures=False)
 
 
-@pytest.mark.parametrize(
```
jialuoo, line 130: Why do we remove this?

bigframes/functions/_function_client.py
```diff
@@ -193,11 +196,20 @@ def provision_bq_managed_function(
...
         """Create a BigQuery managed function."""
-        import cloudpickle
 
-        pickled = cloudpickle.dumps(func)
+        if capture_references:
+            # Check the Python version.
+            python_version = _utils.get_python_version()
+            if python_version != _MANAGED_FUNC_PYTHON_VERSION:
+                raise bf_formatting.create_exception_with_feedback_link(
+                    RuntimeError,
```
jialuoo, line 209: Should it be NotImplementedError?

tests/system/large/functions/test_managed_function.py
jialuoo, line 1: We should add a test for 'none self contained udf'.

bigframes/functions/_function_client.py
shobsi: It's there in the if-else clause down below (line#204)

bigframes/functions/_function_session.py
shobsi: I changed "managed function" -> "udf" in couple of messages, PTAL. I agree it should be consistent. Let's use "udf" to reflect the BigFrames API, as that is what it is named, and consistently maintain that it is built on top of "BigQuery managed function".

tests/system/large/functions/test_managed_function.py
shobsi: Because now we don't support the outside references, parameterize comes as outside reference and resolved dynamically, which was handled well with cloudpickle but not anymore. Once we expose outside reference code paths to user again we will be able to use parametrize again.

bigframes/functions/_function_client.py
shobsi: Updated, thanks!

tests/system/large/functions/test_managed_function.py
shobsi: This functionality is not exposed in the public `udf` API (even though we kept the code path internally to expose later b/406283812). It will require extra consideration to test restricted to the supported python version, let's defer to that work.

bigframes/functions/_function_client.py
shobsi: Rephrased the error message, PTAL, thanks!

jialuoo (approved):
Nice job!

shobsi:
Even though the managed udfs tests are disabled currently, I ran them locally successfully.

PR #1540, opened 2025-03-25
refactor: break up groupby/__init__.py file

sycai:


bigframes/core/groupby/aggs.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2023 Google LLC
```
chelsea-lin, line 1: nit: 2025

chelsea-lin: Also applied to other new files.

chelsea-lin (approved):

PR #1541, opened 2025-03-25
feat!: set `allow_large_results=False` by default

shobsi:
This is to optimize small result operations (which are more common). If the result can exceed 10MB, set `bigframes.pandas.options.bigquery.allow_large_results=True` explicitly.

Fixes internal issue 394658588 🦕

### DIFF
--- a/tests/system/small/test_progress_bar.py
+++ b/tests/system/small/test_progress_bar.py
@@ -55,6 +55,19 @@ def test_progress_bar_scalar(penguins_df_default_index: bf.dataframe.DataFrame, 
     with bf.option_context("display.progress_bar", "terminal"):
         penguins_df_default_index["body_mass_g"].head(10).mean()
 
+    assert capsys.readouterr().out == ""
+
+
+def test_progress_bar_scalar_allow_large_results(
+    penguins_df_default_index: bf.dataframe.DataFrame, capsys
+):
+    capsys.readouterr()  # clear output
+
+    with bf.option_context(
+        "display.progress_bar", "terminal", "bigquery.allow_large_results", "True"
+    ):
+        penguins_df_default_index["body_mass_g"].head(10).mean()
+
     assert_loading_msg_exist(capsys.readouterr().out)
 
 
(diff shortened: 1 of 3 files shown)


tswast:
Looks like we need to update the benchmark script. Some metrics aren't available yet with `allow_large_results=False`. https://github.com/googleapis/python-bigquery/issues/1996 has been closed, but I don't think we've updated bigframes to take advantage of it yet.

```
nox > python scripts/run_and_publish_benchmark.py --notebook --publish-benchmarks=notebooks/
Traceback (most recent call last):
  File "/tmpfs/src/github/python-bigquery-dataframes/scripts/run_and_publish_benchmark.py", line 486, in <module>
    main()
  File "/tmpfs/src/github/python-bigquery-dataframes/scripts/run_and_publish_benchmark.py", line 447, in main
    benchmark_metrics, error_message = collect_benchmark_result(
  File "/tmpfs/src/github/python-bigquery-dataframes/scripts/run_and_publish_benchmark.py", line 102, in collect_benchmark_result
    raise ValueError(
ValueError: Mismatch in the number of report files for bytes, millis, seconds and query char count.
```

tests/system/small/test_progress_bar.py
```diff
@@ -55,6 +55,19 @@ def test_progress_bar_scalar(penguins_df_default_index: bf.dataframe.DataFrame,
...
+    penguins_df_default_index: bf.dataframe.DataFrame, capsys
+):
+    capsys.readouterr()  # clear output
+
+    with bf.option_context(
+        "display.progress_bar", "terminal", "bigquery.allow_large_results", "True"
+    ):
+        penguins_df_default_index["body_mass_g"].head(10).mean()
+
     assert_loading_msg_exist(capsys.readouterr().out)
```
tswast, line 71: I'm curious about this test. Note that we won't always have a "job" object, so this seems like it could end up being flakey.

shobsi:
Waiting for https://github.com/googleapis/python-bigquery-dataframes/pull/1545 to go in first to fix the benchmarking script failure.

tests/system/small/test_progress_bar.py
shobsi: won't always have job object even after explicitly setting `bigquery.allow_large_results=True`?

tswast: Ah, I missed that.

tswast (approved):

PR #1542, opened 2025-03-25
feat!: drop support for locational endpoints

shobsi:
BREAKING CHANGE: Locational endpoints support is not available in BigFrames 2.0.

- [x] Appropriate docs were updated - screen/74LbU3BHHXHz63C

Fixes internal issue 379918500 🦕


shuoweil (approved):

bigframes/_config/bigquery_options.py
```diff
@@ -265,20 +265,16 @@ def use_regional_endpoints(self) -> bool:
             in regions "europe-west3", "europe-west8", "europe-west9",
             "me-central2", "us-central1", "us-central2", "us-east1", "us-east4",
             "us-east5", "us-east7", "us-south1", "us-west1", "us-west2", "us-west3"
```
tswast, line 267: Looks like we cut off a little too much.

```suggestion
            "us-east5", "us-east7", "us-south1", "us-west1", "us-west2", "us-west3"
            and "us-west4".
```

bigframes/constants.py
```diff
@@ -96,22 +96,19 @@
...
     Support for regional endpoints is not yet available in the location
     {location} for BigQuery and BigQuery Storage APIs. For the supported
     locations and APIs see https://cloud.google.com/bigquery/docs/regional-endpoints.
-    For other locations and APIs, currently an older, now deprecated locational
-    endpoints are being used, which requires your project to be allowlisted. In
-    future version 2.0 onwards the locational endpoints will no longer be
-    supported automatically when you enable regional endpoints. However, if you
-    still need them, you will be able to override the endpoints directly by
-    doing the following:
+    For other locations we are falling back to the global APIs. If you have the
```
tswast, line 108: Please don't. I'd rather we fail than fall back to global APIs. Otherwise folks won't know if they are using an unsupported location.

bigframes/session/clients.py
```diff
@@ -104,12 +101,12 @@ def __init__(
...
         ):
             msg = bfe.format_message(
-                bigframes.constants.LEP_DEPRECATION_WARNING_MESSAGE.format(
+                bigframes.constants.REP_NOT_SUPPOERTED_WARNING_MESSAGE.format(
                     location=location
                 ),
                 fill=False,
             )
-            warnings.warn(msg, category=FutureWarning)
+            warnings.warn(msg, category=ResourceWarning)
```
tswast, line 109: This should be an exception, not a warning.

If we want to make sure that we support new regions as they roll out, the alternative is to warn that this might not work but still use the regional endpoint and let the server side do the failing. I'd prefer that alternative, actually.

tswast (changes requested):

bigframes/session/clients.py
tswast: Resolving. I've implemented the alternative. Please test that it works.

tests/system/large/test_location.py
```diff
@@ -172,7 +167,5 @@ def test_bq_lep_endpoints(bigquery_location):
     assert clients_provider.bqclient.location == bigquery_location
     assert (
         clients_provider.bqclient._connection.API_BASE_URL
-        == "https://{location}-bigquery.googleapis.com".format(
-            location=bigquery_location
-        )
+        == "https://bigquery.googleapis.com"
```
tswast, line 170: We should never connect to the global URL if the user requested regional endpoints.

shobsi: Done!

bigframes/session/clients.py
```diff
@@ -195,12 +188,6 @@ def bqconnectionclient(self):
                 bqconnection_options = google.api_core.client_options.ClientOptions(
                     api_endpoint=self._client_endpoints_override["bqconnectionclient"]
                 )
-            elif self._use_regional_endpoints:
```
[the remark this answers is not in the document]
shobsi: REP is not yet implemented for BQ Connection API. I am adding clarification in the `use_regional_endpoints` documentation that it applies only to the BQ and BQ Storage  APIs.

bigframes/constants.py
shobsi: Throwing the exception. Thanks for the commit.

bigframes/_config/bigquery_options.py
shobsi: Thanks for catch!

tswast (approved):
Thanks!

PR #1554, opened 2025-03-27
feat: support bigquery connection in managed function

jialuoo:
b/404568828.


tests/system/large/functions/test_managed_function.py
```diff
@@ -166,13 +180,10 @@ def featurize(x: int) -> list[float]:
...
 
-def test_managed_function_series_apply(
-    session,
-    scalars_dfs,
-):
+def test_managed_function_series_apply(session, scalars_dfs, bq_cf_connection):
     try:
 
-        @session.udf()
+        @session.udf(bigquery_connection=bq_cf_connection)
```
shobsi, line 186: we don't have to pass it in all the tests, just special dedicated test, say `test_managed_function_with_connection` should suffice.

bigframes/functions/_function_session.py
```diff
@@ -907,6 +907,9 @@ def wrapper(func):
                 name=name,
                 packages=packages,
                 is_row_processor=is_row_processor,
+                # If bigquery_connection is not provided, bq_connection_id can
+                # be ignored in provision_bq_managed_function.
+                bq_connection_id=bq_connection_id if bigquery_connection else None,
```
shobsi, line 912: It would be tidier to resolve conditionally at line#810:

```python
        # A connection is optional for BQ managed function.
        bq_connection_id = self._resolve_bigquery_connection_id(
            session, dataset_ref, bq_location, bigquery_connection
        ) if bigquery_connection else None
```

and then simply pass `bq_connection_id=bq_connection_id` here.

tests/system/large/functions/test_managed_function.py
shobsi: Sent a patch to update.

bigframes/functions/_function_session.py
shobsi: Updated in a patch

shobsi (approved):

PR #1556, opened 2025-03-27
feat: add `on`  parameter in `dataframe.rolling()` and `dataframe.groupby.rolling()`

sycai:
I didn't add this parameter for Series because it does seem to make much sense: Series doesn't have columns.


third_party/bigframes_vendored/pandas/core/generic.py
```diff
@@ -1083,6 +1085,16 @@ def rolling(
                 For a window that is specified by an integer, ``min_periods`` will default
                 to the size of the window.
 
+            on (str, optional):
+                For a DataFrame, a column label or Index level on which to calculate the rolling window,
```
chelsea-lin, line 1089: Can you try the "Index level" cases here?

sycai: Good catch. I removed index level here because I do not plan to support that at this time

third_party/bigframes_vendored/pandas/core/generic.py
```diff
@@ -1052,6 +1052,8 @@ def rolling(
         self,
         window,
         min_periods: int | None = None,
+        on: str | None = None,
+        closed: Literal["right", "left", "both", "neither"] = "right",
```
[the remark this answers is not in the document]
sycai: Right. Doc added

chelsea-lin (approved):
LGTM overall with two nit comments. Thanks!

third_party/bigframes_vendored/pandas/core/groupby/__init__.py
```diff
@@ -1035,6 +1035,16 @@ def rolling(self, *args, **kwargs):
                 For a window that is specified by an integer, ``min_periods`` will default
                 to the size of the window.
 
+            on (str, optional):
+                For a DataFrame, a column label or Index level on which to calculate the rolling window,
```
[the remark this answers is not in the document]
sycai: Right.

TrevorBergeron (changes requested):

bigframes/core/window_spec.py
```diff
@@ -187,16 +187,20 @@ def __post_init__(self):
...
+        ordering: A list of columns ids and ordering direction to override base ordering
+        min_periods: The minimum number of observations in window required to have a value
+        on: The id of the column on which to calculate the rolling window
     """
 
     grouping_keys: Tuple[ex.DerefOp, ...] = tuple()
     ordering: Tuple[orderings.OrderingExpression, ...] = tuple()
     bounds: Union[RowsWindowBounds, RangeWindowBounds, None] = None
     min_periods: int = 0
+    on: str | None = None
```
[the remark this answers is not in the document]
sycai: Parameter removed. PTAL

bigframes/core/window/rolling.py
```diff
@@ -34,12 +34,16 @@ def __init__(
...
+        skip_agg_column_id: str | None = None,
     ):
         self._block = block
         self._window_spec = window_spec
         self._value_column_ids = value_column_ids
         self._drop_null_groups = drop_null_groups
         self._is_series = is_series
+        # The column ID that won't be aggregated on.
+        # This is equivalent to pandas `on` parameter in rolling()
+        self._skip_agg_column_id = skip_agg_column_id
```
TrevorBergeron, line 46: I think this will eventually be a `set` in order to handle multiple columns for `on`?

TrevorBergeron (approved):

bigframes/core/window/rolling.py
sycai: That would be unlikely. Both pandas and SQL allow only one column to be used for `on` (or its equivalent)

TrevorBergeron: Hmm, maybe not for `on`, but for grouping columns. We diverge a bit right now I think for some grouping cases. Anyways, can always change later if need be.

PR #1557, opened 2025-03-27
docs: add warning for bigframes 2.0

Genesis929:

Staging result: https://screenshot.googleplex.com/4iMfjL9VsB7ZXy5


README.rst
```diff
@@ -3,6 +3,21 @@ BigQuery DataFrames
 
 |GA| |pypi| |versions|
 
+.. raw:: html
```
shobsi, line 6: Can't we write it in markdown format like the rest of the file?

Genesis929: This is not a markdown file I assume, .. note:: and ..warning:: doesn't work as expected, that's the reasson I'm using html.

shobsi: Here is what I got from Gemini

```
**⚠️ Breaking Changes in BigQuery DataFrames v2.0**

Version 2.0 introduces breaking changes for improved security and performance. Key default behaviors have changed:

* **Large Results (>10GB):** The default value for `allow_large_results` has changed to False. Methods like `to_pandas()` will now fail if the query result's compressed data size exceeds 10GB, unless large results are explicitly permitted.
* **Remote Function Security:** The library no longer defaults to using the Compute Engine service account, and network ingress now defaults to "internal-only".
* **Endpoint Connections:** Automatic fallback to locational endpoints in certain regions is removed.

**Important:** If you are not ready to adapt to these changes, please pin your dependency to a version less than 2.0 (e.g., `bigframes==1.38.0`) to avoid disruption.

To learn about these changes and how to migrate to version 2.0, see: [updated introduction guide](https://cloud.google.com/bigquery/docs/bigquery-dataframes-introduction).
```

README.rst
```diff
@@ -3,6 +3,21 @@ BigQuery DataFrames
...
+
+    <div style="border: 1px solid #f5c6cb;
+    background-color: #f8d7da; color: #721c24; padding: 15px; margin-bottom: 20px; border-radius: 4px; font-size: 90%;">
+        <p style="margin-top: 0; font-weight: bold;">⚠️ Breaking Changes in BigQuery DataFrames v2.0</p>
+        <p>Version 2.0 introduces breaking changes for improved security and performance. Key default behaviors have changed:</p>
+        <ul>
+            <li><strong>Large Results (&gt;10GB):</strong> The default value for <code>allow_large_results</code> has changed to False. Methods like <code>to_pandas()</code> will now fail if the query result's compressed data size exceeds 10GB, unless large results are explicitly permitted.</li>
+            <li><strong>Remote Function Security:</strong> The library no longer defaults to using the Compute Engine service account, and network ingress now defaults to "internal-only".</li>
```
shobsi, line 14: [nit] suggest rephrasing:

The library no longer automatically lets the Compute Engine default service account become the identity of the Cloud Run functions. If that is desired, it has to be indicated by passing `cloud_function_service_account="default"`.

In addition, we should also call out:

Arguments to `remote_function` would have to be passed as keyword arguments.

README.rst
```diff
@@ -3,6 +3,21 @@ BigQuery DataFrames
...
+            <li><strong>Endpoint Connections:</strong> Automatic fallback to locational endpoints in certain regions is removed. </li>
+        </ul>
+        <p><strong>Important:</strong> If you are not ready to adapt to these changes, please pin your dependency to a version less than 2.0 (e.g., <code>bigframes==1.38.0</code>) to avoid disruption.</p>
```
shobsi, line 17: we should say 1.42.0, as that is the latest release

README.rst
shobsi: rst and md both are markup languages designed for readability, rst tends to be more feature rich than md. The above suggestion should work for both.

Genesis929: Tried similar solutions before, this won't make it similar enough to what we get now.

shobsi: discussed offline that even though in the GitHub the rendering may look same, the raw html gives better appearance (warning background color) when rendered in the cloud.google.com docs.

README.rst
Genesis929: Updated, also added a new line for keyword arguments.

README.rst
Genesis929: Updated.

README.rst
```diff
@@ -12,6 +12,23 @@ powered by the BigQuery engine.
...
+        <p style="margin-top: 0; font-weight: bold;">⚠️ Breaking Changes in BigQuery DataFrames v2.0</p>
+        <p>Version 2.0 introduces breaking changes for improved security and performance. Key default behaviors have changed:</p>
+        <ul>
+            <li><strong>Large Results (&gt;10GB):</strong> The default value for <code>allow_large_results</code> has changed to False. Methods like <code>to_pandas()</code> will now fail if the query result's compressed data size exceeds 10GB, unless large results are explicitly permitted.</li>
+            <li><strong>Remote Function Security:</strong>The library no longer automatically lets the Compute Engine default service account become the identity of the Cloud Run functions. If that is desired, it has to be indicated by passing cloud_function_service_account="default". And network ingress now defaults to "internal-only".</li>
+            <li><strong>@remote_function Argument Passing:</strong> Arguments to remote_function must now be passed using keyword syntax, as positional arguments are no longer supported.</li>
```
tswast, line 24: I have https://github.com/googleapis/python-bigquery-dataframes/pull/1560 open to partially revert this change.

```suggestion
            <li><strong>@remote_function Argument Passing:</strong> Arguments other than input_types, output_type, and dataset to remote_function must now be passed using keyword syntax, as positional arguments are no longer supported.</li>
```

tswast (approved):

PR #1561, opened 2025-03-28
[WIP] Log __init__ calls of timedelta-related ops

sycai:


TrevorBergeron (changes requested):
Can we please not log internal op initializers, or really anything not directly tied to user intentions or server requests? I worry this is very fragile.

sycai:
> Can we please not log internal op initializers, or really anything not directly tied to user intentions or server requests? I worry this is very fragile.

I need to gather metrics on the usage of timedelta ops, and it seems this is the most straightforward approach. Could you elaborate on the fragility?

sycai:
Talked with @TrevorBergeron: the concern at the moment is that this PR is trying to log internal mechanisms instead of user intentions, which deviates from the purpose of the `class_logger`. 

Putting this PR on hold for now.

PR #1563, opened 2025-03-28
docs: add message to remove default model for version 3.0

shuoweil:
add message to remove default model for version 3.0


shobsi (approved):

tswast:
forced the merge since this is a readme only change.

PR #1568, opened 2025-03-28
fix: prevent `KeyError` in `bpd.concat` with empty DF and struct/array types DF

chelsea-lin:
Fixes internal issue 407107482 🦕


GarrettWu:
Thanks for the fix.

GarrettWu (approved):

PR #1569, opened 2025-03-28
feat: Use session temp tables for all ephemeral storage

TrevorBergeron:


tests/system/small/test_bq_sessions.py
```diff
@@ -57,8 +58,13 @@ def test_bq_session_create_temp_table_clustered(bigquery_client: bigquery.Client
     assert result_table.clustering_fields == cluster_cols
 
     session_resource_manager.close()
-    with pytest.raises(google.api_core.exceptions.NotFound):
```
tswast, line 60: FWIW: if you sync to main I made a similar fix in https://github.com/googleapis/python-bigquery-dataframes/pull/1572/files. I think we can keep the `pytest.raises` and just do the loop.

TrevorBergeron: ok, just merged in and used your version

bigframes/session/loader.py
```diff
@@ -166,8 +171,13 @@ def read_pandas_load_job(
         job_config.clustering_fields = cluster_cols
 
         job_config.labels = {"bigframes-api": api_name}
+        job_config.schema_update_options = [
+            google.cloud.bigquery.job.SchemaUpdateOption.ALLOW_FIELD_ADDITION
```
tswast, line 175: I'm curious: why `ALLOW_FIELD_ADDITION` here, but using `WRITE_TRUNCATE` for `read_gbq_load_job`? Might be worthwhile to add some comments.

bigframes/session/loader.py
```diff
@@ -525,21 +538,12 @@ def _read_bigquery_load_job(
...
         else:
             load_job = self._bqclient.load_table_from_file(
-                filepath_or_buffer, table, job_config=job_config
+                filepath_or_buffer, destination=table, job_config=job_config
             )
 
         self._start_generic_job(load_job)
         table_id = f"{table.project}.{table.dataset_id}.{table.table_id}"
 
-        # Update the table expiration so we aren't limited to the default 24
```
tswast, line 534: ~I assume `_storage_manager.create_temp_table` handles this in the anonymous dataset case now?~

Edit: Yes, I see we do set an expiration above:

```
        expiration = (
            datetime.datetime.now(datetime.timezone.utc) + constants.DEFAULT_EXPIRATION
        )
        table = bf_io_bigquery.create_temp_table(
            self.bqclient,
            self.allocate_temp_table(),
            expiration,
            schema=schema,
            cluster_columns=list(cluster_cols),
            kms_key=self._kms_key,
        )
```

tswast (approved):
Thanks! Just a few nits.

bigframes/session/__init__.py
```diff
@@ -388,9 +401,13 @@ def close(self):
...
-        temp_storage_manager = getattr(self, "_temp_storage_manager", None)
-        if temp_storage_manager:
-            self._temp_storage_manager.clean_up_tables()
+        anon_dataset_manager = getattr(self, "_anon_dataset_manager", None)
+        if anon_dataset_manager:
+            self._anon_dataset_manager.close()
+
+        if getattr(self, "_session_resource_manager", None):
+            if self._session_resource_manager is not None:
+                self._session_resource_manager.close()
```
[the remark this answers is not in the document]
tswast: Committed my own suggestion. 🤞 hopefully I didn't break anything.

bigframes/session/loader.py
TrevorBergeron: I just want to make sure the ordering_col does not get overridden, as that is what is clustered.
Might still work with TRUNCATE as well?

PR #1570, opened 2025-03-28
docs: deprecate default model in `TextEmbedddingGenerator`, `GeminiTextGenerator`, and other `bigframes.ml.llm` classes

shuoweil:
docs: add remove default model warning, default LLM model will be removed for Bigframes 3.0

### DIFF
--- a/bigframes/ml/llm.py
+++ b/bigframes/ml/llm.py
@@ -21,7 +21,6 @@ import warnings
 
 import bigframes_vendored.constants as constants
 from google.cloud import bigquery
-import typing_extensions
 
 from bigframes import dtypes, exceptions
 import bigframes.bigquery as bbq
@@ -136,6 +135,10 @@ class TextEmbeddingGenerator(base.RetriableRemotePredictor):
         connection_name: Optional[str] = None,
     ):
         self.model_name = model_name
+        if model_name == "text-embedding-004":
+            msg = exceptions.format_message(_REMOVE_DEFAULT_MODEL_WARNING)
+            warnings.warn(msg, category=FutureWarning, stacklevel=2)
+
         self.session = session or global_session.get_global_session()
         self.connection_name = connection_name
 
--- a/bigframes/ml/llm.py
+++ b/bigframes/ml/llm.py
@@ -127,15 +127,18 @@ class TextEmbeddingGenerator(base.RetriableRemotePredictor):
     def __init__(
         self,
         *,
-        model_name: Literal[
-            "text-embedding-005",
-            "text-embedding-004",
-            "text-multilingual-embedding-002",
-        ] = "text-embedding-004",
+        model_name: Optional[
+            Literal[
+                "text-embedding-005",
+                "text-embedding-004",
+                "text-multilingual-embedding-002",
+            ]
+        ] = None,
         session: Optional[bigframes.Session] = None,
         connection_name: Optional[str] = None,
     ):
-        if model_name == _TEXT_EMBEDDING_004_ENDPOINT:
+        if model_name is None:
+            model_name = "text-embedding-004"
             msg = exceptions.format_message(_REMOVE_DEFAULT_MODEL_WARNING)
             warnings.warn(msg, category=FutureWarning, stacklevel=2)
         self.model_name = model_name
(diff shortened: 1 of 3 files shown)


bigframes/ml/llm.py
```diff
@@ -103,7 +104,13 @@
     "You should use this model name only if you are sure that it is supported in BigQuery."
 )
 
+_REMOVE_DEFAULT_MODEL_WARNING = "Due to the frequent deprecation of LLM models, the default model will be removed in BigFrames 3.0."
```
tswast, line 107: We can rephrase this a bit to put less blame on the Vertex team. Perhaps:

```suggestion
_REMOVE_DEFAULT_MODEL_WARNING = "Since upgrading the default model can cause unintended breakages, the default model will be removed in BigFrames 3.0. Please supply an explicit model to avoid this message."
```

bigframes/ml/llm.py
```diff
@@ -103,7 +104,13 @@
...
 
+
+@typing_extensions.deprecated(
+    _REMOVE_DEFAULT_MODEL_WARNING,
+    category=exceptions.DefaultModelRemovalWarning,
+)
```
tswast, line 113: Don't always raise this. Users should be able to avoid the warning by supplying a model name directly.

tswast: See Chelsea's change at https://github.com/googleapis/python-bigquery-dataframes/pull/1419 for an example of warning when the user doesn't explicitly provide an argument.

```
        cloud_function_ingress_settings: Optional[
            Literal["all", "internal-only", "internal-and-gclb"]
        ] = None,
    ):
        """Decorator to turn a user defined function into a BigQuery remote function.
	@@ -311,8 +311,9 @@ def remote_function(
                https://cloud.google.com/functions/docs/configuring/memory.
            cloud_function_ingress_settings (str, Optional):
                Ingress settings controls dictating what traffic can reach the
                function. Options are: `all`, `internal-only`, or `internal-and-gclb`.
                If no setting is provided, `all` will be used by default and a warning
                will be issued. See for more details
                https://cloud.google.com/functions/docs/networking/network-settings#ingress_settings.
        """
        # Some defaults may be used from the session if not provided otherwise
	@@ -429,6 +430,16 @@ def remote_function(
                " For more details see https://cloud.google.com/functions/docs/securing/cmek#before_you_begin"
            )

        if cloud_function_ingress_settings is None:
            cloud_function_ingress_settings = "all"
            msg = (
                "The `cloud_function_ingress_settings` are set to 'all' by default, "
                "which will change to 'internal-only' for enhanced security in future version 2.0 onwards. "
                "However, you will be able to explicitly pass cloud_function_ingress_settings='all' if you need. "
                "See https://cloud.google.com/functions/docs/networking/network-settings#ingress_settings for details."
            )
            warnings.warn(msg, category=FutureWarning, stacklevel=2)
```

shuoweil:
failing notebook is not relevant to this change, please help me with the review

bigframes/ml/llm.py
```diff
@@ -132,6 +135,9 @@ def __init__(
         session: Optional[bigframes.Session] = None,
         connection_name: Optional[str] = None,
     ):
+        if model_name == _TEXT_EMBEDDING_004_ENDPOINT:
+            msg = exceptions.format_message(_REMOVE_DEFAULT_MODEL_WARNING)
+            warnings.warn(msg, category=FutureWarning, stacklevel=2)
```
tswast, line 140: No. If they explicitly set this, we shouldn't show a warning.

Try something like this instead:

```suggestion
        model_name: Optional[Literal[
            "text-embedding-005",
            "text-embedding-004",
            "text-multilingual-embedding-002",
        ]] = None,
        session: Optional[bigframes.Session] = None,
        connection_name: Optional[str] = None,
    ):
        if model_name is None:
            model_name = _TEXT_EMBEDDING_004_ENDPOINT:
            msg = exceptions.format_message(_REMOVE_DEFAULT_MODEL_WARNING)
            warnings.warn(msg, category=FutureWarning, stacklevel=2)
```

tswast (changes requested):
FYI: I updated the PR title to reflect what we're doing in a way the user would understand.

bigframes/ml/llm.py
tswast: Please apply the same for the other models, too.

shuoweil: Thanks Tim, I used to believe we plan to remove default model for Bigframes 3.0, and throw an warning for default model removal in Bigframes 2.0. If I apply the same change, we are removing default model in bigframes 2.0. Please advise should I remove default models in Bigframes 2.0, or should we wait until 3.0 release time? Thanks a lot.

tswast (commented):
Looks good, thank you!

Let's make sure these code paths are covered by tests, such as by adding unit tests that check that the warning is issued at construction time.

For example:

https://github.com/googleapis/python-bigquery-dataframes/blob/b03e44f7fca429a6de41c42ec28504b688cd84f0/tests/unit/ml/test_matrix_factorization.py#L49-L65

but in this case, use `pytest.warns` instead.

shobsi:
May I suggest the title:

chore: warn if default llm is used in `bigframes.ml.llm`  classes

shuoweil:
> chore: warn if default llm is used in `bigframes.ml.llm` classes

Thanks @shobsi for the suggestion. I believe both of the title looks great. Adding @tswast for discussion.

tswast:
@shobsi We don't want  to use `chore: ` because we want this to be in the changelog. `chore: ` should be reserved for changes that don't affect user-facing behaviors like refactors and test infrastructure.

tswast (approved):

shobsi:
> @shobsi We don't want to use `chore: ` because we want this to be in the changelog. `chore: ` should be reserved for changes that don't affect user-facing behaviors like refactors and test infrastructure.

Ack, `feat:` would make more sense, but the main point was to shorten the long title

PR #1573, opened 2025-03-31
feat: warn the deprecated `max_download_size`, `random_state` and `sampling_method` parameters in `(DataFrame|Series).to_pandas()`

chelsea-lin:
Fixes internal issue 391676515 🦕

### DIFF
--- a/bigframes/dataframe.py
+++ b/bigframes/dataframe.py
@@ -1714,14 +1714,14 @@ class DataFrame(vendored_pandas_frame.DataFrame):
                 "DEPRECATED: The `max_download_size` parameters for `DataFrame.to_pandas()` "
                 "are deprecated and will be removed soon. Please use `DataFrame.to_pandas_batches()`."
             )
-            warnings.warn(msg, category=UserWarning)
+            warnings.warn(msg, category=FutureWarning)
         if sampling_method is not None or random_state is not None:
             msg = bfe.format_message(
                 "DEPRECATED: The `sampling_method` and `random_state` parameters for "
                 "`DataFrame.to_pandas()` are deprecated and will be removed soon. "
                 "Please use `DataFrame.sample().to_pandas()` instead for sampling."
             )
-            warnings.warn(msg, category=UserWarning, stacklevel=2)
+            warnings.warn(msg, category=FutureWarning, stacklevel=2)
 
         if dry_run:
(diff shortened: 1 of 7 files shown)


bigframes/dataframe.py
```diff
@@ -1636,17 +1636,27 @@ def to_pandas(
 
         Args:
             max_download_size (int, default None):
+                .. deprecated:: 2.0.0
+                    `max_download_size` parameter is deprecated. Please use `to_pandas_batch()` method
```
tswast, line 1640: Nit: RST uses two backticks for code format. One is italics.

```suggestion
                    ``max_download_size`` parameter is deprecated. Please use ``to_pandas_batches()`` method
```

bigframes/dataframe.py
```diff
@@ -1666,8 +1676,19 @@ def to_pandas(
                 downsampled rows and all columns of this DataFrame. If dry_run is set, a pandas
                 Series containing dry run statistics will be returned.
         """
-
-        # TODO(orrbradford): Optimize this in future. Potentially some cases where we can return the stored query job
+        if max_download_size is not None:
+            msg = bfe.format_message(
+                "DEPRECATED: The `max_download_size` parameters for `DataFrame.to_pandas()` "
+                "are deprecated and will be removed soon. Please use `DataFrame.to_pandas_batch()`."
```
tswast, line 1682:
```suggestion
                "are deprecated and will be removed soon. Please use ``DataFrame.to_pandas_batches()``."
```

tswast (commented):
Looks like theres some typos in the docstrings. Please fix.

bigframes/dataframe.py
chelsea-lin: Good catches. Fixed!

tswast (approved):
Thanks!

I added a commit to use FutureWarning instead of UserWarning. See: https://docs.python.org/3/library/exceptions.html#FutureWarning

Per https://stackoverflow.com/a/55378483/101923 both DeprecationWarning and [PendingDeprecationWarning](https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning) are ignored by default, so FutureWarning is the most appropriate for bigframes which is used in notebooks and won't necessarily have pytest test code or similar where the deprecation warnings would be seen.

PR #1575, opened 2025-03-31
chore: setup branch policy for v1

shobsi:

### DIFF
--- a/.github/sync-repo-settings.yaml
+++ b/.github/sync-repo-settings.yaml
@@ -3,7 +3,7 @@
 branchProtectionRules:
 # Identifies the protection rule pattern. Name of the branch to be protected.
 # Defaults to `main`
-- pattern: main
+- pattern: main|v1
   requiresCodeOwnerReviews: true
   requiresStrictStatusChecks: false
   requiredStatusCheckContexts:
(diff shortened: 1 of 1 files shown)


tswast (approved):

.github/sync-repo-settings.yaml
```diff
@@ -3,7 +3,7 @@
 branchProtectionRules:
 # Identifies the protection rule pattern. Name of the branch to be protected.
 # Defaults to `main`
-- pattern: main
```
tswast, line 6: If this is regex, might have to be (main)|(v1)

shobsi: I did a bit of digging up, this seems to be the code responsible for applying the settings in github: https://github.com/googleapis/repo-automation-bots/blob/49f8e698e1a22c71e98371adc03bd69e4e515d1a/packages/sync-repo-settings/src/sync-repo-settings.ts#L175

It calls octokit REST API which clearly says to not use wildcard screen/A79AhsXhEcbbWQQ

So I wonder why the field is called "pattern" (reason for me thinking its a regex), but changed to having v1 branch protection in its own section. PTAL.

PR #1578, opened 2025-04-01
chore: move multimodal tests from PROD to TEST

shuoweil:
chore: move multimodal tests from PROD to TEST
b/406325742

### DIFF
--- a/tests/system/conftest.py
+++ b/tests/system/conftest.py
@@ -110,6 +110,11 @@ def bigquery_client_tokyo(session_tokyo: bigframes.Session) -> bigquery.Client:
 
 
 @pytest.fixture(scope="session")
+def bigquery_client_test(test_session: bigframes.Session) -> bigquery.Client:
+    return test_session.bqclient
+
+
+@pytest.fixture(scope="session")
 def ibis_client(session: bigframes.Session) -> ibis_backends.BaseBackend:
     return session.ibis_client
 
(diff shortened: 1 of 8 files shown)


tests/system/small/ml/test_llm.py
GarrettWu, line 1: Don't move all the tests, only those MM ones. You can create a separate file.

tests/system/conftest.py
```diff
@@ -1463,18 +1491,18 @@ def images_gcs_path() -> str:
 @pytest.fixture(scope="session")
 def images_uris() -> list[str]:
     return [
-        "gs://bigframes_blob_test/images/img0.jpg",
         "gs://bigframes_blob_test/images/img1.jpg",
+        "gs://bigframes_blob_test/images/img0.jpg",
```
GarrettWu, line 1495: Why change this order? Then you need to modify all the test orders.

shuoweil: I do not want to move this order intentionally. When I create the test session, the order is changed when executing the test.

shuoweil: fix

tests/system/small/ml/test_llm.py
shuoweil: fix

tests/system/conftest.py
```diff
@@ -109,6 +109,11 @@ def bigquery_client_tokyo(session_tokyo: bigframes.Session) -> bigquery.Client:
     return session_tokyo.bqclient
 
 
+@pytest.fixture(scope="session")
```
GarrettWu, line 112: where do you need this?

GarrettWu (approved):

tests/system/conftest.py
shuoweil: This is one is not used anyway, it is a left over for a prior check. I have removed it.

PR #1583, opened 2025-04-02
chore: fix blob functions missed in cleanup set

GarrettWu:

### DIFF
--- a/bigframes/blob/_functions.py
+++ b/bigframes/blob/_functions.py
@@ -102,6 +102,9 @@ AS r\"\"\"
     def udf(self):
         """Create and return the UDF object."""
         udf_name = self._create_udf()
+
+        # TODO(b/404605969): remove cleanups when UDF fixes dataset deletion.
+        self._session._function_session._update_temp_artifacts(udf_name, "")
         return self._session.read_gbq_function(udf_name)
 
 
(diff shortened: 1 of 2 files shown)


shuoweil:
Those code are existing in main, may I ask why we need to add them again? See here for an example: https://github.com/googleapis/python-bigquery-dataframes/blob/b5297f909b08928b97d887764d6e5142c763a5a3/bigframes/operations/blob.py#L416

GarrettWu:
> Those code are existing in main, may I ask why we need to add them again? See here for an example:
> 
> https://github.com/googleapis/python-bigquery-dataframes/blob/b5297f909b08928b97d887764d6e5142c763a5a3/bigframes/operations/blob.py#L416

Those are different paths that were missed.

shobsi (commented):
Strongly recommend adding it in the `TransformFunction.udf` itself, for reduced technical debt and manual errors

```python
    def udf(self):
        """Create and return the UDF object."""
        udf_name = self._create_udf()
        self._session._function_session._update_temp_artifacts(udf_name, "")
        return self._session.read_gbq_function(udf_name)
```

Then we can save the entire `_add_to_cleanup_set` business in the blob.py

GarrettWu:
> Strongly recommend adding it in the `TransformFunction.udf` itself, for reduced technical debt and manual errors
> 
> ```python
>     def udf(self):
>         """Create and return the UDF object."""
>         udf_name = self._create_udf()
>         self._session._function_session._update_temp_artifacts(udf_name, "")
>         return self._session.read_gbq_function(udf_name)
> ```
> 
> Then we can save the entire `_add_to_cleanup_set` business in the blob.py

Great! Done.

shuoweil (approved):

shobsi (approved):

PR #1588, opened 2025-04-03
fix: ensure `page_size` works correctly in `to_pandas_batches` when `max_results` is not set

chelsea-lin:
This change fixes an issue that the `page_size` is ignored in `to_pandas_batches` when `max_results` is not set. Also, this change adds code examples for both `to_pandas` and `to_pandas_batches` and the doctest can verify the fixes.

Fixes internal issue 407521010🦕


shuoweil (approved):

tswast (approved):
Thanks!

PR #1590, opened 2025-04-03
feat: support time range rolling on Series.

sycai:


bigframes/core/compile/compiled.py
```diff
@@ -579,6 +593,19 @@ def _convert_ordering_to_table_values(
     return ordering_values
 
 
+def _convert_range_ordering_to_table_value(
```
TrevorBergeron, line 596: what makes this one different? is it that it doesn't allow NULLS FIRST/LAST overrides and only allows one column?

bigframes/core/compile/compiled.py
```diff
@@ -668,8 +695,8 @@ def _add_boundary(
 ) -> ibis_expr_builders.LegacyWindowBuilder:
     if isinstance(bounds, RangeWindowBounds):
         return ibis_window.range(
-            start=_to_ibis_boundary(bounds.start),
-            end=_to_ibis_boundary(bounds.end),
+            start=_to_ibis_boundary(bounds.start_micros),
+            end=_to_ibis_boundary(bounds.end_micros),
```
TrevorBergeron, line 699: Why define as micros? Sure, now its always micros, but not really a necessary property.

bigframes/core/rewrite/windows.py
```diff
@@ -0,0 +1,45 @@
...
+
+from __future__ import annotations
+
+import dataclasses
+
+from bigframes import operations as ops
+from bigframes.core import nodes
+
+
+def rewrite_range_rolling(node: nodes.BigFrameNode) -> nodes.BigFrameNode:
```
TrevorBergeron, line 23: I know I always propose compile-time rewrites, but I think validations need to happen earlier, in the API. The tree should always be valid, rewrites just transform from one valid tree to another.

bigframes/core/window/rolling.py
```diff
@@ -118,3 +124,81 @@ def _aggregate_block(
...
+        window = pandas.Timedelta(window)
+    spec = window_spec.WindowSpec(
+        bounds=window_spec.RangeWindowBounds.from_timedelta_window(window, closed),
+        min_periods=1 if min_periods is None else min_periods,
+        ordering=(
+            ordering.OrderingExpression(
+                ex.deref(block.index_columns[0]), order_direction
+            ),
+        ),
+    )
```
TrevorBergeron, line 158: I do like this where we are picking the ordering column immediately and validating

bigframes/core/window/rolling.py
```diff
@@ -118,3 +124,81 @@ def _aggregate_block(
...
+    return Window(block, spec, block.value_columns, is_series=is_series)
+
+
+@singledispatch
+def _find_order_direction(
```
TrevorBergeron, line 163: nit: probably should have tree stuff elsewhere, but whatever

bigframes/core/window/rolling.py
```diff
@@ -118,3 +124,81 @@ def _aggregate_block(
...
+    """Returns the order of the given column with tree traversal. If the column cannot be found,
+    or the ordering information is not available, return None.
+    """
+    return None
+
+
+@_find_order_direction.register
+def _(root: nodes.OrderByNode, column_id: str):
+    if len(root.by) == 0:
+        return None
```
TrevorBergeron, line 175: in this case, the node is a no-op so we can just call child

bigframes/core/window/rolling.py
```diff
@@ -118,3 +124,81 @@ def _aggregate_block(
...
+@_find_order_direction.register
+def _(root: nodes.SelectionNode, column_id: str):
+    for alias_ref in root.input_output_pairs:
+        if alias_ref.id.name == column_id:
+            return _find_order_direction(root.child, alias_ref.ref.id.name)
+
+
+@_find_order_direction.register
+def _(root: nodes.FilterNode, column_id: str):
+    return _find_order_direction(root.child, column_id)
```
TrevorBergeron, line 204: can probably also extend to the additive nodes (projection, window, isin), for which you can just ignore the contents and call child (or you can try to identify strictly increasing functions if you want to get fancy).

TrevorBergeron (commented):
I think you will also want to modify the ordering pull-up logic to not pull ordering into range windows? Should be some similar logic already there.

bigframes/core/compile/compiled.py
sycai: Yeah. If we have to deal with NULLs, there will be multiple expressions after "ORDER BY", but it is not allowed by SQL window syntax.

bigframes/core/compile/compiled.py
sycai: It has something to do with dealing with nullness. Property removed.

bigframes/core/rewrite/windows.py
sycai: Are you talking about the length checking for `ordering`? 

I will check the rolling window key on the API level too. Here this is simply to make sure we fail fast if some other error sneaks in.

bigframes/core/window/rolling.py
sycai: TIL. code updated.

TrevorBergeron: yeah, all re-orderings are stable re-orderings, so it prepends to the existing ordering key rather than replaceing

bigframes/core/window/rolling.py
sycai: Added windowNode and inNode. The ProjectNode is left out because some operations may invalidate ordering (e.g. multiplication with a negative value)

TrevorBergeron: ProjetNode is always additive, it doesn't mutate values, so it should be safe. If anything, it can provide alternative ordering keys if you know an operation is strictly increasing.

bigframes/core/compile/compiled.py
sycai: Added doc that explains the reasons for future references.

bigframes/core/rewrite/windows.py
```diff
@@ -0,0 +1,45 @@
...
+    ordering_expr = node.window_spec.ordering[0]
+
+    new_ordering = dataclasses.replace(
+        ordering_expr,
+        scalar_expression=ops.UnixMicros().as_expr(ordering_expr.scalar_expression),
+    )
+
+    return dataclasses.replace(
+        node,
+        window_spec=dataclasses.replace(node.window_spec, ordering=(new_ordering,)),
```
TrevorBergeron, line 44: it would be consistent to redefine the window spec bounds to integers rather than timestamp to be consistent with the underlying value now being an integer.

bigframes/core/window/rolling.py
sycai: Done!

TrevorBergeron (approved):

bigframes/core/rewrite/windows.py
sycai: I think I still prefer the timedeltas because it describes best what the range windows are for

PR #1595, opened 2025-04-04
refactor: refactor read_csv and tests based on bigquery vs. pandas behavior comparison

chelsea-lin:


bigframes/session/__init__.py
```diff
@@ -910,112 +910,168 @@ def read_csv(
...
                 index_col=index_col,
                 usecols=usecols,  # type: ignore
                 dtype=dtype,
-                engine=engine,
                 encoding=encoding,
-                **kwargs,
             )
-            return self._read_pandas(pandas_df, api_name="read_csv", write_engine=write_engine)  # type: ignore
+
+    def _read_csv_w_pandas_engines(
```
sycai, line 942: naming nit: "_read_csv_w_pandas_engine" ?

bigframes/session/__init__.py
```diff
@@ -910,112 +910,168 @@ def read_csv(
...
+        job_config.encoding = encoding
+        job_config.labels = {"bigframes-api": "read_csv"}
+
+        # We want to match pandas behavior. If header is 0, no rows should be skipped, so we
+        # do not need to set `skip_leading_rows`. If header is None, then there is no header.
+        # Setting skip_leading_rows to 0 does that. If header=N and N>0, we want to skip N rows.
+        if header is None:
+            job_config.skip_leading_rows = 0
+        elif header > 0:
+            job_config.skip_leading_rows = header + 1
```
sycai, line 1067: Non-rhetorical question: why "header + 1" here when the original version is just "header"?

tests/system/small/test_session.py
```diff
@@ -39,6 +39,33 @@
 from tests.system import utils
 
 
+@pytest.fixture(scope="module")
+def write_df_to_local_csv_file(scalars_df_index):
```
sycai, line 43: naming nit: maybe "df_and_local_csv" is better? We are using a lot of noun phrases for fixtures after all

tests/system/small/test_session.py
```diff
@@ -39,6 +39,33 @@
...
+
+    with tempfile.TemporaryDirectory() as dir:
+        # Prepares local CSV file for reading
+        path = dir + "/write_df_to_local_csv_file.csv"
+        scalars_df_index.to_csv(path, index=True)
+        yield scalars_df_index, path
+
+
+@pytest.fixture(scope="module")
+def write_df_to_gcs_csv_file(scalars_df_index, gcs_folder):
```
sycai, line 57: naming nit "df_and_gcs_csv"?

tests/system/small/test_session.py
```diff
@@ -1165,59 +1143,61 @@ def test_read_csv_local_default_engine(session, scalars_dfs, sep):
...
+    assert bf_df.shape[0] == pd_df.shape[0]
 
-    df = session.read_csv(
-        read_path,
-        # Normally, pandas uses the first column as the index. index_col=False
-        # turns off that behavior.
-        index_col=False,
-    )
-    assert df.shape[0] == scalars_df_index.shape[0]
+    if index_col is False:
```
sycai, line 1190: Perhaps we should separate this test method into two. "if-else" branches based on test input are usually a code smell.

go/unit-testing-practices?polyglot=python#logic:

> "Tests written without operators or control structures are clearer since the reader doesn't have to do any mental computations to understand them, and are more likely to be correct since it's harder to have bugs in code without these constructs."

sycai (commented):
Thanks for the clean up!

bigframes/session/__init__.py
chelsea-lin: `Pandas.read_csv` support multiple engines, such as "c", "python", "pyarrow", "python-fwf". Added docstring in the method for easier read.

bigframes/session/__init__.py
chelsea-lin: New refactoring tests compare BigQuery and Pandas behavior and catch the bug mentioned here. However, despite the fix, column naming mismatches still exist, as reported in internal issue 409070192.

tests/system/small/test_session.py
chelsea-lin: Done.

tests/system/small/test_session.py
chelsea-lin: Good points. Done.

tests/system/small/test_session.py
chelsea-lin: Done.

sycai (approved):

PR #1597, opened 2025-04-04
docs: deprecate `bpd.options.bigquery.allow_large_results` in favor of `bpd.options.compute.allow_large_results`

Genesis929:


bigframes/_config/bigquery_options.py
```diff
@@ -89,7 +89,6 @@ def __init__(
         kms_key_name: Optional[str] = None,
         skip_bq_connection_check: bool = False,
         *,
-        allow_large_results: bool = False,
```
tswast, line 92: We already have public documentation including the MSA referencing this option. Let's make this immutable instead and raise a warning that this is deprecated. Direct folks to the compute version in the warning.

tswast (changes requested):

bigframes/_config/bigquery_options.py
Genesis929: Updated

bigframes/_config/bigquery_options.py
```diff
@@ -239,22 +237,43 @@ def skip_bq_connection_check(self, value: bool):
...
+
+        Warning: Setting this option here is deprecated, ignored, and this setter
+        will be removed in a future version. The configuration for handling large
+        results has moved.
+        """
+        warnings.warn(
+            "Setting `bpd.options.bigquery.allow_large_results` is deprecated, ignored, "
+            "and will be removed soon. "
+            "Please use `bpd.options.compute.allow_large_results = <value>` instead.",
+            DeprecationWarning,
```
tswast, line 273:
```suggestion
            FutureWarning,
```

bigframes/_config/bigquery_options.py
```diff
@@ -239,22 +237,43 @@ def skip_bq_connection_check(self, value: bool):
...
-            bool: True if large results are allowed with an explicit destination table,
-            False if results are limited to 10 GB and errors are raised when exceeded.
+            bool: The value of the deprecated setting.
         """
+        warnings.warn(
+            "`bpd.options.bigquery.allow_large_results` is deprecated and will be removed soon. "
+            "Please use `bpd.options.compute.allow_large_results` instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
tswast, line 255: Warning at read time is going to be too noisy.

```suggestion
```

bigframes/_config/bigquery_options.py
```diff
@@ -239,22 +237,43 @@ def skip_bq_connection_check(self, value: bool):
...
+        results has moved.
+        """
+        warnings.warn(
+            "Setting `bpd.options.bigquery.allow_large_results` is deprecated, ignored, "
+            "and will be removed soon. "
+            "Please use `bpd.options.compute.allow_large_results = <value>` instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        pass
```
tswast, line 276: Please don't ignore this option. We should still use it if the compute option is not set. Also, let's make this option immutable-ish like the other BigQuery options.

```suggestion
        if self._session_started and self._allow_large_results != value:
            raise ValueError(
                SESSION_STARTED_MESSAGE.format(attribute="allow_large_results")
            )

        self._allow_large_results = value
```

bigframes/_config/compute_options.py
```diff
@@ -97,7 +103,9 @@ class ComputeOptions:
     semantic_ops_threshold_autofail = False
 
     ai_ops_confirmation_threshold: Optional[int] = 0
-    ai_ops_threshold_autofail = False
+    ai_ops_threshold_autofail: bool = False
+
+    allow_large_results: bool = False
```
tswast, line 108: Please make this an `Optional[bool]` so that we can disambiguate when someone sets it. This should take precedence over the bigquery option, not replace it. It's too late to make a breaking change.

```suggestion
    allow_large_results: Optional[bool] = None
```

bigframes/session/bq_caching_executor.py
```diff
@@ -112,7 +112,7 @@ def execute(
         max_results: Optional[int] = None,
     ) -> executor.ExecuteResult:
         if use_explicit_destination is None:
-            use_explicit_destination = bigframes.options.bigquery.allow_large_results
+            use_explicit_destination = bigframes.options.compute.allow_large_results
```
tswast, line 115: Need to use the bigquery option if the compute option is not set.

```suggestion
            allow_large_results = bigframes.options.compute.allow_large_results
            if allow_large_results is None:
                allow_large_results = bigframes.options.bigquery.allow_large_results
            use_explicit_destination = allow_large_results
```

Since this is going to be used in many places, please make a helper function somewhere in `bigframes._config` for this.

tests/unit/_config/test_compute_options.py
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import bigframes._config.compute_options as compute_options
+
+
+def test_default_options():
+    options = compute_options.ComputeOptions()
+
+    assert options.allow_large_results is False
```
tswast, line 21: Should be `None` because we need a fallback to the bigquery options if it's not set.

tswast (changes requested):

bigframes/session/bq_caching_executor.py
Genesis929: added one in options, not sure if that's the best place though.

bigframes/_config/__init__.py
```diff
@@ -150,6 +150,12 @@ def is_bigquery_thread_local(self) -> bool:
         """
         return self._local.bigquery_options is not None
 
+    @property
+    def _allow_large_results(self) -> bool:
```
tswast, line 154: Even though this is private, please add a docstring explaining the purpose.

bigframes/_config/bigquery_options.py
```diff
@@ -239,21 +237,41 @@ def skip_bq_connection_check(self, value: bool):
...
+        ``bpd.options.compute.allow_large_results`` instead.
+
+        Warning: Setting this option here will be deprecated, and this setter
+        will be removed in a future version. The configuration for handling large
+        results has moved.
+        """
+        warnings.warn(
+            "Setting `bpd.options.bigquery.allow_large_results` is deprecated, ignored, "
+            "and will be removed in the future. "
+            "Please use `bpd.options.compute.allow_large_results = <value>` instead.",
```
tswast, line 266: This option shouldn't be ignored. Let's be more precise about when it is ignored.

```suggestion
            "Setting `bpd.options.bigquery.allow_large_results` is deprecated, "
            "and will be removed in the future. "
            "Please use `bpd.options.compute.allow_large_results = <value>` instead. "
            "The `bpd.options.bigquery.allow_large_results` option is ignored if "
            "`bpd.options.compute.allow_large_results` is set.",
```

bigframes/_config/__init__.py
Genesis929: Updated

bigframes/_config/bigquery_options.py
```diff
@@ -239,21 +237,43 @@ def skip_bq_connection_check(self, value: bool):
...
     def allow_large_results(self, value: bool):
+        """
+        DEPRECATED: Setting ``allow_large_results`` via ``bpd.options.bigquery``
+        will be deprecated. Use
+        ``bpd.options.compute.allow_large_results`` instead.
+
+        Warning: Setting this option here will be deprecated, and this setter
+        will be removed in a future version. The configuration for handling large
+        results has moved.
+        """
```
tswast, line 262: Generally, we don't need a docstring on the setter, just the getter.

```suggestion
```

tswast (approved):

PR #1598, opened 2025-04-05
feat: Improve local data validation

TrevorBergeron:

### DIFF
--- a/bigframes/core/local_data.py
+++ b/bigframes/core/local_data.py
@@ -18,8 +18,12 @@ from __future__ import annotations
 
 import dataclasses
 import functools
+from typing import Union
 import uuid
 
+import geopandas
+import numpy as np
+import pandas
 import pyarrow as pa
 
 import bigframes.core.schema as schemata
@@ -32,8 +36,14 @@ class LocalTableMetadata:
     row_count: int
 
     @classmethod
-    def from_arrow(cls, table: pa.Table):
+    def from_arrow(cls, table: pa.Table) -> LocalTableMetadata:
         return cls(total_bytes=table.nbytes, row_count=table.num_rows)
+
+
+_MANAGED_STORAGE_TYPES_OVERRIDES: dict[bigframes.dtypes.Dtype, pa.DataType] = {
+    # wkt to be precise
+    bigframes.dtypes.GEO_DTYPE: pa.string()
+}
 
 
 @dataclasses.dataclass(frozen=True)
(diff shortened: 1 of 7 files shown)


bigframes/core/local_data.py
```diff
@@ -32,51 +36,109 @@ class LocalTableMetadata:
     row_count: int
 
     @classmethod
-    def from_arrow(cls, table: pa.Table):
+    def from_arrow(cls, table: pa.Table) -> LocalTableMetadata:
         return cls(total_bytes=table.nbytes, row_count=table.num_rows)
 
 
+_MANAGED_STORAGE_TYPES_OVERRIDES: dict[bigframes.dtypes.Dtype, pa.DataType] = {
```
chelsea-lin, line 43: Now that we use ManagedArrowTable (leveraging PyArrow) for local data, should we centralize our conversion logic based on that? Specifically, `DataFrameAndLabels` also handles conversions (Pandas <-> BigQuery), unifying these under a single class seems beneficial. If we standardize on PyArrow as the intermediate "transfer station," the Pandas-to-BigQuery conversion could simply become a Pandas -> Arrow -> BigQuery process, simplifying the overall system.

chelsea-lin (changes requested):

tests/unit/session/test_io_pandas.py
```diff
@@ -504,17 +503,3 @@ def test_read_pandas_with_bigframes_dataframe():
         ValueError, match=re.escape("read_pandas() expects a pandas.DataFrame")
     ):
         session.read_pandas(df)
-
-
-def test_read_pandas_inline_w_noninlineable_type_raises_error():
```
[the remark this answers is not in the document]
TrevorBergeron: everything is inlinable now, no way to get this error anymore

bigframes/core/local_data.py
TrevorBergeron: Yes, my plan is that everything goes through local managed storage, to ensure consistent normalization and validation.

bigframes/core/local_data.py
```diff
@@ -32,51 +36,109 @@ class LocalTableMetadata:
...
+    if dtype in _MANAGED_STORAGE_TYPES_OVERRIDES.keys():
+        return _MANAGED_STORAGE_TYPES_OVERRIDES[dtype]
+    else:
+        return bigframes.dtypes.bigframes_dtype_to_arrow_dtype(dtype)
+
+
+def _adapt_pandas_series(
+    series: pandas.Series,
+) -> tuple[Union[pa.ChunkedArray, pa.Array], bigframes.dtypes.Dtype]:
+    if series.dtype == np.dtype("O"):
```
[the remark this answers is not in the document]
TrevorBergeron: Yeah, valid concern, especially since we do not control geopandas constructor. Probably should try pyarrow conversion first, and try geopandas as a fallback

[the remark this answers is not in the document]
TrevorBergeron: ok, changed to fallback to attempting object->geo only after other conversions fail.

bigframes/core/local_data.py
```diff
@@ -32,51 +36,113 @@ class LocalTableMetadata:
...
+def _adapt_pandas_series(
+    series: pandas.Series,
+) -> tuple[Union[pa.ChunkedArray, pa.Array], bigframes.dtypes.Dtype]:
+    # Mostly rely on pyarrow conversions, but have to convert geo without its help.
+    if series.dtype == bigframes.dtypes.GEO_DTYPE:
+        series = geopandas.GeoSeries(series).to_wkt(rounding_precision=-1)
+        return pa.array(series, type=pa.string()), bigframes.dtypes.GEO_DTYPE
+    try:
+        return _adapt_arrow_array(pa.array(series))
+    except Exception as e:
```
chelsea-lin, line 120: Instead of catching a generic Exception, what specific exception types should I expect when performing geographic data type conversions? Could TypeError be one of them?

chelsea-lin (approved):

bigframes/core/local_data.py
TrevorBergeron: There are a couple TypeError, but also ArrowInvalid. I don't really trust it not to change though? the docs don't specify what error types they will give you.

PR #1609, opened 2025-04-10
docs: add details for `bigquery_connection` in `@bpd.udf` docstring

jialuoo:


bigframes/functions/_function_session.py
```diff
@@ -775,6 +775,8 @@ def udf(
                 Name of the BigQuery connection. It is used to provide an
                 identity to the serverless instances running the user code. It
                 helps BigQuery manage and track the resources used by the udf.
+                When it defaults to None, the udf will be created without
```
GarrettWu, line 778: Can we explicit say that for internet transactions, a connection is required? And for talking with other GCP services, corresponding permissions need to be granted to the connection SA.

bigframes/functions/_function_session.py
```diff
@@ -775,6 +775,8 @@ def udf(
...
+                referencing a specific connection in its BigQuery DDL.
```
tswast, line 779: Nit: `DDL` isn't really relevant to Python users. Let's rephrase. Accounting for @GarrettWu's feedback, let's try this:

```suggestion
                any connection. A udf without a connection has no
                internet access and no access to other GCP services.
```

bigframes/functions/_function_session.py
jialuoo: Added. Thanks for the suggestion.

bigframes/functions/_function_session.py
jialuoo: Applied. Thanks.

GarrettWu (approved):

tswast (approved):

PR #1610, opened 2025-04-10
deps: remove parsy dependency

tswast:
We do not need to specify ibis types as strings.

Towards b/319116851
🦕


sycai (approved):

tswast:
Looks like I missed a few `.cast("timestamp")` calls. Will remove.

PR #1614, opened 2025-04-12
feat: enable local json string validations

chelsea-lin:
Fixes internal issue 401055880 🦕

### DIFF
--- a/bigframes/core/local_data.py
+++ b/bigframes/core/local_data.py
@@ -31,6 +31,7 @@ import pyarrow as pa
 import pyarrow.parquet  # type: ignore
 
 import bigframes.core.schema as schemata
+import bigframes.core.utils as utils
 import bigframes.dtypes
 
 
@@ -166,6 +167,9 @@ def _validate_content(array: pa.Array, dtype: bigframes.dtypes.Dtype):
     if dtype == bigframes.dtypes.JSON_DTYPE:
         values = array.to_pandas()
         for data in values:
+            # Skip scalar null values to avoid `TypeError` from json.load.
+            if not utils.is_list_like(data) and pd.isna(data):
+                continue
             try:
                 # Attempts JSON parsing.
                 json.loads(data)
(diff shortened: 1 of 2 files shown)


bigframes/core/local_data.py
```diff
@@ -150,6 +151,38 @@ def validate(self):
                     f"Field {bf_field} has arrow array type: {arrow_type}, expected type: {expected_arrow_type}"
                 )
 
+        if include_context:
```
TrevorBergeron, line 154: content, not context

bigframes/core/local_data.py
```diff
@@ -150,6 +151,38 @@ def validate(self):
...
+    Recursively validates the content of a PyArrow Array based on the
+    expected BigFrames dtype, focusing on complex types like JSON, structs,
+    and arrays where the Arrow type alone isn't sufficient.
+    """
+    # TODO: validate GEO data context.
+    if dtype == bigframes.dtypes.JSON_DTYPE:
+        values = array.to_pandas()
+        for data in values:
+            # Skip scalar null values to avoid `TypeError` from json.load.
+            if not utils.is_list_like(data) and pd.isna(data):
```
TrevorBergeron, line 171: Isn't everything a string or a null? how do you get a list-like?

bigframes/core/local_data.py
```diff
@@ -329,3 +362,15 @@ def _physical_type_replacements(dtype: pa.DataType) -> pa.DataType:
     if dtype in _ARROW_MANAGED_STORAGE_OVERRIDES:
         return _ARROW_MANAGED_STORAGE_OVERRIDES[dtype]
     return dtype
+
+
+def _is_valid_json_series(s: pd.Series):
```
TrevorBergeron, line 367: can drop this?

chelsea-lin: Dropped.

bigframes/core/local_data.py
chelsea-lin: Now, the data is formatted using pyarrow storage, so these checks are no longer necessary. They are only required to handle values like `None`, or `[]` in pandas.

bigframes/core/local_data.py
chelsea-lin: Fixed. Thanks for catching that.

TrevorBergeron (approved):

bigframes/core/local_data.py
chelsea-lin: Re-adding null value checks to ensure compatibility with Python 3.9 tests.

PR #1617, opened 2025-04-15
feat: Support write api as loading option

TrevorBergeron:


bigframes/core/local_data.py
```diff
@@ -96,27 +96,65 @@ def from_pyarrow(self, table: pa.Table) -> ManagedArrowTable:
...
+                if pa.types.is_duration(type):
+                    return pa.int64()
+                return type
+
+            schema = pa.schema(
+                pa.field(field.name, durations_to_ints(field.type))
+                for field in self.data.schema
+            )
+
+            # Can't use RecordBatch.cast until set higher min pyarrow version
```
tswast, line 126: Could we include which version of pyarrow that is? Also, this seems like a `TODO` to switch to RecordBatch.cast when we can.

```
# TODO: Use RecordBatch.cast once our minimum pyarrow version is at least X.Y.Z+.
```

bigframes/core/local_data.py
```diff
@@ -96,27 +96,65 @@ def from_pyarrow(self, table: pa.Table) -> ManagedArrowTable:
...
+            def convert_batch(batch: pa.RecordBatch) -> pa.RecordBatch:
+                return pa.record_batch(
+                    [arr.cast(type) for arr, type in zip(batch.columns, schema.types)],
+                    schema=schema,
+                )
+
+            batches = map(convert_batch, batches)
```
tswast, line 133: This type conversion logic seems like it could be pulled out into a separate function for easier unit testing.

Optional: parameterize it to support conversions from any type into any type. This might be useful for JSON in future, for example. Then again https://wiki.c2.com/?YouArentGonnaNeedIt, so let's just start with a refactor of the duration logic and we can parameterize it when we need such functionality.

bigframes/session/loader.py
```diff
@@ -277,6 +283,53 @@ def stream_data(self, data: local_data.ManagedArrowTable) -> core.ArrayValue:
...
+                    batch.serialize().to_pybytes()
+                )
+                yield request
+
+        for response in self._write_client.append_rows(requests=request_gen()):
+            if response.row_errors:
+                raise ValueError(
+                    f"Problem loading at least one row from DataFrame: {response.row_errors}. {constants.FEEDBACK_LINK}"
+                )
+        destination_table = self._bqclient.get_table(bq_table_ref)
```
tswast, line 324: Do we need to do something here to finalize the stream? https://cloud.google.com/bigquery/docs/write-api-streaming

TrevorBergeron: Not strictly necessary, but can avoid limits, per docs: "This step is optional in committed type, but helps to prevent exceeding the limit on active streams"

bigframes/core/local_data.py
TrevorBergeron: yeah, I don't want to generalize this too hard, as the general case might involve more complex things than a simple cast, and I don't want to build the scaffolding for it without reason. Next iteration will pull out duration logic though.

bigframes/core/local_data.py
TrevorBergeron: Yeah, updated to TODO with the needed version, 16.0, which is higher than our current floor, which is 15.0.2

bigframes/session/loader.py
TrevorBergeron: added finalize in new iteration

bigframes/core/utils.py
```diff
@@ -142,6 +142,9 @@ def label_to_identifier(label: typing.Hashable, strict: bool = False) -> str:
         identifier = re.sub(r"[^a-zA-Z0-9_]", "", identifier)
         if not identifier:
             identifier = "id"
+        elif identifier[0].isdigit():
+            # first character must be letter or underscore
+            identifier = "_" + identifier
```
tswast, line 147: Are you sure this is necessary? This query runs fine for me:

```
SELECT 1 AS `123xyz`;
```

So does

```
SELECT 1 AS `123`;
```

TrevorBergeron: In my testing, parquet uploads were more strict than the sql interface.

tswast: Mailed https://github.com/googleapis/python-bigquery-dataframes/pull/1645 with some tests with some of the weirder column names like starting with a number and just numbers.

TrevorBergeron: ok, merged your change in your tests. looks fine, I think the reason is that the final sql ids for to_gbq are generated by _prepare_export, which uses the non-strict version of get_standardized_ids.

bigframes/session/__init__.py
```diff
@@ -805,6 +806,10 @@ def _read_pandas(
             return self._loader.read_pandas(
                 pandas_dataframe, method="stream", api_name=api_name
             )
+        elif write_engine == "bigquery_write":
```
tswast, line 809: Could you also add this to the various docstrings? Let's mark the "bigquery_write" option as [Preview] in the docs, too.

https://github.com/googleapis/python-bigquery-dataframes/blob/26ae5e7e769c5bde2d02063f7a475ed5024b6054/bigframes/session/__init__.py#L716

tswast (commented):
Love it. Thanks! Just one last thing

tswast (approved):
Thanks!

PR #1618, opened 2025-04-15
feat: Short circuit query for local scan

TrevorBergeron:


bigframes/core/local_data.py
```diff
@@ -280,6 +296,34 @@ def _adapt_pandas_series(
...
+        )
+        return struct_array, dtype
+    if pa.types.is_list(array.type):
+        assert isinstance(array, pa.ListArray)
+        values, values_type = _adapt_arrow_array(array.values)
+        new_value = pa.ListArray.from_arrays(
+            array.offsets, values, mask=array.is_null()
+        )
+        return new_value.fill_null([]), bigframes.dtypes.list_type(values_type)
+    if array.type == bigframes.dtypes.JSON_ARROW_TYPE:
```
chelsea-lin, line 325: With `_canonicalize_json`, you probably can remove `_validate_content` method in this module

bigframes/core/local_data.py
```diff
@@ -292,6 +336,19 @@ def _adapt_arrow_array(
     return array, bf_type
 
 
+def _canonicalize_json(array: pa.Array) -> pa.Array:
+    def _canonicalize_scalar(json_string):
+        if json_string is None:
+            return None
+        return json.dumps(
```
chelsea-lin, line 343: Please add comments why we need to canonicalize json here? Refer to:
            # `sort_keys=True` sorts dictionary keys before serialization, making
            # JSON comparisons deterministic.
            # `separators=(',', ':')` eliminate whitespace to get the most compact
            # JSON representation.

bigframes/session/local_scan_execution.py
```diff
@@ -0,0 +1,68 @@
+# Copyright 2025 Google LLC
```
chelsea-lin, line 1: rename the file as `local_scan_executor` to match with `bq_caching_executor`?

bigframes/session/local_scan_execution.py
```diff
@@ -0,0 +1,68 @@
...
+# limitations under the License.
+from __future__ import annotations
+
+from typing import Optional
+
+from bigframes.core import bigframe_node, rewrite
+from bigframes.session import executor, semi_executor
+
+
+class LocalScanExecutor(semi_executor.SemiExecutor):
```
chelsea-lin, line 22: if it inherits from `executor.Executor`, what would be the problem? The polar executor and sql compiler executor (coming for sqlglot unit tests) are inheriting from `executor.Executor`. Thinking if we can avoid introduce additional classes for better readable and less maintaining.

tests/system/small/bigquery/test_json.py
```diff
@@ -155,7 +155,7 @@ def test_json_extract_array_from_json_strings():
     )
     actual = bbq.json_extract_array(s, "$.a")
     expected = bpd.Series(
-        [['"ab"', '"2"', '"3 xy"'], [], ['"4"', '"5"'], None],
+        [['"ab"', '"2"', '"3 xy"'], [], ['"4"', '"5"'], []],
```
chelsea-lin, line 158: as we discussed, can we please check if `None`->`[]` is required or not?

TrevorBergeron: Reverting, it works without this change

bigframes/session/local_scan_execution.py
TrevorBergeron: SemiExecutor is a much smaller interface with only a single method. The general executor interface supports exporting, caching, etc. Can hopefully much simplify the general executor interface at some point, but until then, semi-executor helps provide the interface for hybrid execution components

bigframes/core/local_data.py
TrevorBergeron: added comment

bigframes/core/local_data.py
TrevorBergeron: removed

bigframes/session/local_scan_execution.py
TrevorBergeron: done

chelsea-lin (approved):

PR #1619, opened 2025-04-16
fix!: make `dataset` and `name` params mandatory in `udf`

shobsi:
This change will force the user to provide a dataset and name for the BigQuery managed python udf created through BigFrames, for easy discovery and cleanup when necessary.

Fixes internal issue 409580589 🦕


tswast:
@shobsi we should not use `chore` for things that should be visible to the user in the changelog. I've renamed to `fix`.

tswast (approved):

shobsi:
> @shobsi we should not use `chore` for things that should be visible to the user in the changelog. I've renamed to `fix`.

Wouldn't the `!` make it visible under BREAKING CHANGES?

tswast:
Those two things really shouldn't be mixed. `chore: ` implies this isn't something that should affect the user, but this clearly does.

See https://github.com/googleapis/release-please/blob/422df5a6a9d44a52c6c682563d1795b8fd246215/src/strategies/python.ts#L31-L44 for options.

PR #1627, opened 2025-04-17
fix: add retry for test_clean_up_via_context_manager

sycai:


tests/system/large/test_session.py
```diff
@@ -151,6 +151,7 @@ def test_clean_up_by_session_id():
         pytest.param(bigframes.connect, id="connect-method"),
     ],
 )
+@pytest.mark.flaky(retries=3)
```
shobsi, line 154: Curious what flakiness we ran into.. quota issue?

sycai: Feels like something about server latency: b/411420369

Example error: https://fusion2.corp.google.com/invocations/8f00eb76-d4fd-4c84-899d-5fbf2645e44c/targets/bigframes%2Fcontinuous%2Fe2e/log

shobsi (approved):

PR #1634, opened 2025-04-18
fix: improve robustness of managed udf code extraction

jialuoo:
Fix the issue in b/409571106.

### DIFF
--- a/bigframes/functions/_function_client.py
+++ b/bigframes/functions/_function_client.py
@@ -19,6 +19,7 @@ import inspect
 import logging
 import os
 import random
+import re
 import shutil
 import string
 import tempfile
@@ -272,7 +273,9 @@ class FunctionClient:
             # i.e. there are no references to variables or imports outside the
             # body.
             udf_code = textwrap.dedent(inspect.getsource(func))
-            udf_code = udf_code[udf_code.index("def") :]
+            match = re.search(r"^def ", udf_code, flags=re.MULTILINE)
+            assert match is not None, "The UDF is not defined correctly."
+            udf_code = udf_code[match.start() :]
 
         with_connection_clause = (
             (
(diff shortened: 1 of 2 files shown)


shuoweil (approved):

bigframes/functions/_function_client.py
```diff
@@ -272,7 +273,9 @@ def provision_bq_managed_function(
             # i.e. there are no references to variables or imports outside the
             # body.
             udf_code = textwrap.dedent(inspect.getsource(func))
-            udf_code = udf_code[udf_code.index("def") :]
+            match = re.search(r"^def ", udf_code, flags=re.MULTILINE)
```
shobsi, line 276: let's add a test case for which kind of usage which is not covered currently but gets covered after this change.

bigframes/functions/_function_client.py
```diff
@@ -272,7 +273,9 @@ def provision_bq_managed_function(
...
+            assert match is not None, "The UDF is not defined correctly."
```
shobsi, line 277: ValueError would be more appropriate than AssertionError

tests/system/large/functions/test_managed_function.py
```diff
@@ -171,7 +171,14 @@ def featurize(x: int) -> list[float]:
 def test_managed_function_series_apply(session, dataset_id, scalars_dfs):
     try:
 
-        @session.udf(dataset=dataset_id, name=prefixer.create_prefix())
+        @session.udf(
+            dataset=dataset_id,
+            # The suffix '_def_to_test_code_extraction' is included in the UDF
```
shobsi, line 176: Suggest rewriting this to make the intent explicit and less prone to accidental edits:

```python
        # An explicit name with "def" in it is used to test the robustness of the
        # user code extraction logic, which depends on that term.
        bq_name = f"{prefixer.create_prefix()}_def_to_test_code_extraction"
        assert "def" in bq_name

        @session.udf(
            dataset=dataset_id,
            name=bq_name,
        )
```

shobsi (approved):

PR #1636, opened 2025-04-19
refactor: Unify compile paths with ResultNode

TrevorBergeron:

### DIFF
--- a/bigframes/core/compile/compiler.py
+++ b/bigframes/core/compile/compiler.py
@@ -44,7 +44,7 @@ class CompileRequest:
     node: nodes.BigFrameNode
     sort_rows: bool
     materialize_all_order_keys: bool = False
-    peek_rows: typing.Optional[int] = None
+    peek_count: typing.Optional[int] = None
 
 
 @dataclasses.dataclass(frozen=True)
@@ -59,7 +59,7 @@ def compile_sql(request: CompileRequest) -> CompileResult:
     result_node = nodes.ResultNode(
         request.node,
         output_cols=output_names,
-        limit=request.peek_rows,
+        limit=request.peek_count,
     )
     if request.sort_rows:
         # Can only pullup slice if we are doing ORDER BY in outermost SELECT
--- a/bigframes/core/compile/compiler.py
+++ b/bigframes/core/compile/compiler.py
@@ -59,7 +59,6 @@ def compile_sql(request: CompileRequest) -> CompileResult:
     result_node = nodes.ResultNode(
         request.node,
         output_cols=output_names,
-        is_ordered=request.sort_rows,
         limit=request.peek_rows,
     )
     if request.sort_rows:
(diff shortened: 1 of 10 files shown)


bigframes/core/compile/compiler.py
```diff
@@ -34,48 +37,61 @@
...
-        if (pulled_up_limit is not None) and (
-            (limit is None) or limit > pulled_up_limit
-        ):
-            limit = pulled_up_limit
+@dataclasses.dataclass(frozen=True)
+class CompileRequest:
+    node: nodes.BigFrameNode
+    sort_rows: bool
+    materialize_all_order_keys: bool = False
+    peek_rows: typing.Optional[int] = None
```
sycai, line 47: naming nit: "peek_count" to indicate that is a numeric value?

bigframes/core/compile/googlesql/query.py
```diff
@@ -63,14 +63,23 @@ class Select(abc.SQLSyntax):
...
+            typing.Iterable[str], typing.Iterable[tuple[str, str]], str, None
+        ] = None,
         distinct: bool = False,
     ) -> Select:
         if isinstance(columns, str):
             columns = [columns]
         self.select_list: typing.List[typing.Union[SelectExpression, SelectAll]] = (
             [
                 SelectExpression(expression=expr.ColumnExpression(name=column))
+                if isinstance(column, str)
```
sycai, line 76: This logic is probably complicated enough to be defined in a separate helper function.

bigframes/core/nodes.py
```diff
@@ -1602,11 +1602,51 @@ def remap_refs(
...
+    @property
+    def consumed_ids(self) -> COLUMN_SET:
+        out_refs = frozenset(ref.id for ref, _ in self.output_cols)
+        order_refs = self.order_by.referenced_columns if self.order_by else frozenset()
+        return out_refs | order_refs
+
+    @property
+    def row_count(self) -> Optional[int]:
+        child_count = self.child.row_count
+        if child_count is not None:
```
sycai, line 1641: Shall we unnest the code?

```
if child_count is None:
  return None
if self.limit is None:
  return child_count
return min(self.limit, child_count)
```

bigframes/core/nodes.py
```diff
@@ -1602,11 +1602,51 @@ def remap_refs(
 
 
 # Introduced during planing/compilation
+# TODO: Enforce more strictly that this should never be a child node
 @dataclasses.dataclass(frozen=True, eq=False)
 class ResultNode(UnaryNode):
-    output_names: tuple[str, ...]
-    order_by: Tuple[OrderingExpression, ...] = ()
+    output_cols: tuple[tuple[ex.DerefOp, str], ...]
+    is_ordered: bool
```
sycai, line 1609: is `is_ordered` necessary? Can we rely on the nullness of `order_by` to decide whether it is ordered?

If not, we might need to document how we are going to handle this confusing case when `is_ordered == False` and `order_by == None`

bigframes/core/compile/compiler.py
```diff
@@ -34,48 +37,61 @@
...
+        # Can only pullup slice if we are doing ORDER BY in outermost SELECT
+        # Need to do this before replacing unsupported ops, as that will rewrite slice ops
+        result_node = rewrites.pull_up_limits(result_node)
+    result_node = _replace_unsupported_ops(result_node)
+    # prune before pulling up order to avoid unnnecessary row_number() ops
+    result_node = cast(nodes.ResultNode, rewrites.column_pruning(result_node))
+    result_node = rewrites.defer_order(
+        result_node, output_hidden_row_keys=request.materialize_all_order_keys
+    )
+    if not request.sort_rows:
```
sycai, line 75: style nit: let's re-write this if-else as:

if request.sort_rows:
   ...
else:
   ...

(the "else" indentation might be unnecessary too, because the "if" branch always returns)

bigframes/core/compile/compiler.py
TrevorBergeron: done

bigframes/core/compile/googlesql/query.py
TrevorBergeron: done

bigframes/core/nodes.py
TrevorBergeron: done

bigframes/core/nodes.py
TrevorBergeron: yeah, not needed, removed

bigframes/core/compile/compiler.py
TrevorBergeron: done

sycai (approved):

PR #1638, opened 2025-04-21
fix: Remove itertools.pairwise usage

TrevorBergeron:


bigframes/core/local_data.py
```diff
@@ -229,7 +229,7 @@ def _(
         value_generator = iter_array(
             array.flatten(), bigframes.dtypes.get_array_inner_type(dtype)
         )
-        for (start, end) in itertools.pairwise(array.offsets):
+        for (start, end) in _pairwise(array.offsets):
```
shobsi, line 232: Any test we can add to showcase what this change fixes?

TrevorBergeron: added a unit test for this path

shobsi (approved):

bigframes/core/local_data.py
shobsi: Thank you!

PR #1654, opened 2025-04-23
perf: Use flyweight for node fields

TrevorBergeron:


bigframes/core/chain_list.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+T = TypeVar("T")
+
+# Further optimizations possible:
+# * Support mapping operators
+# * Support insertions and deletions
+
+
+class ChainList(collections.abc.Sequence[T]):
```
sycai, line 31: We should add some more doc to describe what this classes does. It's not obvious that the goal is to quickly locate an element from a concatenated sequence.

Plus, perhaps "ChainedSequence" is a better name?

In addition, this implementation might hurt performance if the provided "parts" are all very short. In that case, we  are effectively using iterations in __getitem__(). Does it make sense to have certain kind of threshold? That is, if the total number of element is below certain threshold (10 maybe?), just concat everything into a tuple. Otherwise, use the current technique.

bigframes/core/chain_list.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2024 Google LLC
```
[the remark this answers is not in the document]
TrevorBergeron: done

bigframes/core/chain_list.py
TrevorBergeron: added docstring and renamed

Added some conservative compaction logic as well. Key here is memory efficiency/reuse rather than lookup efficiency, so don't want to go to wild on those types of optimizations.

sycai (approved):

PR #1656, opened 2025-04-24
perf: Fold row count ops when known

TrevorBergeron:


bigframes/core/rewrite/fold_row_count.py
```diff
@@ -0,0 +1,44 @@
...
+from __future__ import annotations
+
+import pyarrow as pa
+
+from bigframes import dtypes
+from bigframes.core import local_data, nodes
+from bigframes.operations import aggregations
+
+
+def fold_row_counts(node: nodes.BigFrameNode) -> nodes.BigFrameNode:
```
sycai, line 23: A naming suggestion for this module: I think we can name the file as "row_counts" 

The good thing is that if you want to add more re-write functions for row_counts in the future, they will all look like this `row_counts.expand(..)`, `row_counts.reverse(...)`, which keeps the code organized.

bigframes/core/blocks.py
```diff
@@ -221,7 +221,14 @@ def shape(self) -> typing.Tuple[int, int]:
...
 
-        row_count = self.session._executor.get_row_count(self.expr)
+        row_count = next(
+            iter(
+                self.session._executor.execute(self.expr.row_count())
+                .to_arrow_table()
+                .to_pydict()
+                .values()
+            )
+        )[0]
```
sycai, line 231: Do we want to check and fail fast if the number of values is not equal to 1?

bigframes/core/rewrite/fold_row_count.py
TrevorBergeron: `fold_row_count` isn't the module though, `core.compile.rewrite` is. For now this file just defines a single symbol, so figure best to be as precise as possible? Easy enough to change later if add more to this file.

bigframes/core/blocks.py
TrevorBergeron: Actually pulled this in to the execute result object itself, and add asserts

sycai (approved):

PR #1660, opened 2025-04-25
chore: replace api_name with updated method_logger

Genesis929:


tswast:
@Genesis929 It looks like there are some conflicts to resolve before this can be fully reviewed.

bigframes/core/blocks.py
```diff
@@ -2646,7 +2646,7 @@ def _get_rows_as_json_values(self) -> Block:
         # The only ways this code is used is through df.apply(axis=1) cope path
         # TODO: Stop using internal API
```
tswast, line 2647: I don't understand what this TODO is about. I think it's probably out of date. Let's remove it.

```suggestion
```

tests/unit/core/test_log_adapter.py
```diff
@@ -63,44 +82,24 @@ def test_method_logging(test_instance):
...
-def test_property_logging(test_instance):
-    test_instance.my_field
-
-    # Check if the properties were added to the _api_methods list
+def test_method_logging(test_method):
+    test_method()
     api_methods = log_adapter.get_and_reset_api_methods()
-    assert "testclass-my_field" in api_methods
-
+    assert "<locals>-method1" in api_methods
```
tswast, line 88: Are these characters allowed?

> Keys and values can contain only lowercase letters, numeric characters, underscores, and dashes.

https://cloud.google.com/bigquery/docs/labels-intro#requirements

I think we need some logic to remove invalid characters.

Genesis929: These characters should only exist in tests, unless someone uses them in a similar way. Updated in case someone actually uses them this way.

bigframes/core/log_adapter.py
```diff
@@ -213,12 +224,24 @@ def wrapped(*args, **kwargs):
     )
 
 
+def log_name_override(name: str):
```
Genesis929, line 227: added a decorator for cases where the logged method name is not the same as the real method name. This is for _read_gbq_colab.

tswast: Thank you!

tswast:
@Genesis929 It looks like there are some conflicts. Please resolve and ping me when ready to review.

tswast (approved):

PR #1661, opened 2025-04-27
perf: defer query in `read_gbq` with wildcard tables

tswast:

Fixes internal issue 405773140 🦕


tswast:
Failures look like real failures.

```
        if not 200 <= response.status_code < 300:
>           raise exceptions.from_http_response(response)
E           google.api_core.exceptions.BadRequest: 400 GET https://bigquery.googleapis.com/bigquery/v2/projects/python-docs-samples-tests/queries/32f42306-e95f-48bc-a2fb-56761aec5476?maxResults=0&location=US&prettyPrint=false: Invalid field name "_TABLE_SUFFIX". Field names are not allowed to start with the (case-insensitive) prefixes _PARTITION, _TABLE_, _FILE_, _ROW_TIMESTAMP, __ROOT__ and _COLIDENTIFIER
E           
E           Location: US
E           Job ID: 32f42306-e95f-48bc-a2fb-56761aec5476
```

While we can query such fields, it looks like we can't materialize them.

Genesis929 (dismissed):

tswast:
Added `do not merge`. Need to make sure this is compatible with `to_gbq()` and `cached()`.

tswast:
From the notebook tests:

```
File /[tmpfs/src/github/python-bigquery-dataframes/bigframes/core/nodes.py:711](https://cs.corp.google.com/piper///depot/google3/tmpfs/src/github/python-bigquery-dataframes/bigframes/core/nodes.py?l=711), in GbqTable.from_table(table, columns)
    708 @staticmethod
    709 def from_table(table: bq.Table, columns: Sequence[str] = ()) -> GbqTable:
    710     # Subsetting fields with columns can reduce cost of row-hash default ordering
--> 711     table_schema = bigframes.core.tools.bigquery.get_schema_and_pseudocolumns(table)
    713     if columns:
    714         schema = tuple(item for item in table_schema if item.name in columns)

AttributeError: module 'bigframes.core.tools' has no attribute 'bigquery'
```

Looks like we're missing some imports too.

tswast:
Tested the failing samples tests locally. I think my latest commits solve the issue of not being able to materialize to `_TABLE_SUFFIX` as the column name.

bigframes/session/bq_caching_executor.py
```diff
@@ -373,6 +374,15 @@ def _cache_with_cluster_cols(
     ):
         """Executes the query and uses the resulting table to rewrite future executions."""
 
+        # Pseudocolumns can be queried and even written to anonymous query
+        # results tables, but they can't be materialized to an explicit
+        # destination table. Therefore, we rename the columns to all be writable
+        # before executing the SQL. See: b/405773140 for discussion about the
+        # _TABLE_SUFFIX pseudocolumn.
+        prev_col_ids = array_value.column_ids
+        new_col_ids, _ = utils.get_standardized_ids(prev_col_ids)
```
tswast, line 383: In the case that standardization makes duplicate column names, we need another pass to remove duplicates.

tswast: Nevermind, `get_standardized_ids` already does this. It's `label_to_identifier` that doesn't.

tswast:
e2e and notebook failures are for the same:

```
E               google.api_core.exceptions.BadRequest: 400 'FOR SYSTEM_TIME AS OF' expression for table 'bigframes-load-testing.bigframes_testing.penguins_dcdc3525965d3bf2805a055ee80a0ae7' evaluates to a TIMESTAMP value in the future: 2025-04-29 15:16:24.477885 UTC.; reason: invalidQuery, location: query, message: 'FOR SYSTEM_TIME AS OF' expression for table 'bigframes-load-testing.bigframes_testing.penguins_dcdc3525965d3bf2805a055ee80a0ae7' evaluates to a TIMESTAMP value in the future: 2025-04-29 15:16:24.477885 UTC.
```

I don't think these relate to this change, but I do recall BQML having a hard time with time travel. Potentially we're missing a force cache somewhere now?

bigframes/session/bq_caching_executor.py
```diff
@@ -355,35 +356,63 @@ def _cache_with_cluster_cols(
...
+        renamed = array_value.rename_columns(dict(zip(prev_col_ids, new_col_ids)))
+
         sql, schema, ordering_info = self.compiler.compile_raw(
-            self.logical_plan(array_value.node)
+            self.logical_plan(renamed.node)
         )
```
TrevorBergeron, line 370: maybe there is a better way, where the compiler just does not return invalid sql ever?

bigframes/session/bq_caching_executor.py
```diff
@@ -355,35 +356,63 @@ def _cache_with_cluster_cols(
...
-            ordering=ordering_info,
-        ).node
+        cached_replacement = (
+            renamed.as_cached(
+                cache_table=self.bqclient.get_table(tmp_table),
+                ordering=ordering_info,
+            )
+            .rename_columns(dict(zip(new_col_ids, prev_col_ids)))
+            .node
+        )
```
TrevorBergeron, line 383: I think we can apply the renaming as part of the scan definition of the cache node rather than as a selection node (via rename_columns) on top of it. Should be more robust that way

bigframes/session/loader.py
```diff
@@ -727,6 +729,7 @@ def _query_to_destination(
         api_name: str,
         configuration: dict = {"query": {"useQueryCache": True}},
         do_clustering=True,
+        max_results: Optional[int] = None,
```
tswast, line 732: Ended up using a different workaround (renamed _TABLE_SUFFIX to _BF_TABLE_SUFFIX), so we can revert this change.

bigframes/session/bq_caching_executor.py
tswast: I don't think that's possible. This is valid SQL before the renaming. It's just restricted as to where we can materialize.

bigframes/session/bq_caching_executor.py
tswast: I tried that in https://github.com/googleapis/python-bigquery-dataframes/pull/1661/commits/5b0d0a0a78eb19493cce8ac1185cc062259fb2f3

The problem I was having is that the physical schema often contains more columns than the array value. If we rename there, then we lose track of which columns belong to which id.

tswast: Added a "renames" argument to `as_cached`, which I does allow me to use the scan list to rename columns instead. Indeed that does seem more robust. I think some of these renames might have been pruned out in recent commits.

tswast:
e2e failure might indicate a real issue `FAILED tests/system/large/operations/test_semantics.py::test_sim_join[has_score_column]`

```
>           raise exceptions.from_http_response(response)
E           google.api_core.exceptions.BadRequest: 400 GET [https://bigquery.googleapis.com/bigquery/v2/projects/bigframes-load-testing/queries/e2e2cede-621f-47fc-9fbc-4077c0579eaf?maxResults=0&location=US&prettyPrint=false](https://www.google.com/url?q=https://bigquery.googleapis.com/bigquery/v2/projects/bigframes-load-testing/queries/e2e2cede-621f-47fc-9fbc-4077c0579eaf?maxResults%3D0%26location%3DUS%26prettyPrint%3Dfalse&sa=D): Duplicate column names in the result are not supported when a destination table is present. Found duplicate(s): creatures
```

tswast:
Different set of e2e failures this time:

```
FAILED tests/system/large/blob/test_function.py::test_blob_pdf_chunk[True-expected0]
FAILED tests/system/large/blob/test_function.py::test_blob_image_normalize_to_series
FAILED tests/system/large/blob/test_function.py::test_blob_image_blur_to_series
FAILED tests/system/large/blob/test_function.py::test_blob_image_normalize_to_folder
FAILED tests/system/large/blob/test_function.py::test_blob_image_blur_to_folder
FAILED tests/system/large/blob/test_function.py::test_blob_image_normalize_to_bq
FAILED tests/system/large/blob/test_function.py::test_blob_image_blur_to_bq
FAILED tests/system/large/blob/test_function.py::test_blob_image_resize_to_series
ERROR tests/system/large/blob/test_function.py::test_blob_pdf_extract[True-expected0]
```

Maybe it was just an LLM-related flake last time?

tswast:
I'll split to defer part from the pseudocolumn work into https://github.com/googleapis/python-bigquery-dataframes/pull/1689

I think the defer step can be done without out that to make this PR a lot smaller.

TrevorBergeron (approved):

PR #1672, opened 2025-04-29
refactor: implements `compile_selection` method

chelsea-lin:


bigframes/core/compile/sqlglot/scalar_compiler.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2023 Google LLC
```
sycai, line 1: nit: 2025?

bigframes/core/compile/sqlglot/scalar_compiler.py
```diff
@@ -0,0 +1,33 @@
...
+import sqlglot.expressions as sge
+
+from bigframes.core import expression
+
+
+@functools.singledispatch
+def compile_scalar_expression(
+    expression: expression.Expression,
+) -> sge.Expression:
+    """Defines transformation but isn't cached, always use compile_node instead"""
```
sycai, line 27: non-rhetorical question: I don't quite understand why we should "always use compile_node" here. Does it mean that this method should not be invoked directly, and under what circumstances do we expect to use this method?

Could you put more details in the doc?

chelsea-lin: The docstring has been corrected; it was previously copied by mistake. Please let me know if further details are needed.

bigframes/core/compile/sqlglot/scalar_compiler.py
chelsea-lin: Fixed. Thanks!

sycai (approved):

PR #1674, opened 2025-04-30
feat: add dry_run parameter to `read_gbq()`, `read_gbq_table()` and `read_gbq_query()`

sycai:
~If a table reference is fed to `read_gbq()` with dry_run set to True, we will use `SELECT * FROM {table_ref}` for dry run~

For `read_gbq()`, and `read_gbq_table()` calls that do not ultimately lead to SQL conversions, we use the table metadata for dry run stats report.


sycai:
> > If a table reference is fed to read_gbq() with dry_run set to True, we will use SELECT * FROM {table_ref} for dry run
> 
> 👎 That's a bit misleading. There are some code paths that do fallback to query (e.g. if max_results) is set. Those should have a dry run because they do immediately run a query. But for a deferred operation, I don't think dry run makes sense. Instead, let's populate what we can from the table metadata and have some indicator that no query is actually run.

Sounds good. Code updated. Now read_gbq_table dry run looks like this: https://screenshot.googleplex.com/AHaxiSsafniVFRN

bigframes/session/dry_runs.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # Indicate that no query is executed.
+    index.append("isQuery")
+    values.append(False)
+
+    # Populate column and index types
+    col_dtypes = dtypes.bf_type_from_type_kind(table.schema)
+    index.append("tableColumnCount")
+    values.append(len(col_dtypes))
+    index.append("tableColumnTypes")
```
tswast, line 37: It's not super easy for end user too predict if something will result in a query or just read the table directly. Could we try to align these names so that they don't need as much logic to handle one case over the other?

sycai: SG. Done

tswast (approved):
Thanks!

PR #1676, opened 2025-04-30
refactor: Factor out exectuor.head method into cache and execute slice

TrevorBergeron:

### DIFF
--- a/bigframes/core/blocks.py
+++ b/bigframes/core/blocks.py
@@ -69,6 +69,7 @@ import bigframes.dtypes
 import bigframes.exceptions as bfe
 import bigframes.operations as ops
 import bigframes.operations.aggregations as agg_ops
+from bigframes.session import executor
 
 # Type constraint for wherever column labels are used
 Label = typing.Hashable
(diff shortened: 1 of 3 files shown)


bigframes/core/blocks.py
```diff
@@ -69,6 +69,7 @@
 import bigframes.exceptions as bfe
 import bigframes.operations as ops
 import bigframes.operations.aggregations as agg_ops
+from bigframes.session import executor
```
sycai, line 72: Shall we use  `from bigfreames.session import executor as executors` so that its name does not conflict with your local variable?

bigframes/session/bq_caching_executor.py
```diff
@@ -243,46 +243,37 @@ def peek(
...
-        cluster_cols: Sequence[str] = (),
+        self, array_value: bigframes.core.ArrayValue, *, config: executor.CacheConfig
     ) -> None:
         """Write the block to a session table."""
-        # use a heuristic for whether something needs to be cached
-        if (not force) and self._is_trivially_executable(array_value):
-            return
-        if use_session:
+        # First, see if we can reuse the existing cache
+        # TODO: Provide feedback to user on whether new caching action was deemed necessary
```
sycai, line 251: Maybe we can open bugs to track these TODOs? go/todo-style

bigframes/core/blocks.py
TrevorBergeron: done

sycai (approved):

PR #1677, opened 2025-04-30
refactor: support to compile project and add_op

chelsea-lin:


bigframes/core/compile/sqlglot/scalar_compiler.py
```diff
@@ -29,5 +31,42 @@ def compile_scalar_expression(
...
+    if isinstance(op, ops.UnaryOp):
+        return method(op, args[0])
+    elif isinstance(op, ops.BinaryOp):
+        return method(op, args[0], args[1])
+    elif isinstance(op, ops.TernaryOp):
+        return method(op, args[0], args[1], args[2])
+    elif isinstance(op, ops.NaryOp):
+        return method(op, *args)
+    else:
+        raise NotImplementedError(f"Cannot compile operator {method_name}")
```
sycai, line 66: non-rhetorical question: do we expect more operator types beyond the ones that are listed above? If not, maybe 

`ValueError(f"Unrecognized operator {...}") `

is a better choice?

sycai (approved):

bigframes/core/compile/sqlglot/scalar_compiler.py
chelsea-lin: Good points. Update it into `TypeError` and `ValueError` above.

PR #1679, opened 2025-05-01
refactor: add uid generator and encasualate query as cte in SQLGlotCompiler

chelsea-lin:


bigframes/core/guid.py
```diff
@@ -19,3 +19,21 @@ def generate_guid(prefix="col_"):
...
+        self.prefix_counters = {}
+
+    def generate_sequential_uid(self, prefix: str) -> str:
+        """Generates a sequential UID with specified prefix."""
+        if prefix not in self.prefix_counters:
+            self.prefix_counters[prefix] = 0
+
+        uid = f"{prefix}{self.prefix_counters[prefix]}"
+        self.prefix_counters[prefix] += 1
+        return uid
```
TrevorBergeron, line 39: Probably nicer to define as a generator (eg. `yield uid` in loop). And maybe we want thread safety? (i know, we don't have on existing guid generators)

bigframes/core/rewrite/identifiers.py
```diff
@@ -13,30 +13,25 @@
...
-    root: bigframes.core.nodes.BigFrameNode,
-    id_generator: Generator[bigframes.core.identifiers.ColumnId, None, None],
-) -> Tuple[
-    bigframes.core.nodes.BigFrameNode,
-    dict[bigframes.core.identifiers.ColumnId, bigframes.core.identifiers.ColumnId],
-]:
-    """
-    Remap all variables in the BFET using the id_generator.
+    root: nodes.BigFrameNode,
+    uid_gen: guid.SequentialUIDGenerator,
```
TrevorBergeron, line 24: Seems unneeded to restrict the type like this.

tests/unit/core/compile/sqlglot/compiler_session.py
```diff
@@ -41,7 +42,9 @@ def to_sql(
 
         # Compared with BigQueryCachingExecutor, SQLCompilerExecutor skips
         # caching the subtree.
-        return self.compiler.compile(array_value.node, ordered=ordered)
+        return self.compiler.SQLGlotCompiler(
+            uid_gen=bigframes.core.guid.SequentialUIDGenerator()
+        ).compile(array_value.node, ordered=ordered)
```
TrevorBergeron, line 47: Why do we want the uid_gen to be part of the object if its not reusable?

bigframes/core/compile/sqlglot/sqlglot_ir.py
```diff
@@ -95,21 +102,44 @@ def from_pyarrow(
...
+        return cls(expr=sg.select(sge.Star()).from_(expr), uid_gen=uid_gen)
 
     def select(
         self,
-        select_cols: typing.Dict[str, sge.Expression],
+        selected_cols: tuple[tuple[str, sge.Expression], ...],
     ) -> SQLGlotIR:
-        selected_cols = [
+        cols_expr = [
             sge.Alias(
```
TrevorBergeron, line 112: Do we always want to alias, even if the ids don't change?

bigframes/core/guid.py
chelsea-lin: I've added a note indicating that the class is not thread-safe, as it's currently intended for internal use only. While I explored defining it as a generator, it seemed incompatible with the dictionary prefix, so I've kept the existing implementation for now.

bigframes/core/rewrite/identifiers.py
chelsea-lin: Agree. Leaving the type here for more readable but can be changed if we want to support more type here.

TrevorBergeron (approved):

PR #1689, opened 2025-05-05
feat: expose `_TABLE_SUFFIX` pseudocolumn for wildcard tables

tswast:
Follow-up to https://github.com/googleapis/python-bigquery-dataframes/pull/1661 and https://github.com/googleapis/python-bigquery-dataframes/pull/1691

Towards internal issues b/405773140 and b/416285785 🦕


tswast:
Marking as `do not merge`, because maybe there's a way we can rename invalid names earlier in the process to internal IDs to avoid having to have the "renames" in cached logic? Could also help with `read_gbq_query`, as well?

PR #1692, opened 2025-05-05
feat: Add "dayofyear" property for `dt` accessors

sycai:
Now you can get the day of year for dates, datetimes and timestamps.

### DIFF
--- a/bigframes/operations/datetimes.py
+++ b/bigframes/operations/datetimes.py
@@ -44,6 +44,10 @@ class DatetimeMethods(
         return self._apply_unary_op(ops.dayofweek_op)
 
     @property
+    def dayofyear(self) -> series.Series:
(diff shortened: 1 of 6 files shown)


TrevorBergeron (approved):

bigframes/operations/datetimes.py
```diff
@@ -43,6 +43,10 @@ def day(self) -> series.Series:
     def dayofweek(self) -> series.Series:
         return self._apply_unary_op(ops.dayofweek_op)
 
+    @property
+    def dayofyear(self) -> series.Series:
```
TrevorBergeron, line 47: Do we need a docstring?

sycai: Good call. Doc updated.

PR #1697, opened 2025-05-06
feat: add bigframes.bigquery.json_value

chelsea-lin:
This change is adding `bigframes.bigquery.json_value` function, which is tested through doctest.


GarrettWu (approved):
Can we add sys tests in addition to doc tests?

PR #1688, opened 2025-05-05
chore: create private `_read_gbq_colab` which will enable partial ordering mode even when session is in strict mode

tswast:
Also,

* chore: create private _read_gbq_colab which will enable partial ordering mode, disable progress bars, disable default index, and communicate via callbacks

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery-dataframes/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Towards internal issue  b/409104302 🦕

PR #1704, opened 2025-05-07
chore: add pyformat_args to _read_gbq_colab

tswast:
Follow-up to https://github.com/googleapis/python-bigquery-dataframes/pull/1688


tswast:
Chatted offline. Will update to avoid injecting unescaped strings. I'm considering adding google-cloud-bigquery Table* objects as a simpler alternative that can be used to represent a table before following up with DataFrame support.

TrevorBergeron (approved):

PR #1706, opened 2025-05-07
chore: improve get_started notebook

sycai:
Key changes:
* Reworded some section to differentiate it from the `template` notebook.
* Replaced `head()` calls with `peek()` when necessary
* Removed the output cell of installing bigframes
* Replaced a string concatenation with f-string
* Tweaked some constants to make the results more expressive


chelsea-lin (approved):
LTGM overall

PR #1708, opened 2025-05-08
refactor: Streamline executor paths

TrevorBergeron:


tswast (approved):

tests/system/large/test_dataframe_io.py
```diff
@@ -32,23 +29,6 @@ def test_to_pandas_batches_raise_when_large_result_not_allowed(session):
         next(df.to_pandas_batches(page_size=500, max_results=1500))
 
 
-def test_to_pandas_batches_override_global_option(
```
tswast, line 35: I'm curious why we're deleting this test. Do we have a test that makes sure that we do still create a destination table when `allow_large_results=True`?

PR #1709, opened 2025-05-08
chore: fix AttributeError when rowiterator has no attribute in g3

chelsea-lin:
When updating BigQuery DataFrames in g3, an `AttributeError` occurs when accessing the `total_bytes_processed` and `query` attributes of RowIterator. This is because g3 does not have an up-to-date Python client library. Here is the error message: https://screenshot.googleplex.com/4GhbZBdFZrVrCwA


bigframes/session/metrics.py
```diff
@@ -40,8 +40,11 @@ def count_job_stats(
     ):
         if query_job is None:
             assert row_iterator is not None
-            if (row_iterator.total_bytes_processed is None) or (
-                row_iterator.query is None
+            if (
+                not hasattr(row_iterator, "total_bytes_processed")
```
tswast, line 44: When was this attribute added? It seems to me this isn't just a "chore: " but a true bug fix for being incompatible with older google-cloud-bigquery library.

Please add a unit test and make sure it fails on at least one nox session before this fix is added.

tswast: Actually, the version is very old internally. I still think it'd be good to make sure we have tested against this version, but that might be too much for this PR.

bigframes/session/metrics.py
```diff
@@ -40,8 +40,11 @@ def count_job_stats(
...
+                or not hasattr(row_iterator, "query")
+                or (row_iterator.total_bytes_processed is None)
+                or (row_iterator.query is None)
```
tswast, line 47: The object shouldn't be mutated while we're iterating over it but we can be a bit more robust by using `getattr` instead.

```suggestion
                (total_bytes_processed := getattr(row_iterator, "total_bytes_processed", None)) is None
                (query := getattr(row_iterator, "query", None)) is None
```

tswast, line 47: The object shouldn't be mutated while we're iterating over it but we can be a bit more robust by using `getattr` instead.

```suggestion
                (total_bytes_processed := getattr(row_iterator, "total_bytes_processed", None)) is None
                (query := getattr(row_iterator, "query", None)) is None
```

bigframes/session/metrics.py
chelsea-lin: As we discussed offline, this change is not necessary for github because the `setup.py` asks `google-cloud-bigquery>=3.31.0`, but only needed for g3, which has lower version. The alternative solution would upgrade g3 version and this PR can be a short-term solution to unblock BF g3 import.

tswast (approved):

PR #1710, opened 2025-05-08
feat: Add inplace arg support to sort methods

TrevorBergeron:

### DIFF
--- a/bigframes/dataframe.py
+++ b/bigframes/dataframe.py
@@ -2234,7 +2234,7 @@ class DataFrame(vendored_pandas_frame.DataFrame):
         col_ids_strs: List[str] = [col_id for col_id in col_ids if col_id is not None]
         return DataFrame(self._block.set_index(col_ids_strs, append=append, drop=drop))
 
-    @overload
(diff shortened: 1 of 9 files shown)


bigframes/dataframe.py
```diff
@@ -2234,10 +2234,34 @@ def set_index(
...
-    @validations.requires_index
+    @overload
     def sort_index(
-        self, ascending: bool = True, na_position: Literal["first", "last"] = "last"
+        self,
+        *,
+        ascending: bool = ...,
+        inplace: Literal[False] = ...,
+        na_position: Literal["first", "last"] = ...,
     ) -> DataFrame:
```
chelsea-lin, line 2244: Why can we not just return `Optional[DataFrame]` here?

chelsea-lin (approved):
Thanks!

bigframes/dataframe.py
TrevorBergeron: this is the typing overload, see https://docs.python.org/3/library/typing.html#overload .

chelsea-lin (approved):

PR #1715, opened 2025-05-08
chore: change error to warning for gemini models

shuoweil:
chore: change error to warning for gemini models


bigframes/ml/llm.py
```diff
@@ -437,6 +444,7 @@ def __init__(
                 "gemini-2.0-flash-exp",
                 "gemini-2.0-flash-001",
                 "gemini-2.0-flash-lite-001",
+                "gemini-2.5-pro-preview-05-06",
```
GarrettWu, line 447: It appears in the IDE hints for users. That we don't want to expose yet. Just remove it.

shuoweil: done

GarrettWu (approved):

PR #1720, opened 2025-05-09
feat: Add deferred data uploading

TrevorBergeron:


bigframes/session/bq_caching_executor.py
```diff
@@ -103,6 +111,17 @@ def cache_results_table(
...
         self._cached_executions[original_root] = cached_replacement
 
+    def cache_remote_replacement(
+        self,
+        local_data: local_data.ManagedArrowTable,
+        bq_data: nodes.BigqueryDataSource,
+    ):
+        mapping = {
+            local_data.schema.items[i].column: bq_data.table.physical_schema[i].name
+            for i in range(len(local_data.schema))
```
tswast, line 121: Should we add any safeguards here in case the lengths don't match?

bigframes/session/bq_caching_executor.py
```diff
@@ -502,6 +526,48 @@ def _validate_result_schema(
...
             )
 
+    def _substitute_large_local_sources(self, original_root: nodes.BigFrameNode):
+        """
+        Replace large local sources with the uploaded version of those datasources.
+        """
+        # Step 1: Upload all previously un-uploaded data
+        for leaf in original_root.unique_nodes():
+            if isinstance(leaf, nodes.ReadLocalNode):
+                if leaf.local_data_source.metadata.total_bytes > 5000:
```
tswast, line 536: Let's create a constant for this 5000.

bigframes/session/bq_caching_executor.py
```diff
@@ -502,6 +526,48 @@ def _validate_result_schema(
...
+            bq_source, source_mapping = self.cache._uploaded_local_data[
+                node.local_data_source
+            ]
+            scan_list = node.scan_list.remap_source_ids(source_mapping)
+            if node.offsets_col is not None:
+                scan_list = scan_list.append(
+                    bq_source.table.physical_schema[-1].name,
+                    bigframes.dtypes.INT_DTYPE,
+                    node.offsets_col,
+                )
```
tswast, line 554: What's the reason that offsets_col isn't in `source_mapping`, already? Could you add a comment explaining it?

TrevorBergeron: Its a bit clumsy due to offsets_col only really being a thing for the local node, so when we upload, we will set the final physical column to be those offsets, but then we just have to add to the end of the scan_list for the ReadTableNode. Added a comment to explain. Probably some room to respec the leaf nodes for a bit more of a 1:1 mapping.

bigframes/session/bq_caching_executor.py
TrevorBergeron: Done

bigframes/session/bq_caching_executor.py
TrevorBergeron: Added an assert for exact schema length

bigframes/session/bq_caching_executor.py
```diff
@@ -28,7 +29,7 @@
 import google.cloud.bigquery_storage_v1
 
 import bigframes.core
```
tswast, line 31:
```suggestion
import bigframes.constants
import bigframes.core
```

tswast (approved):

PR #1723, opened 2025-05-12
feat: add deprecation warning to Gemini-1.5-X, text-embedding-004, and remove remove legacy models in notebooks and docs

shuoweil:
feat: add deprecation warning to Gemini-1.5-X, text-embedding-004, and remove Remove legacy models in notebooks and docs
b/416566349

### DIFF
--- a/bigframes/ml/llm.py
+++ b/bigframes/ml/llm.py
@@ -21,6 +21,7 @@ import warnings
 
 import bigframes_vendored.constants as constants
 from google.cloud import bigquery
+import typing_extensions
 
 from bigframes import dtypes, exceptions
 import bigframes.bigquery as bbq
@@ -113,9 +114,16 @@ _GEMINI_MULTIMODAL_MODEL_NOT_SUPPORTED_WARNING = (
 )
 
 
+@typing_extensions.deprecated(
+    "text-embedding-004 is going to be deprecated. Use text-embedding-005 (https://cloud.google.com/python/docs/reference/bigframes/latest/bigframes.ml.llm.TextEmbeddingGenerator) instead. ",
+    category=exceptions.ApiDeprecationWarning,
+)
 @log_adapter.class_logger
 class TextEmbeddingGenerator(base.RetriableRemotePredictor):
     """Text embedding generator LLM model.
+
+    .. note::
+        text-embedding-004 is going to be deprecated. Use text-embedding-005(    https://cloud.google.com/python/docs/reference/bigframes/latest/bigframes.ml.llm.TextEmbeddingGenerator) instead.
 
     Args:
         model_name (str, Default to "text-embedding-004"):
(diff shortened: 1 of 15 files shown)


bigframes/ml/llm.py
```diff
@@ -416,6 +432,7 @@ class GeminiTextGenerator(base.RetriableRemotePredictor):
             default and a warning will be issued.
 
         .. note::
+            "gemini-1.5-X" is going to be deprecated. Bigframes 2 will transition to using gemini-2.0-X.
```
GarrettWu, line 435: "Bigframes 2 will transition to using gemini-2.0-X. " doesn't read. Perhaps just say use gemini-2.0-X instead and give a example like gemini-2.0-flash-001?

GarrettWu:
Semantic tests are failing

shuoweil:
> Semantic tests are failing

I debug one of the failed tests (screen/6TVFXZfFb57tcn8). We are expecting the results "capybara is heavier than spider". But the actual result returns "capybara is heavier than capybara". I changed the gemini model from gemini-1.5 to gemini-2.0-flash-001. It looks like the result with "gemini-2.0-flash-001" is wrong. I am stuck on this part, since I do not have control on the result generated by gemini model.

sycai:
Updated the `ai.self_join()` test case with a more dramatic example. Let's see how it goes.

shuoweil:
> Updated the `ai.self_join()` test case with a more dramatic example. Let's see how it goes.

It looks like gemini-2.0-flash-001 does not like capybara. Test case change helps. Thanks!

shuoweil:
> Semantic tests are failing

all failed tests are fix

bigframes/ml/llm.py
```diff
@@ -113,17 +114,28 @@
 )
 
 
+@typing_extensions.deprecated(
```
GarrettWu, line 117: It deprecates the entire class. Which isn't the intention. Only warning creation is enough.

bigframes/ml/llm.py
```diff
@@ -398,6 +410,10 @@ def to_gbq(
         return new_model.session.read_gbq_model(model_name)
 
 
+@typing_extensions.deprecated(
```
GarrettWu, line 413: same

GarrettWu (dismissed):

bigframes/ml/llm.py
shuoweil: fixed

bigframes/ml/llm.py
shuoweil: fixed

GarrettWu (approved):

bigframes/ml/llm.py
```diff
@@ -112,18 +112,29 @@
...
+
     Args:
         model_name (str, Default to "text-embedding-004"):
             The model for text embedding. Possible values are "text-embedding-005", "text-embedding-004"
             or "text-multilingual-embedding-002". text-embedding models returns model embeddings for text inputs.
             text-multilingual-embedding models returns model embeddings for text inputs which support over 100 languages.
             If no setting is provided, "text-embedding-004" will be used by
             default and a warning will be issued.
+
+            .. note::
```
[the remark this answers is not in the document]
shuoweil: done

PR #1733, opened 2025-05-14
chore: add callback argument to _read_gbq_colab

tswast:

### DIFF
--- a/bigframes/session/__init__.py
+++ b/bigframes/session/__init__.py
@@ -24,7 +24,6 @@ import secrets
 import typing
 from typing import (
     Any,
-    Callable,
     Dict,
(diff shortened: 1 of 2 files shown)


bigframes/core/blocks.py
```diff
@@ -679,6 +680,7 @@ def to_pandas_batches(
         page_size: Optional[int] = None,
         max_results: Optional[int] = None,
         allow_large_results: Optional[bool] = None,
+        callback: Callable = lambda _: None,
```
tswast, line 683: Per discussion offline, we may want to provide a global progress publisher object rather than add callbacks to each method that needs them.

PR #1740, opened 2025-05-15
chore: move polar and scikit-learn to test extras pacakges

chelsea-lin:


setup.py
```diff
@@ -72,13 +72,12 @@
...
         "freezegun",
-        "pytest-snapshot",
         "google-cloud-bigtable >=2.24.0",
         "google-cloud-pubsub >=2.21.4",
+        "polars >= 1.7.0",
+        "pytest-snapshot",
+        "scikit-learn>=1.2.2",
     ],
-    # used for local engine, which is only needed for unit tests at present.
-    "polars": ["polars >= 1.7.0"],
```
TrevorBergeron, line 80: Might actually be using this stuff for user-facing features soon, so would rather not move to test-only dependencies

chelsea-lin: The `extract` category in `setup.py` is being generalized for better organization, moving away from test-specific configurations. This is primarily a cleanup effort; feel free to add Polars later change if needed.

PR #1741, opened 2025-05-15
test: fix snippets test of using partial ordering mode

GarrettWu:

### DIFF
--- a/samples/snippets/quickstart_test.py
+++ b/samples/snippets/quickstart_test.py
@@ -34,3 +34,6 @@ def test_quickstart(
     quickstart.run_quickstart(your_project_id)
     out, _ = capsys.readouterr()
     assert "average_body_mass:" in out
+
+    # close session so not to affect other tests
+    bigframes.pandas.close_session()


samples/snippets/quickstart_test.py
```diff
@@ -34,3 +34,6 @@ def test_quickstart(
     quickstart.run_quickstart(your_project_id)
     out, _ = capsys.readouterr()
     assert "average_body_mass:" in out
+
+    # close session so not to affect other tests
+    bigframes.pandas.close_session()
```
shobsi, line 39: If the objective is to reset the partially ordered mode, I'm afraid that would not happen here, as the value set stills lingers on, and picked up by the next session created

GarrettWu: right. reset the option and moved to run_quickstart() where the option is changed.

samples/snippets/quickstart.py
```diff
@@ -71,3 +71,7 @@ def run_quickstart(project_id: str) -> None:
     model.fit(X, y)
     model.score(X, y)
     # [END bigquery_bigframes_quickstart]
+
```
shobsi, line 74: This is better approach, thanks!. For additional sanity let's restore the original ordering mode. (The reason for this PR is because multiple tests are sharing the global session, in which case we give back what we got).

```python
    original_ordering_mode = bpd.options.bigquery.ordering_mode

    # [START bigquery_bigframes_quickstart]
    ...
    # [END bigquery_bigframes_quickstart]
    bpd.close_session()
    bpd.options.bigquery.ordering_mode = original_ordering_mode
```

GarrettWu: Lemme know what you think about https://github.com/googleapis/python-bigquery-dataframes/pull/1743

PR #1746, opened 2025-05-16
feat!: add structured output for ai map, ai filter and ai join

sycai:
BEGIN_COMMIT_OVERRIDE
feat!: add structured output for ai map, ai filter and ai join (#1746)

The metric shows that there are no recent external calls to our ai operators, this change should be safe.

Also added:
`logprob` will be available as an output if the user wants to check them.

BREAKING CHANGE: the updated `ai.map()` parameter list is not backward-compatible
Release-As: 2.5.0
END_COMMIT_OVERRIDE

### DIFF
--- a/tests/system/large/operations/test_ai.py
+++ b/tests/system/large/operations/test_ai.py
@@ -66,6 +66,31 @@ def test_filter(session, gemini_flash_model):
     )
 
 
+def test_filter_attach_logprob(session, gemini_flash_model):
+    df = dataframe.DataFrame(
+        data={
+            "number_1": [1, 2],
+            "number_2": [2, 1],
+            "col": [0, 0],
+        },
+        session=session,
+    )
+
+    with bigframes.option_context(
+        AI_OP_EXP_OPTION,
+        True,
+        THRESHOLD_OPTION,
+        10,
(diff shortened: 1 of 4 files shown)


chelsea-lin:
The change only breaks pre-release features, so please add a `Release-As:` footer to the final commit message to make sure this does not trigger the 3.0 release. 
See: https://github.com/googleapis/release-please/blob/main/README.md#how-do-i-change-the-version-number
For example: https://github.com/googleapis/python-bigquery-dataframes/pull/1458

tests/system/large/operations/test_ai.py
```diff
@@ -244,8 +301,8 @@ def test_map_multimodel(session, gemini_flash_model):
         )
         result = df.ai.map(
```
chelsea-lin, line 302: nit: assert the results' columns are expected?

chelsea-lin (approved):
LGTM overall, with leaving two comments.

tests/system/large/operations/test_ai.py
sycai: I would refrain from doing that because the answer is not stable.

PR #1751, opened 2025-05-19
test: Add dataframe unit test suite

TrevorBergeron:


tests/unit/test_dataframe_polars.py
```diff
@@ -0,0 +1,4481 @@
+# Copyright 2023 Google LLC
```
chelsea-lin, line 1: Does this file contain context same as `tests/system/small/test_dataframe.py`? If so, can we reduce code redundancy?

TrevorBergeron: Its a slightly smaller set of tests (maybe like 80% of them?) with some slightly modified validations (eg, stop counting executions). Can unify, but goal is to eventually delete the system test versions of these tests.

chelsea-lin:
The unit-test coverage is decreased from 49% into 46% ([link](https://github.com/googleapis/python-bigquery-dataframes/actions/runs/15355511194/job/43213661476?pr=1751)). Could you please investigate reasons please. Here may be the root causes:
```
tests/unit/test_dataframe_polars.py                           1995   1977     46      0     1%   42-4479
tests/unit/test_local_engine.py                                 89     82      0      0     8%   24-206
```

noxfile.py
```diff
@@ -202,14 +202,11 @@ def install_unittest_dependencies(session, install_test_extra, *constraints):
...
-        extras = UNIT_TEST_EXTRAS_BY_PYTHON.get(session.python, [])
-    if install_test_extra and UNIT_TEST_EXTRAS:
-        extras = UNIT_TEST_EXTRAS
-    else:
-        extras = []
-
-    if extras:
+    if install_test_extra:
+        if session.python in UNIT_TEST_EXTRAS_BY_PYTHON:
+            extras = UNIT_TEST_EXTRAS_BY_PYTHON[session.python]
```
chelsea-lin, line 207: iiuc, this branch should be like
`extras = [**UNIT_TEST_EXTRAS_BY_PYTHON[session.python], **UNIT_TEST_EXTRAS]`.

chelsea-lin (approved):

PR #1765, opened 2025-05-21
refactor: compile the selected_cols for the ResultNode

chelsea-lin:


bigframes/core/compile/sqlglot/sqlglot_ir.py
```diff
@@ -254,3 +261,62 @@ def _table(table: bigquery.TableReference) -> sge.Table:
         db=sg.to_identifier(table.dataset_id, quoted=True),
         catalog=sg.to_identifier(table.project, quoted=True),
     )
+
+
+def _squash_selections(
+    old_expr: list[sge.Expression], new_expr: list[sge.Alias]
+) -> typing.Optional[list[sge.Alias]]:
```
sycai, line 268: Can we return just `list[sge.Alias]`? We can use an empty list to indicate "selection not found / not applicable". 

This is from Java best practices: go/java-practices/null#collection but I think it applies to most languages.

bigframes/core/compile/sqlglot/sqlglot_ir.py
```diff
@@ -254,3 +261,62 @@ def _table(table: bigquery.TableReference) -> sge.Table:
...
+                        old_alias_map[column_alias_pair[0]], quoted=True
+                    )
+                ),
+                alias=sg.to_identifier(column_alias_pair[1], quoted=True),
+            )
+            new_selected_cols.append(new_alias_expr)
+    return new_selected_cols
+
+
+def _try_get_column_alias_pair(
```
sycai, line 304: naming nit: "_get_column_alias_pair" should suffice. The fact that this function can return `None` is already reflected in the return type hint.

bigframes/core/compile/sqlglot/sqlglot_ir.py
```diff
@@ -254,3 +261,62 @@ def _table(table: bigquery.TableReference) -> sge.Table:
...
+    expr: sge.Expression,
+) -> typing.Optional[typing.Tuple[str, str]]:
+    """Checks if an expression is a simple alias of a column definition
+    (e.g., "column_name AS alias_name").
+    If it is, returns a tuple containing the alias name and original column name.
+    Returns `None` otherwise.
+    """
+    if not isinstance(expr, sge.Alias):
```
sycai, line 312: non-rhetorical question: is it possible to check alias at the BFET level? 

We have something like this: https://github.com/googleapis/python-bigquery-dataframes/blob/1f6442e576c35ec784ccf9cab3d081d46e45a5ce/bigframes/core/nodes.py#L1097-L1113

Checking BigFrames objects instead of SQLGlot ones might make or architecture cleaner.

This suggestion is totally optional, though.

chelsea-lin: That's a true concern. I created b/419595879 to revisit the issues considering the complexity of current architecture.

bigframes/core/compile/sqlglot/sqlglot_ir.py
chelsea-lin: Fixed. Thanks!

bigframes/core/compile/sqlglot/sqlglot_ir.py
chelsea-lin: Good to know. Thanks!

sycai (approved):

PR #1777, opened 2025-05-28
chore: use faster query_and_wait API in _read_gbq_colab

tswast:

Fixes internal issue b/405372623 🦕


bigframes/session/loader.py
```diff
@@ -820,18 +827,60 @@ def read_gbq_query(
...
+                # local node. Likely there are a wide range of sizes in which it
+                # makes sense to download the results beyond the first page, even if
+                # there is a job and destination table available.
+            else:
+                # It's possible that we've already downloaded all the results and
+                # there's no job. In this case, we must have a local node.
+                #
+                # This is somewhat wasteful, but we convert from Arrow to pandas
+                # to try to duplicate the same dtypes we'd have if this were a
+                # table node as best we can.
```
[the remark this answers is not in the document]
tswast: Fixed in https://github.com/googleapis/python-bigquery-dataframes/pull/1777/commits/12fd221930fc970d63527480a8fa62d06f162096

bigframes/session/metrics.py
```diff
@@ -79,17 +79,24 @@ def get_performance_stats(
...
     execution_secs = (query_job.ended - query_job.created).total_seconds()
     query_char_count = len(query_job.query)
 
-    return query_char_count, bytes_processed, slot_millis, execution_secs
+    return (
+        query_char_count,
+        # Not every job populates these. For example, slot_millis is missing
+        # from queries that came from cached results.
+        bytes_processed if bytes_processed else 0,
+        slot_millis if slot_millis else 0,
```
TrevorBergeron, line 97: do slightly worry about these zeros being misinterpreted if we ever summarize into averages, but I guess not a problem for now

bigframes/session/loader.py
```diff
@@ -824,27 +829,100 @@ def read_gbq_query(
...
+                (None,),
+            )
+            return dataframe.DataFrame(block)
 
-        # If there was no destination table, that means the query must have
-        # been DDL or DML. Return some job metadata, instead.
+        # If there was no destination table and we've made it this far, that
+        # means the query must have been DDL or DML. Return some job metadata,
+        # instead.
         if not destination:
```
TrevorBergeron, line 907: this function is getting absurdly long, maybe we can pull this job -> stats_df out

bigframes/session/loader.py
```diff
@@ -824,27 +829,100 @@ def read_gbq_query(
...
+            # to canonical formats since it comes from either the BQ Storage
+            # Read API or has been transformed by google-cloud-bigquery to look
+            # like the output of the BQ Storage Read API.
+            pa_table = rows.to_arrow()
+            mat = local_data.ManagedArrowTable(
+                pa_table, schemata.ArraySchema.from_bq_schema(rows.schema)
+            )
+            mat.validate()
+            array_value = core.ArrayValue.from_managed(mat, self._session)
+            array_with_offsets, offsets_col = array_value.promote_offsets()
```
TrevorBergeron, line 895: The promote_offsets makes sense, the only problem is that this will invalidate the local engines until I implement that node type. Might be better, if messier, just to manually add offsets to the pyarrow table for now?

bigframes/session/loader.py
```diff
@@ -912,22 +992,49 @@ def _query_to_destination(
...
+        self,
+        sql: str,
+        job_config: Optional[google.cloud.bigquery.QueryJobConfig] = None,
+        timeout: Optional[float] = None,
+        *,
+        query_with_job: Literal[False],
+    ) -> Tuple[google.cloud.bigquery.table.RowIterator, Optional[bigquery.QueryJob]]:
+        ...
+
+    def _start_query(
```
TrevorBergeron, line 1030: maybe two separate methods would be less code than working around all the mypy nonsense to reuse the same symbol?

bigframes/session/metrics.py
tswast: In some cases (cached query) 0 probably makes sense for the average. I agree that it could be misleading, though.

TrevorBergeron (dismissed):

TrevorBergeron (approved):

tswast:
e2e failure for isdigit in prerelease tests tracked in b/333484335. It's due to a fix in pyarrow that bigframes hasn't been updated to emulate yet.

PR #1780, opened 2025-05-29
fix: include location in Session-based temporary storage manager DDL queries

tswast:
…queries

### DIFF
--- a/tests/system/large/test_location.py
+++ b/tests/system/large/test_location.py
@@ -14,7 +14,6 @@
 
 import typing
 
-from google.cloud import bigquery
 import pandas
 import pandas.testing
 import pytest
@@ -40,7 +39,15 @@ def _assert_bq_execution_location(
     if expected_location is None:
         expected_location = session._location
 
-    assert typing.cast(bigquery.QueryJob, df.query_job).location == expected_location
+    query_job = df.query_job
+    assert query_job is not None
+    assert query_job.location == expected_location
+    destination = query_job.destination
+    assert destination is not None
+    destination_dataset = session.bqclient.get_dataset(
+        f"{destination.project}.{destination.dataset_id}"
+    )
+    assert destination_dataset.location == expected_location
 
     # Ensure operation involving BQ client suceeds
     result = (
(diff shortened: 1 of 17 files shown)


tswast:
Tested manually locally and so far it seems to fix the issue on HEAD.

```
(venv) ➜  python-bigquery-dataframes git:(tswast-session-location) pytest tests/system/large/test_location.py -v
================================================================================================================================= test session starts ==================================================================================================================================
platform linux -- Python 3.12.10, pytest-8.3.5, pluggy-1.6.0 -- /home/swast/src/github.com/googleapis/python-bigquery-dataframes/venv/bin/python
cachedir: .pytest_cache
rootdir: /home/swast/src/github.com/googleapis/python-bigquery-dataframes
configfile: pytest.ini
plugins: snapshot-0.9.0
collected 159 items                                                                                                                                                                                                                                                                    

tests/system/large/test_location.py::test_bq_location_default PASSED                                                                                                                                                                                              [  0%]
tests/system/large/test_location.py::test_bq_location[EU] PASSED                                                                                                                                                                                                  [  1%]
tests/system/large/test_location.py::test_bq_location[US] PASSED                                                                                                                                                                                                  [  1%]
tests/system/large/test_location.py::test_bq_location[africa-south1] PASSED                                                                                                                                                                                       [  2%]
tests/system/large/test_location.py::test_bq_location[asia-east1] PASSED                      
...
```

Genesis929 (dismissed):

tests/system/large/test_location.py
```diff
@@ -63,28 +62,6 @@ def _assert_bq_execution_location(
         expected_result, result.to_pandas(), check_dtype=False, check_index_type=False
     )
 
-    # Ensure BQ Storage Read client operation succceeds
-    table = result.query_job.destination
```
tswast, line 67: This table and query_job might not actually exist / refer to an older object since `to_pandas()` will likely end up in the "jobless" API path.

Genesis929 (dismissed):

tests/system/large/test_location.py
shobsi: I think this code is testing that the result of a BigFrames operation are stored in that location itself, which has value in asserting the location behavior, especially with regional endpoints. Rather than deleting, can we not do one of the following: 

1. Force a query job via `allow_large_results=True` in all tests
2. Move this test code to a special test which enforces `allow_large_results=True`

tswast: I don't think this would verify the location though. You'd have to check that the dataset is in the correct location. Otherwise, this would read from whereever the table is.

tswast: I have forced a job and validate the location of the stored results in 15fa7289e13ce63ae04382c126ac167254d33549.

@Genesis929 @shobsi Please review.

Genesis929 (approved):

tests/system/large/test_location.py
shobsi: With the regional endpoints, the read would fail if the data is in another location?

shobsi: Thanks for checking the location of the destination dataset.

PR #1783, opened 2025-05-30
chore: add snippet tests for type system doc

sycai:
unblock b/407575739

### DIFF
--- a/samples/snippets/type_system_test.py
+++ b/samples/snippets/type_system_test.py
@@ -1,0 +1,285 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def test_type_system_examples():
+    # [START bigquery_dataframes_type_sytem_local_type_conversion]
+    import bigframes.pandas as bpd
+    import pandas as pd
+
+    s = pd.Series([pd.Timestamp('20250101')])
+    print(s.dtype)
+    # datetime64[ns]
+    print(bpd.read_pandas(s).dtype) 
+    # timestamp[us][pyarrow]
+    # [END bigquery_dataframes_type_sytem_local_type_conversion]
+
+
+    # [START bigquery_dataframes_type_system_pyarrow_preference]
+    import bigframes.pandas as bpd
+    import pandas as pd
+    import datetime
+
+    s = pd.Series([datetime.date(2025,1,1)])
+    s + pd.Timedelta(hours=12)
+    # 0	2025-01-01
+    # dtype: object
+
+    bpd.read_pandas(s) + pd.Timedelta(hours=12)
+    # 0    2025-01-01 12:00:00
+    # dtype: timestamp[us][pyarrow]
+    # [END bigquery_dataframes_type_system_pyarrow_preference]
+
+
+    # [START bigquery_dataframes_type_system_simple_json]
+    import bigframes.pandas as bpd
+    import db_dtypes
+    import pandas as pd
+
+    json_data = [
+        "1",
+        '"str"',
+        "false",
+        '["a", {"b": 1}, null]',
+        '{"a": {"b": [1, 2, 3], "c": true}}',
+        None,
+    ]
+    bpd.Series(json_data, dtype=pd.ArrowDtype(db_dtypes.JSONArrowType()))
+    # 0                               1
+    # 1                           "str"
+    # 2                           false
+    # 3              ["a",{"b":1},null]
+    # 4    {"a":{"b":[1,2,3],"c":true}}
+    # 5                            <NA>
+    # dtype: extension<dbjson<JSONArrowType>>[pyarrow]
+    # [END bigquery_dataframes_type_system_simple_json]
+
+
+    # [START bigquery_dataframes_type_system_mixed_json]
+    import bigframes.pandas as bpd
+    import db_dtypes
+    import pandas as pd
+    import pyarrow as pa
+
+    list_data = [
+        [{"key": "1"}],
+        [{"key": None}],
+        [{"key": '["1","3","5"]'}],
+        [{"key": '{"a":1,"b":["x","y"],"c":{"x":[],"z":false}}'}],
+    ]
+    pa_array = pa.array(list_data, type=pa.list_(pa.struct([("key", pa.string())])))
+    bpd.Series(
+        pd.arrays.ArrowExtensionArray(pa_array),
+        dtype=pd.ArrowDtype(
+            pa.list_(pa.struct([("key", db_dtypes.JSONArrowType())])),
+        ),
+    )
+    # 0                                       [{'key': '1'}]
+    # 1                                      [{'key': None}]
+    # 2                           [{'key': '["1","3","5"]'}]
+    # 3    [{'key': '{"a":1,"b":["x","y"],"c":{"x":[],"z"...
+    # dtype: list<item: struct<key: extension<dbjson<JSONArrowType>>>>[pyarrow]
+    # [END bigquery_dataframes_type_system_mixed_json]
+
+
+    # [START bigquery_dataframes_type_system_load_timedelta]
+    import bigframes.pandas as bpd
+    import pandas as pd
+
+    s = pd.Series([pd.Timedelta('1s'), pd.Timedelta('2m')])
+    bpd.read_pandas(s)
+    # 0    0 days 00:00:01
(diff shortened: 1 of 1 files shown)


samples/snippets/type_system_test.py
```diff
@@ -0,0 +1,284 @@
...
+    # [START bigquery_dataframes_type_sytem_local_type_conversion]
+    import pandas as pd
+
+    import bigframes.pandas as bpd
+
+    s = pd.Series([pd.Timestamp("20250101")])
+    print(s.dtype)
+    # datetime64[ns]
+    print(bpd.read_pandas(s).dtype)
+    # timestamp[us][pyarrow]
```
tswast, line 26: Might be good to confirm that these comments are accurate with some assertions. Otherwise, we won't know when our docs are incorrect and we're possibly making breaking changes.

```suggestion
    pandas_dtype = s.dtype
    assert str(pandas_dtype) == "datetime64[ns]"
    bigframes_dtype = bpd.read_pandas(s).dtype
    assert str(bigframes_dtype) == "timestamp[us][pyarrow]"
```

Same applies to the other samples where we want to demonstrate bigframes behavior.

In the cases where an assertion in the code sample itself doesn't make sense, we can put the assertions after the `[END ...]` comment and before the next sample starts.

samples/snippets/type_system_test.py
```diff
@@ -0,0 +1,284 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def test_type_system_examples() -> None:
+    # [START bigquery_dataframes_type_sytem_local_type_conversion]
```
tswast, line 17: Region tags need to be globally unique. Maybe it makes more sense to include "timestamp" somewhere in here? I imagine `bigquery_dataframes_type_sytem_local_type_conversion` could apply to lots of different types beyond the scope of this sample.

samples/snippets/type_system_test.py
sycai: Sure. Tests added.

samples/snippets/type_system_test.py
sycai: Right. Changed it to "..._timestamp_local_type_conversion"

samples/snippets/set_options_test.py
```diff
@@ -39,3 +39,6 @@ def test_bigquery_dataframes_set_options() -> None:
     # [END bigquery_dataframes_set_options]
     assert bpd.options.bigquery.project == PROJECT_ID
     assert bpd.options.bigquery.location == REGION
+
+    bpd.close_session()
+    bpd.options.reset()
```
tswast, line 44: Let's put anything that changes global options in a try/finally block so that this cleanup can run no matter what.

sycai: SG!

samples/snippets/type_system_test.py
```diff
@@ -0,0 +1,357 @@
...
+        None,
+    ]
+    bpd.Series(json_data, dtype=pd.ArrowDtype(db_dtypes.JSONArrowType()))
+    # 0                               1
+    # 1                           "str"
+    # 2                           false
+    # 3              ["a",{"b":1},null]
+    # 4    {"a":{"b":[1,2,3],"c":true}}
+    # 5                            <NA>
+    # dtype: extension<dbjson<JSONArrowType>>[pyarrow]
```
tswast, line 77: If we show this, let's include a comment with a disclaimer that this will be switching to pyarrow json_ type in future.

https://github.com/googleapis/python-bigquery-dataframes/blob/1d4564604baff612c3455fb088e442198084bf26/bigframes/dtypes.py#L65-L66

CC @chelsea-lin

samples/snippets/type_system_test.py
```diff
@@ -0,0 +1,357 @@
...
+        [{"key": "1"}],
+        [{"key": None}],
+        [{"key": '["1","3","5"]'}],
+        [{"key": '{"a":1,"b":["x","y"],"c":{"x":[],"z":false}}'}],
+    ]
+    pa_array = pa.array(list_data, type=pa.list_(pa.struct([("key", pa.string())])))
+    bpd.Series(
+        pd.arrays.ArrowExtensionArray(pa_array),
+        dtype=pd.ArrowDtype(
+            pa.list_(pa.struct([("key", db_dtypes.JSONArrowType())])),
```
tswast, line 104: Same here. I think we'll always allow this for backwards compatibility but we might convert this to the arrow built-in json_ type in future.

CC @chelsea-lin 

Maybe we suggest folks use `bigframes.dtypes.JSON_ARROW_TYPE` instead?

samples/snippets/type_system_test.py
tswast: Maybe we suggest folks use `bigframes.dtypes.JSON_DTYPE` instead?

samples/snippets/type_system_test.py
```diff
@@ -0,0 +1,357 @@
...
+
+    import bigframes.bigquery as bbq
+    import bigframes.pandas as bpd
+
+    fruits = [
+        '{"fruits": [{"name": "apple"}, {"name": "cherry"}]}',
+        '{"fruits": [{"name": "guava"}, {"name": "grapes"}]}',
+    ]
+
+    json_s = bpd.Series(fruits, dtype=pd.ArrowDtype(db_dtypes.JSONArrowType()))
```
tswast, line 323: Same here re: JSON_DTYPE maybe changing in future.

samples/snippets/type_system_test.py
chelsea-lin: Agree. Let's use `bigframes.dtypes.JSON_DTYPE`. Further more, we may want to create a more official APIs for these dtypes, similar to: https://github.com/pandas-dev/pandas/blob/v2.3.0/pandas/core/arrays/integer.py#L228-L232

samples/snippets/type_system_test.py
```diff
@@ -0,0 +1,357 @@
...
+
+    json_data = [
+        "1",
+        '"str"',
+        "false",
+        '["a",{"b":1},null]',
+        '{"a":{"b":[1,2,3],"c":true}}',
+        None,
+    ]
+    # db_dtypes.JSONArrowType() might be changed to pa.json_(pa.string()) in the future
```
tswast, line 70: Let's rephrase to make clear that we intend to support both for at least some time.

```suggestion
    # In addition to db_dtypes.JSONArrowType(), support for pa.json_(pa.string())
    # will be added once the ArrowDtype supports this extension type in pandas 3.0.0.
    # See: https://github.com/pandas-dev/pandas/pull/61103
```

CC @chelsea-lin I wonder if there's anything we can do in the meantime to prepare to support this? Maybe it just works already because of our struct and array of JSON support?

samples/snippets/type_system_test.py
```diff
@@ -0,0 +1,357 @@
...
+    import bigframes.pandas as bpd
+
+    list_data = [
+        [{"key": "1"}],
+        [{"key": None}],
+        [{"key": '["1","3","5"]'}],
+        [{"key": '{"a":1,"b":["x","y"],"c":{"x":[],"z":false}}'}],
+    ]
+    pa_array = pa.array(list_data, type=pa.list_(pa.struct([("key", pa.string())])))
+    # db_dtypes.JSONArrowType() might be changed to pa.json_(pa.string()) in the future
```
tswast, line 101: I thought we already supported `pa.json_(pa.string())` in structs? If not, let's update this similarly:

```suggestion
    # In addition to db_dtypes.JSONArrowType(), support for pa.json_(pa.string())
    # will be added once the ArrowDtype supports this extension type in pandas 3.0.0.
    # See: https://github.com/pandas-dev/pandas/pull/61103
```

samples/snippets/type_system_test.py
```diff
@@ -0,0 +1,359 @@
...
+
+    fruits = [
+        '{"fruits": [{"name": "apple"}, {"name": "cherry"}]}',
+        '{"fruits": [{"name": "guava"}, {"name": "grapes"}]}',
+    ]
+
+    # db_dtypes.JSONArrowType() might be changed to pa.json_(pa.string()) in the future
+    json_s = bpd.Series(fruits, dtype=pd.ArrowDtype(db_dtypes.JSONArrowType()))
+
+    bbq.json_extract_array(json_s, "$.fruits")
```
tswast, line 353: This is deprecated: https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_extract_array

We're supposed to use json_query_array, instead.

samples/snippets/type_system_test.py
```diff
@@ -0,0 +1,359 @@
...
+
+    import bigframes.bigquery as bbq
+    import bigframes.pandas as bpd
+
+    fruits = [
+        '{"fruits": [{"name": "apple"}, {"name": "cherry"}]}',
+        '{"fruits": [{"name": "guava"}, {"name": "grapes"}]}',
+    ]
+
+    # db_dtypes.JSONArrowType() might be changed to pa.json_(pa.string()) in the future
```
tswast, line 350: Same comment here re: being more explicit about what combination of pandas + pyarrow this will change.

tswast (approved):

PR #1798, opened 2025-06-06
feat: add groupby cumcount

TrevorBergeron:


tests/system/small/test_groupby.py
```diff
@@ -401,16 +401,44 @@ def test_dataframe_groupby_multi_sum(
...
+    )
     bf_result_computed = bf_result.to_pandas()
 
     pd.testing.assert_frame_equal(pd_result, bf_result_computed, check_dtype=False)
 
 
+@pytest.mark.parametrize(
+    ("ascending", "dropna"),
+    [
+        (True, True),
```
shobsi, line 424: why leave (False, True) and (True, False)?

TrevorBergeron: ah, just that they are orthogonal, and mostly just wanted to test both modes for each flag, but there isn't really any interaction

shobsi: I see, might be worth adding a note (if not adding the combinations)

shobsi (approved):

PR #1804, opened 2025-06-10
fix: correct read_csv behaviours with use_cols, names, index_col

chelsea-lin:
Fixes internal issue 421466334 🦕

### DIFF
--- a/bigframes/session/loader.py
+++ b/bigframes/session/loader.py
@@ -550,6 +550,7 @@ class GbqDataLoader:
         )
 
         columns = list(columns)
+        include_all_columns = columns is None or len(columns) == 0
         filters = typing.cast(list, list(filters))
 
         # ---------------------------------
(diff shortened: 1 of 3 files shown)


bigframes/session/loader.py
```diff
@@ -601,18 +589,67 @@ def read_gbq_table(
...
+                        possibility = min(
+                            table_column_names,
+                            key=lambda item: bigframes._tools.strings.levenshtein_distance(
+                                column_name, item
+                            ),
+                        )
+                        raise ValueError(
+                            f"Column '{column_name}' of `columns` not found in this table. "
+                            f"Did you mean '{possibility}'?"
+                        )
```
sycai, line 644: nit: Shall we use a helper function/method? The indentation is very deep here. (https://goto.google.com/tott/733)

bigframes/session/loader.py
```diff
@@ -601,18 +589,67 @@ def read_gbq_table(
...
+                    if set(list(names)) != set(list(columns)):
+                        raise ValueError("Usecols do not match columns")
+
+        rename_to_schema: Optional[Dict[str, str]] = None
+        if len(columns) == 0:
+            table_column_names = [field.name for field in table.schema]
+            if names is not None:
+                rename_to_schema = dict(zip(list(names), table_column_names))
+        else:
+            if names is not None:
```
sycai, line 614: It feels like this is the same as `elif names is not None:` which can save you a level of indentation.

bigframes/session/loader.py
```diff
@@ -601,18 +589,67 @@ def read_gbq_table(
...
+                invalid_columns = set(columns) - set(names)
+                if len(invalid_columns) != 0:
+                    raise ValueError(
+                        "Usecols do not match columns, columns expected but not "
+                        f"found: {invalid_columns}"
+                    )
+
+                rename_to_schema = dict(zip(list(names), table_column_names))
+                names = columns
+                columns = [rename_to_schema[renamed_name] for renamed_name in columns]
```
sycai, line 630: The hosting function is very long. I think there might be an opportunity to make this code block a helper function/method

go/pystyle#function-length

bigframes/session/loader.py
chelsea-lin: Done.

bigframes/session/loader.py
chelsea-lin: Done.

bigframes/session/loader.py
chelsea-lin: Refactored for more readable. Please check.

sycai (dismissed):

sycai (approved):

PR #1819, opened 2025-06-13
chore: make sure SQLGlot scalar compiler doesn't encounter DerefOp

sycai:
The key changes are:
* Postpone schema binding until the last moment
* Provide ways to handle DerefOps embedded in the ResultNode and SelectionNode. These references cannot be updated with the schema binding function because their types are explicitly defined.


sycai:
This change might be unnecessary because all we care about are opExpression types.

PR #1827, opened 2025-06-16
feat: add describe() method to Series

sycai:
Moved the describe() logic from `DataFrame` to a separate library file so that it can be re-used by `Series` too.

Fixes #385141970


chelsea-lin (commented):
we probably want to update docs too. Moving `describe` method from `third_party/bigframes_vendored/pandas/core/frame.py` to `common.py`.

chelsea-lin (approved):

jialuoo (approved):

PR #1832, opened 2025-06-17
feat: create `deploy_remote_function` and `deploy_udf` functions to immediately deploy functions to BigQuery

tswast:


tswast:
CC @ivansmf -- this PR should add those aliases for your sample.

tswast:
> We already have two ways for each - the decorator @bpd.udf and the function call bpd.udf(...)(func), why to add a third way - an extra API with very little value add?

1. To be explicit that we're deploying immediately. With `udf` and `remote_function` we may choose to defer deployment in future.
2. To follow the general convention of `verb_plus_description` for functions. (https://thepythoncodingbook.com/2023/01/18/best-practices-in-python-functions/#:~:text=A%20function%20performs%20an%20action,describe%20what%20the%20function%20does.)

@ivansmf for additional thoughts.

shobsi (approved):

PR #1838, opened 2025-06-20
fix: revert dict back to protobuf in the iam binding update

shobsi:
This is needed as some recent updates to a dependency (google-cloud-resource-manager or its dependencies) broke the existing logic, and we are seeing the error like this:

```
--> policy.bindings.append(new_binding)
TypeError: Expected a message object, but got {'role': 'roles/run.invoker', 'members': [...]}
```

Fixes internal issues 419589974, 425442503 🦕


shobsi:
@tswast looks like the `google.iam` usage was done away with in https://github.com/googleapis/python-bigquery-dataframes/pull/1611, so this change is effectively reverting that. Let's discuss how should we keep things working in both external and internal environments

bigframes/clients.py
```diff
@@ -172,10 +173,9 @@ def _ensure_iam_binding(
                     return
 
         # Create a new binding
-        new_binding = {
-            "role": role,
-            "members": [service_account],
-        }  # Use a dictionary to avoid problematic google.iam namespace package.
+        new_binding = google.iam.v1.policy_pb2.Binding(
+            role=role, members=[service_account]
+        )  # Use a dictionary to avoid problematic google.iam namespace package.
```
tswast, line 178: This comment is now out-of-date.

```suggestion
        )
```

Could you try this in google3? I wonder if a `//google/iam/v1:iam_policy_py_pb2` dependency was missing last time I tried it?

tswast (commented):
I'm a bit concerned that this indicates a breaking change happened in the resource manager client library. Please find the change that broke dictionaries. We might need to update our dependencies in `setup.py` and the constraints files if this breaking change was intended.

bigframes/clients.py
shobsi: Trying. That one should be indirectly included via resourcemanager_v3 (which does include it)?

shobsi:
> CC @parthea Did the gapic generator make a breaking change to no longer allow dictionary inputs?

I tested that even with google-cloud-resource-manager==1.12.0 (more than an year old) the dictionary update does not work. So the break happened with the proto->dictionary change. It is a case of missing test coverage for the IAM update code path (which is not integration tested due to internal test project policies where IAM permissions are managed via config). I am adding a mocked unit test which covers at least the binding update part.

tswast (approved):

PR #1841, opened 2025-06-23
feat: Add pagination buttons (prev/next) to anywidget mode for DataFrames

shuoweil:
Adding interactive pagination buttons controls (Previous/Next buttons) to the anywidget mode for BigFrames DataFrames, enabling users to navigate through large datasets page by page in Jupyter notebooks.

b/391599608


bigframes/display/anywidget.py
```diff
@@ -0,0 +1,179 @@
...
+
+class TableWidget(anywidget.AnyWidget):
+    """
+    An interactive, paginated table widget for BigFrames DataFrames.
+    """
+
+    # The _esm variable contains the JavaScript source code for the frontend
+    # component of the widget.
+    _esm = """
+    function render({ model, el }) {
```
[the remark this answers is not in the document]
shuoweil: Thanks for the suggestion, Tim. I will separate the javascript out.

bigframes/display/table_widget.js
tswast, line 1: So cool! Thanks for adding.

Could you also add a formatter / linter to our pre-commit (https://github.com/googleapis/python-bigquery-dataframes/blob/main/.pre-commit-config.yaml)

https://github.com/biomejs/pre-commit?tab=readme-ov-file#using-biomes-pre-commit-hooks seems a good option.

```
repos:
  - repo: https://github.com/biomejs/pre-commit
    rev: v2.0.2 # Use the sha / tag you want to point at
    hooks:
      - id: biome-check
```

It's the first listed under "web" at https://pre-commit.com/hooks.html#featured-hooks

tests/system/small/test_anywidget.py
```diff
@@ -0,0 +1,122 @@
...
+
+import pytest
+
+import bigframes as bf
+
+
+def test_repr_anywidget_initial_state(
+    penguins_df_default_index: bf.dataframe.DataFrame,
+):
+    pytest.importorskip("anywidget")
```
tswast, line 25: Nit: Since all tests in this file require anywidget, let's move the importorskip to the module level.

I think this should work.

```suggestion


pytest.importorskip("anywidget")


def test_repr_anywidget_initial_state(
    penguins_df_default_index: bf.dataframe.DataFrame,
):
```

Same for the other functions in this file.

tests/system/small/test_anywidget.py
```diff
@@ -0,0 +1,122 @@
...
+        from bigframes.display.anywidget import TableWidget
+
+        widget = TableWidget(penguins_df_default_index)
+
+        # Test initial state
+        assert widget.page == 0
+
+        # Simulate next page click
+        widget.page = 1
+        assert widget.page == 1
```
tswast, line 50: How do we tell if the data displayed has changed? Can we check the table HTML that it doesn't contain the first page of rows and that it does contain the second page?

Rather than the penguin data, it might be better to define our own DataFrame in this test and set a really small page size so we can make sure we visualize the correct rows. See: https://testing.googleblog.com/2018/02/testing-on-toilet-cleanly-create-test.html

> tests should never rely on default values that are specified by a helper method since that forces readers to read the helper method’s implementation details in order to understand the test.

tests/system/small/test_anywidget.py
```diff
@@ -0,0 +1,122 @@
...
+        total_pages = math.ceil(widget.row_count / 5)
+        assert total_pages > 1  # Should have multiple pages
+
+        # Navigate through several pages
+        for page in range(min(3, total_pages)):
+            widget.page = page
+            assert widget.page == page
+
+
+def test_repr_anywidget_pagination_buttons_functionality(
```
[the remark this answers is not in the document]
shuoweil: I used to want to uses a loop instead of explicit page checks. However, for loop is not a good practice for unit test. I have removed this testcase.

tests/system/small/test_anywidget.py
```diff
@@ -0,0 +1,235 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pandas as pd
+import pytest
+
+import bigframes as bf
+from bigframes.display import TableWidget
```
tswast, line 19: We're getting the following error in our e2e tests

```
____________ ERROR collecting tests/system/small/test_anywidget.py _____________
ImportError while importing test module '/tmpfs/src/github/python-bigquery-dataframes/tests/system/small/test_anywidget.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
[tests/system/small/test_anywidget.py:19](https://cs.corp.google.com/piper///depot/google3/tests/system/small/test_anywidget.py?l=19): in <module>
    from bigframes.display import TableWidget
E   ImportError: cannot import name 'TableWidget' from 'bigframes.display' (/tmpfs/src/github/python-bigquery-dataframes/bigframes/display/__init__.py)
```

Also,

> Use import statements for packages and modules only, not for individual types, classes, or functions.

https://google.github.io/styleguide/pyguide.html#22-imports The only exceptions are the built-in typing module and a few other common built-in base types.

I recommend moving the imports to inside the function body as you've already done for some tests.

For type checking, you can do this:

```
from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from bigframes import display
```

bigframes/display/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+
+try:
+    import anywidget  # noqa
+
+    from bigframes.display.anywidget import TableWidget
+
+    __all__ = ["TableWidget"]
+except Exception:
+    msg = "Anywidget mode not available as anywidget is not installed."
+    warnings.warn(msg)
```
tswast, line 27: This is an inappropriate time to raise a warning because users might not be able to avoid it. Better to do the warnings / errors when they explicitly do something like set the `repr_mode`.

bigframes/display/anywidget.py
```diff
@@ -0,0 +1,133 @@
...
+
+from importlib import resources
+import functools
+import math
+from typing import Iterator
+import uuid
+
+import anywidget  # type: ignore
+import pandas as pd
+import traitlets
```
tswast, line 23: This needs to be added as an explicit dependency in https://github.com/googleapis/python-bigquery-dataframes/blob/main/setup.py if we are using it directly.

If it is optional, then we should make it part of the extras and make sure we do conditional imports too.

shuoweil: I set it as optional

bigframes/display/__init__.py
shuoweil: I move this warning to the proper place.

bigframes/display/anywidget.py
```diff
@@ -0,0 +1,158 @@
...
+    import traitlets
+else:
+    try:
+        import anywidget
+        import traitlets
+    except Exception:
+        ANYWIDGET_INSTALLED = False
+
+
+class TableWidget(anywidget.AnyWidget):
```
tswast, line 39: Presumably this line would fail if AnyWidget is not installed.

Maybe we can put the whole class definition inside an if block instead? Something like:

```python
if ANYWIDGET_INSTALLED:
    class TableWidget(anywidget.AnyWidget):
        ...
```

Alternatively, we could define a fallback for AnyWidget:

```python
if ANYWIDGET_INSTALLED:
    WIDGET_BASE = anywidget.AnyWidget
else:
    WIDGET_BASE = object


class TableWidget(WIDGET_BASE):
    ...
```

tests/system/small/test_anywidget.py
```diff
@@ -0,0 +1,236 @@
...
+import pytest
+
+import bigframes as bf
+
+pytest.importorskip("anywidget")
+
+
+@pytest.fixture(scope="module")
+def paginated_pandas_df() -> pd.DataFrame:
+    """Create a minimal test DataFrame with exactly 3 pages of 2 rows each."""
```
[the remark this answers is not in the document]
shuoweil: The TableWidget should not be affected by global option changes after creation. Other tests are added.

bigframes/display/anywidget.py
```diff
@@ -0,0 +1,164 @@
...
+        try:
+            iterator = self._get_batch_iterator()
+            batch = next(iterator)
+            self._cached_data = pd.concat([self._cached_data, batch], ignore_index=True)
+            return True
+        except StopIteration:
+            self._all_data_loaded = True
+            # update row count if we loaded all data
+            if self.row_count == 0:
+                self.row_count = len(self._cached_data)
```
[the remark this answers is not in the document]
shuoweil: The fall back logic is removed.

bigframes/dataframe.py
```diff
@@ -777,20 +777,19 @@ def _repr_html_(self) -> str:
...
+                # Return a fallback HTML string
+                return "Interactive table widget (anywidget mode)"
+            except (AttributeError, ValueError):
+                # Fallback if anywidget is not available
+                warnings.warn(
+                    "Anywidget mode is not available, falling back to deferred mode."
+                )
+                return formatter.repr_query_job(self._compute_dry_run())
 
         self._cached()
```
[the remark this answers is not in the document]
shuoweil: Here is my change:
(1) Move blob processing before mode check: The blob column processing from now happens before the anywidget mode check, ensuring blob URLs are generated regardless of display mode.
(2) Pass processed DataFrame to widget: Instead of passing self to TableWidget, we pass the processed df that already has blob URLs resolved.
(3) Maintain feature parity: The anywidget mode now inherits all the blob display capabilities that were previously only available in the default HTML mode.

This approach ensures that when anywidget becomes the default mode, users won't lose the multimodal data display capabilities that are critical for working with images, videos, and other blob content as shown in.

bigframes/dataframe.py
```diff
@@ -803,7 +793,40 @@ def _repr_html_(self) -> str:
...
+
+                from bigframes import display
+
+                # Check if widget instance already exists and reuse it
+                widget = None
+                if (
+                    hasattr(self, "_anywidget_instance")
+                    and self._anywidget_instance is not None
+                ):
+                    widget = self._anywidget_instance()
```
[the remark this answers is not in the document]
shuoweil: It will display two widgets, see here: screen/3eWzK2NSEoPn56o

bigframes/dataframe.py
```diff
@@ -111,6 +113,9 @@ class DataFrame(vendored_pandas_frame.DataFrame):
     # Must be above 5000 for pandas to delegate to bigframes for binops
     __pandas_priority__ = 15000
 
+    # Type annotation for anywidget instance
+    _anywidget_instance: Optional[weakref.ReferenceType["TableWidget"]] = None
```
tswast, line 117: I thought we discussed that having the DataFrame store a widget isn't necessary anymore?

tswast (approved):
So cool! Thanks.

PR #1842, opened 2025-06-23
chore!: remove ai.top_k().

sycai:
We will add it back once AI scoring is ready on the SQL engine side.

BEGIN_COMMIT_OVERRIDE
BREAKING CHANGE: ai.top_k() is removed.
Release-As: 2.9.0
END_COMMIT_OVERRIDE


GarrettWu:
Will it be a different API or the same API for AI scoring?

sycai:
> Will it be a different API or the same API for AI scoring?

We will probably provide both `ai.score()` and `ai.top_k()`, but the parameter list remains to be decided.  

Another concern is that the current approach (divide-and-conquer) is unlikely to be consistent with AI scoring. Due to the mercuriality of the AI responses, there's a chance that the current approach gets stuck indefinitely. We'd better play it safe.

GarrettWu (approved):

PR #1857, opened 2025-06-26
feat: df.join lsuffix and rsuffix support

Genesis929:


bigframes/dataframe.py
```diff
@@ -3412,37 +3412,72 @@ def join(
...
                 right_join_ids=[],
-                suffixes=("", ""),
+                suffixes=(lsuffix, rsuffix),
                 how="cross",
                 sort=True,
             )
             return DataFrame(result_block)
 
         # Join left columns with right index
         if on is not None:
```
tswast, line 3445: Nit: This `if` block is getting pretty long. Might be time for a helper function.

tests/system/small/test_dataframe.py
```diff
@@ -2816,12 +2816,99 @@ def test_join_different_table(
...
+    bf_result = bf_df_a.join(bf_df_b, how=how, lsuffix="_l", rsuffix="_r").to_pandas()
+    print(bf_result)
+    pd_df_a = scalars_pandas_df_index[["string_col", "int64_col", "int64_too"]].rename(
+        columns={"int64_too": "int64_col"}
+    )
+    pd_df_b = scalars_pandas_df_index.dropna()[
+        ["string_col", "int64_col", "int64_too"]
+    ].rename(columns={"int64_too": "int64_col"})
+    pd_result = pd_df_a.join(pd_df_b, how=how, lsuffix="_l", rsuffix="_r")
+    print(pd_result)
```
tswast, line 2838: Remove leftover print() statements.

PS. Adding `--pdb` to your `pytest` command line arguments makes dropping into a debugger to inspect variables really easy. https://docs.pytest.org/en/stable/how-to/failures.html#dropping-to-pdb-on-failures

tests/system/small/test_dataframe.py
```diff
@@ -2816,12 +2816,99 @@ def test_join_different_table(
...
+    scalars_df_index, scalars_pandas_df_index, how
+):
+    bf_df_a = scalars_df_index[["string_col", "int64_col", "int64_too"]].rename(
+        columns={"int64_too": "int64_col"}
+    )
+    bf_df_b = scalars_df_index.dropna()[
+        ["string_col", "int64_col", "int64_too"]
+    ].rename(columns={"int64_too": "int64_col"})
+    bf_result = bf_df_a.join(bf_df_b, how=how, lsuffix="_l", rsuffix="_r").to_pandas()
+    print(bf_result)
```
tswast, line 2830: Remove

tests/system/small/test_dataframe.py
```diff
@@ -2816,12 +2816,99 @@ def test_join_different_table(
...
+    bf_df_a = scalars_df_index[
+        ["string_col", "datetime_col", "timestamp_col", "int64_too"]
+    ].rename(columns={"timestamp_col": "datetime_col"})
+    bf_df_b = scalars_df_index.dropna()[
+        ["string_col", "datetime_col", "timestamp_col", "int64_too"]
+    ].rename(columns={"timestamp_col": "datetime_col"})
+    bf_result = bf_df_a.join(
+        bf_df_b, on="int64_too", how=how, lsuffix="_l", rsuffix="_r"
+    ).to_pandas()
+    print(bf_result)
```
tswast, line 2895: remove print

tests/system/small/test_dataframe.py
```diff
@@ -2816,12 +2816,99 @@ def test_join_different_table(
...
+    pd_df_a = scalars_pandas_df_index[
+        ["string_col", "datetime_col", "timestamp_col", "int64_too"]
+    ].rename(columns={"timestamp_col": "datetime_col"})
+    pd_df_b = scalars_pandas_df_index.dropna()[
+        ["string_col", "datetime_col", "timestamp_col", "int64_too"]
+    ].rename(columns={"timestamp_col": "datetime_col"})
+    pd_result = pd_df_a.join(
+        pd_df_b, on="int64_too", how=how, lsuffix="_l", rsuffix="_r"
+    )
+    print(pd_result)
```
tswast, line 2905: remove print

tests/unit/test_dataframe_polars.py
```diff
@@ -2443,12 +2443,36 @@ def test_join_different_table(
...
+@all_joins
+def test_join_raise_when_param_on_duplicate_with_column(scalars_df_index, how):
+    if how == "cross":
+        return
+    bf_df_a = scalars_df_index[["string_col", "int64_col"]].rename(
+        columns={"int64_col": "string_col"}
+    )
+    bf_df_a.index.name = "string_col"
+    bf_df_b = scalars_df_index.dropna()["string_col"]
+    with pytest.raises(ValueError):
```
tswast, line 2455: Please add a `match=` argument so we know the expected `ValueError` is reached, not a different one by accident. https://docs.pytest.org/en/stable/reference/reference.html#pytest.raises

tests/unit/test_dataframe_polars.py
```diff
@@ -2443,12 +2443,36 @@ def test_join_different_table(
...
+        bf_df_a.join(bf_df_b, on="string_col", how=how, lsuffix="_l", rsuffix="_r")
+
+
+def test_join_duplicate_columns_raises_value_error(scalars_dfs):
     scalars_df, _ = scalars_dfs
     df_a = scalars_df[["string_col", "float64_col"]]
     df_b = scalars_df[["float64_col"]]
-    with pytest.raises(NotImplementedError):
-        df_a.join(df_b, how="outer").to_pandas()
+    with pytest.raises(ValueError):
```
tswast, line 2463: Same here. Please add a `match` argument.

tests/unit/test_dataframe_polars.py
```diff
@@ -2443,12 +2443,36 @@ def test_join_different_table(
...
+
+
+@all_joins
+def test_join_param_on_duplicate_with_index_raises_value_error(scalars_df_index, how):
+    if how == "cross":
+        return
+    bf_df_a = scalars_df_index[["string_col"]]
+    bf_df_a.index.name = "string_col"
+    bf_df_b = scalars_df_index.dropna()["string_col"]
+    with pytest.raises(ValueError):
```
tswast, line 2474: Here as well.

tests/unit/test_dataframe_polars.py
```diff
@@ -2443,12 +2443,36 @@ def test_join_different_table(
...
-    with pytest.raises(NotImplementedError):
-        df_a.join(df_b, how="outer").to_pandas()
+    with pytest.raises(ValueError):
+        df_a.join(df_b, how="outer")
+
+
+@all_joins
+def test_join_param_on_duplicate_with_index_raises_value_error(scalars_df_index, how):
+    if how == "cross":
+        return
```
tswast, line 2470: I think it'd be worth added a test that `ValueError` is _not_ raise for this condition with a cross join.

Genesis929: Cross join actually raise another error, match added.

bigframes/dataframe.py
Genesis929: Added _join_on_key function.

bigframes/dataframe.py
```diff
@@ -3485,70 +3485,138 @@ def join(
...
+        left, right = self, other
+        # Replace all columns names with unique names for reordering.
+        left_col_original_names = left.columns
+        on_col_name = "bigframes_left_col_on"
+        dup_on_col_name = "bigframes_left_col_on_dup"
+        left_col_temp_names = [
+            f"bigframes_left_col_name_{i}" if col_name != on else on_col_name
+            for i, col_name in enumerate(left_col_original_names)
+        ]
+        left.columns = pandas.Index(left_col_temp_names)
```
tswast, line 3571: This seems dangerous. We haven't made a copy of self, so I'm uncomfortable with mutating it. If we must do this, then please either:

1. make a copy of self first
2. or put a `finally` block that resets the names back to the original in case anything when wrong.

I prefer (1) since it's less likely to have problems in we're in a multi-threaded environment.

bigframes/dataframe.py
```diff
@@ -3485,70 +3485,138 @@ def join(
...
+            left_col_temp_names = [on_col_name] + left_col_temp_names
+            left = left[left_col_temp_names]
+
+        # Switch left index with on column
+        left_idx_original_names = left.index.names if left._has_index else ()
+        left_idx_names_in_cols = [
+            f"bigframes_left_idx_name_{i}" for i in range(len(left_idx_original_names))
+        ]
+        if left._has_index:
+            left.index.names = left_idx_names_in_cols
```
tswast, line 3586: Same here. Mutating the index is dangerous. Can we avoid this?

bigframes/dataframe.py
```diff
@@ -3485,70 +3485,138 @@ def join(
...
+        left = left.reset_index(drop=False)
+        left = left.set_index(on_col_name)
+
+        right_col_original_names = right.columns
+        right_col_temp_names = [
+            f"bigframes_right_col_name_{i}"
+            for i in range(len(right_col_original_names))
+        ]
+        right.columns = pandas.Index(right_col_temp_names)
```
tswast, line 3595: Same here.

bigframes/dataframe.py
```diff
@@ -3562,6 +3630,30 @@ def _perform_join_by_index(
         )
         return DataFrame(block)
 
+    def _add_join_suffix(
+        self,
+        left_columns,
+        right_columns,
+        lsuffix: str = "",
+        rsuffix: str = "",
+        extra_col: typing.Optional[str] = None,
```
tswast, line 3639: Please add a docstring explaining this `extra_col` parameter and when it is intended to be used.

bigframes/dataframe.py
```diff
@@ -3562,6 +3630,30 @@ def _perform_join_by_index(
...
+                final_col_names.append(f"{col_name}{lsuffix}")
+            else:
+                final_col_names.append(col_name)
+
+        for col_name in right_columns:
+            if col_name in col_intersection:
+                final_col_names.append(f"{col_name}{rsuffix}")
+            else:
+                final_col_names.append(col_name)
+        self.columns = pandas.Index(final_col_names)
```
tswast, line 3654: I think we should only be modifying `self` if we're doing an `inplace` operation, right? Why is self getting changed? Can we avoid this?

tests/system/small/test_dataframe.py
```diff
@@ -2899,12 +2899,99 @@ def test_join_different_table(
...
+def test_join_different_table_with_duplicate_column_name(
+    scalars_df_index, scalars_pandas_df_index, how
+):
+    bf_df_a = scalars_df_index[["string_col", "int64_col", "int64_too"]].rename(
+        columns={"int64_too": "int64_col"}
+    )
+    bf_df_b = scalars_df_index.dropna()[
+        ["string_col", "int64_col", "int64_too"]
+    ].rename(columns={"int64_too": "int64_col"})
+    bf_result = bf_df_a.join(bf_df_b, how=how, lsuffix="_l", rsuffix="_r").to_pandas()
```
tswast, line 2912: Can we add some checks that `bf_df_a`'s column names and index names didn't get modified?

bigframes/dataframe.py
Genesis929: Added.

bigframes/dataframe.py
Genesis929: We need to avoid duplicates in names when join, or the reordering columns won't work, so for current join logic, we can't avoid this.

bigframes/dataframe.py
Genesis929: We need to avoid duplicates in names when join, or the reordering columns won't work, so for current join logic, we can't avoid this.

tswast (approved):

PR #1864, opened 2025-06-26
deps: replace `google-cloud-iam` with `grpc-google-iam-v1`

shobsi:
`google-cloud-iam` dependency was added for `google.iam.v1.policy_pb2` module for the purpose of updating IAM policy for the BQ connection service account. It turns out that this was unnecessary. What was really needed was `grpc-google-iam-v1` package, which is available via both `bigframes` -> `google-cloud-resource-manager` -> [`grpc-google-iam-v1`](https://github.com/googleapis/google-cloud-python/blob/cdfb4a337fa96b482f4db0e6e4691dc58d0f79c2/packages/google-cloud-resource-manager/setup.py#L51) and `bigframes` -> `google-cloud-iam` -> [`grpc-google-iam-v1`](https://github.com/googleapis/google-cloud-python/blob/b373910f0252de1abe4c060b1ee7af351fdf3063/packages/google-cloud-iam/setup.py#L48) dependencies. Since bigframes is making use of the `policy_pb2` module directly, we are adding `grpc-google-iam-v1` as direct dependency and removing the `google-cloud-iam` dependency in this change.


tswast (changes requested):
Please don't. If we use the IAM module directly (which we do), don't depend on getting it via a transitive dependency.

shobsi:
> Please don't. If we use the IAM module directly (which we do), don't depend on getting it via a transitive dependency.

Added `grpc-google-iam-v1` dependency instead, PTAL, thanks!

tswast (approved):

PR #1874, opened 2025-06-30
refactor: subclass DerefOp as ResolvedDerefOp

chelsea-lin:
Fixes internal issue 427501553 🦕

### DIFF
--- a/bigframes/core/expression.py
+++ b/bigframes/core/expression.py
@@ -432,23 +432,12 @@ class DerefOp(Expression):
 class ResolvedDerefOp(DerefOp):
     """An expression that refers to a column by ID and resolved with schema bound."""
 
-    field: field.Field
+    dtype: dtypes.Dtype
+    nullable: bool = True
 
-    # Re-declare 'id' from the parent to remove it from the __init__ method
-    id: ids.ColumnId = dataclasses.field(init=False)
-
-    def __post_init__(self):
-        # Initialize the parent's 'id' field after the object is created.
-        # We must use object.__setattr__ because the dataclass is frozen.
-        object.__setattr__(self, "id", self.field.id)
-
-    @property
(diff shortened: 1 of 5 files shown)


bigframes/core/rewrite/schema_binding.py
```diff
@@ -52,4 +53,23 @@ def bind_schema_to_node(
 
         return dataclasses.replace(node, by=tuple(bound_bys))
 
+    if isinstance(node, nodes.JoinNode):
+        conditions = tuple(
+            (
+                typing.cast(
+                    ex.ResolvedDerefOp,
+                    ex.bind_schema_fields(left, node.left_child.field_by_id),
+                ),
```
sycai, line 62: Or: `ResolvedDerefOp.from_field(node.left_child.field_by_id[left.id])`? In this case you don't need to cast the type.

Same suggestion for to the right node.

bigframes/core/expression.py
```diff
@@ -429,30 +429,23 @@ def transform_children(self, t: Callable[[Expression], Expression]) -> Expressio
...
 
 @dataclasses.dataclass(frozen=True)
-class SchemaFieldRefExpression(Expression):
-    """An expression representing a schema field. This is essentially a DerefOp with input schema bound."""
+class ResolvedDerefOp(DerefOp):
+    """An expression that refers to a column by ID and resolved with schema bound."""
 
-    field: field.Field
+    dtype: dtypes.Dtype
+    nullable: bool = True
```
sycai, line 436: Could we not provide a default for `nullable`? A resolved deref op should have nullability resolved too.

chelsea-lin: It works!

bigframes/core/rewrite/schema_binding.py
chelsea-lin: Good point.

sycai (approved):

PR #1876, opened 2025-06-30
docs: update gsutil commands to gcloud commands

aliciawilliams:
Change `gsutil` commands to `gcloud storage` in notebook samples. gsutil is the legacy CLI for interacting with Cloud Storage, and we want to encourage migration to `gcloud storage`.


chelsea-lin (approved):
Thanks!

PR #1881, opened 2025-07-01
fea: support multi index for dataframe where

jialuoo:
b/383835018

Fixes #b/383835018 🦕


tests/system/small/test_dataframe.py
```diff
@@ -395,8 +386,102 @@ def test_mask_series_cond(scalars_df_index, scalars_pandas_df_index):
     pandas.testing.assert_frame_equal(bf_result, pd_result)
 
 
-def test_where_series_multi_index(scalars_df_index, scalars_pandas_df_index):
-    # Test when a dataframe has multi-index or multi-columns.
+def test_where_multi_index(scalars_df_index, scalars_pandas_df_index):
```
chelsea-lin, line 389: Maybe moving `test_where_multi_index` to `https://github.com/googleapis/python-bigquery-dataframes/blob/main/tests/system/small/test_multiindex.py` file, and moving `test_where_series_cond` to `test_series.py`?

tests/system/small/test_dataframe.py
```diff
@@ -395,8 +386,102 @@ def test_mask_series_cond(scalars_df_index, scalars_pandas_df_index):
...
+        bf_result,
+        pd_result,
+        check_index_type=False,
+        check_dtype=False,
+    )
+    # Assert the index is still MultiIndex after the operation.
+    assert isinstance(bf_result.index, pd.MultiIndex), "Expected a MultiIndex"
+    assert isinstance(pd_result.index, pd.MultiIndex), "Expected a MultiIndex"
+
+    # Case2: when condition is series and other is dataframe.
```
chelsea-lin, line 438: Write a separate test for each behavior defined by a method? See go/unit-testing-practices?polyglot=java#behavior-testing-dos-and-donts

chelsea-lin (commented):
Can we please include the internal bug id in the PR descriptions to give more context for reviewers. Thanks!

tests/system/small/test_dataframe.py
jialuoo: Yeah, moved `test_where_multi_index ` into `test_multiindex.py`;

`test_where_series_cond` is another test for the DataFrame .where() method. In this case, the condition is a Series, so `_series_cond` is used.

tests/system/small/test_dataframe.py
jialuoo: Done.

chelsea-lin (approved):
Thanks!

PR #1902, opened 2025-07-11
feat: support `date` data type for to_datetime()

sycai:
Fixes #431038295🦕

### DIFF
--- a/bigframes/core/tools/datetimes.py
+++ b/bigframes/core/tools/datetimes.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from collections.abc import Mapping
-from datetime import datetime
+from datetime import date, datetime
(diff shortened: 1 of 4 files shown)


bigframes/core/tools/datetimes.py
```diff
@@ -28,7 +28,7 @@
 
 def to_datetime(
     arg: Union[
-        Union[int, float, str, datetime],
+        Union[int, float, str, datetime, date],
```
chelsea-lin, line 31: The docstring located at `third_party/bigframes_vendored/pandas/core/tools/datetimes.py` should be updated too.

sycai: Good call. Done.

chelsea-lin (approved):

PR #1908, opened 2025-07-15
feat: Support audio_transcribe with partial ordering

shuoweil:
feat: Support audio transcription with partial ordering
This change also fixes a related issue where Block.join would fail on joins with null indexes when operating in this partial ordering mode.

b/430572560


bigframes/core/blocks.py
```diff
@@ -2488,6 +2488,11 @@ def join(
             )
             if result is not None:
                 return result
+
+            # For block identify joins with null indices, perform cross join
```
tswast, line 2492: This doesn't seem desirable. If df1 is n rows and df2 is m rows, won't this end up with n x m rows?

PR #1921, opened 2025-07-18
feat: implement Index.get_loc

shuoweil:
Implement index.get_loc

### DIFF
--- a/bigframes/core/indexes/base.py
+++ b/bigframes/core/indexes/base.py
@@ -329,6 +329,8 @@ class Index(vendored_pandas_index.Index):
         else:
             # Return boolean mask for non-monotonic duplicates
             mask_block = windowed_block.select_columns([match_col_id])
+            # Reset the index to use positional integers instead of original index values
+            mask_block = mask_block.reset_index(drop=True)
             return bigframes.series.Series(mask_block)
 
     def _get_monotonic_slice(
--- a/bigframes/core/indexes/base.py
+++ b/bigframes/core/indexes/base.py
@@ -252,7 +252,9 @@ class Index(vendored_pandas_index.Index):
             self._query_job = query_job
         return self._query_job
 
-    def get_loc(self, key):
+    def get_loc(
+        self, key: typing.Any
+    ) -> typing.Union[int, slice, "bigframes.series.Series"]:
         """Get integer location, slice or boolean mask for requested label.
 
         Args:
--- a/bigframes/core/indexes/base.py
+++ b/bigframes/core/indexes/base.py
@@ -330,8 +330,9 @@ class Index(vendored_pandas_index.Index):
             # Return boolean mask for non-monotonic duplicates
             mask_block = windowed_block.select_columns([match_col_id])
             # Reset the index to use positional integers instead of original index values
-            result_series = bigframes.series.Series(mask_block)
+            mask_block = mask_block.reset_index(drop=True)
             # Ensure correct dtype and name to match pandas behavior
+            result_series = bigframes.series.Series(mask_block)
             return result_series.astype("boolean")
 
     def _get_monotonic_slice(
--- a/bigframes/core/indexes/base.py
+++ b/bigframes/core/indexes/base.py
@@ -256,13 +256,10 @@ class Index(vendored_pandas_index.Index):
         self, key: typing.Any
     ) -> typing.Union[int, slice, "bigframes.series.Series"]:
         """Get integer location, slice or boolean mask for requested label.
-
         Args:
             key: The label to search for in the index.
-
         Returns:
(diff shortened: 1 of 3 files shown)


third_party/bigframes_vendored/pandas/core/indexes/base.py
```diff
@@ -741,6 +741,41 @@ def argmin(self) -> int:
         """
         raise NotImplementedError(constants.ABSTRACT_METHOD_ERROR_MESSAGE)
 
+    def get_loc(self, key):
```
chelsea-lin, line 744: Please add type hints for the key parameter and the return value in both the function signature and the docstring. To ensure the documentation is accurate, please specify the return type as bpd.Series, as it will be a boolean Series rather than a simple array.

bigframes/core/indexes/base.py
```diff
@@ -247,6 +252,95 @@ def query_job(self) -> bigquery.QueryJob:
...
+            ).to_py_scalar()
+            return int(position_scalar)
+
+        # Multiple matches - need to determine if monotonic or not
+        is_monotonic = self.is_monotonic_increasing or self.is_monotonic_decreasing
+        if is_monotonic:
+            # Get min and max positions for slice
+            min_agg = ex.UnaryAggregation(agg_ops.min_op, ex.deref(row_num_col_id.name))
+            max_agg = ex.UnaryAggregation(agg_ops.max_op, ex.deref(row_num_col_id.name))
+            min_result = filtered_block._expr.aggregate([(min_agg, "min_pos")])
```
chelsea-lin, line 330: I am wondering if we call `filtered_block._expr.aggregate([(min_agg, "min_pos"), (max_agg, "max_pos")])`, then we can only need to execute once. This can save the number of queries for better performance.

bigframes/core/indexes/base.py
```diff
@@ -247,6 +252,95 @@ def query_job(self) -> bigquery.QueryJob:
             self._query_job = query_job
         return self._query_job
 
+    def get_loc(self, key):
```
chelsea-lin, line 255: What happens in pandas when the key is `None`, if index has one or two None element?

chelsea-lin (changes requested):

bigframes/core/indexes/base.py
chelsea-lin: @TrevorBergeron do you have any suggestions here?

bigframes/core/indexes/base.py
```diff
@@ -247,6 +252,97 @@ def query_job(self) -> bigquery.QueryJob:
...
+        if self.nlevels != 1:
+            raise NotImplementedError("get_loc only supports single-level indexes")
+
+        # Get the index column from the block
+        index_column = self._block.index_columns[0]
+
+        # Apply row numbering to the original data
+        win_spec = window_spec.unbound()
+        row_num_agg = ex.NullaryAggregation(agg_ops.RowNumberOp())
+        row_num_col_id = ids.ColumnId.unique()
```
sycai, line 280: nit: it looks like these 3 local variables are single-use only. Maybe we can just merge them into the code that creates the window node. 

I find this suggestion helpful: go/totw/161#eliminating-a-local-variable-thats-immediately-returned (even though it's for C++)

bigframes/core/indexes/base.py
```diff
@@ -247,6 +252,97 @@ def query_job(self) -> bigquery.QueryJob:
...
+            min_agg = ex.UnaryAggregation(agg_ops.min_op, ex.deref(row_num_col_id.name))
+            position_result = filtered_block._expr.aggregate([(min_agg, "position")])
+            position_scalar = self._block.session._executor.execute(
+                position_result
+            ).to_py_scalar()
+            return int(position_scalar)
+
+        # Multiple matches - need to determine if monotonic or not
+        is_monotonic = self.is_monotonic_increasing or self.is_monotonic_decreasing
+        if is_monotonic:
```
sycai, line 328: The body of this if branch is long. Shall we create a helper method instead for better readability?

bigframes/core/indexes/base.py
shuoweil: Thanks for the suggestion, I use temporary saving instead

bigframes/core/indexes/base.py
shuoweil: For sure, I create a new helper function for better readibility

bigframes/core/indexes/base.py
shuoweil: I have updated it and use one query

sycai (dismissed):

chelsea-lin (approved):

PR #1939, opened 2025-07-25
chore: use `bigframes.pandas._read_gbq_colab()` in benchmarks

tswast:
…pandas`.

I updated the benchmarks in `tests/benchmark/read_gbq_colab` to call `bigframes.pandas._read_gbq_colab(...)` instead of `session._read_gbq_colab(...)`.

Since the `session` object is no longer needed in these benchmarks, I also removed that dependency.


tests/benchmark/read_gbq_colab/aggregate_output.py
```diff
@@ -56,9 +54,8 @@ def aggregate_output(
         project_id,
         dataset_id,
         table_id,
-        session,
         suffix,
-    ) = utils.get_configuration(include_table_id=True)
+    ) = utils.get_configuration(include_table_id=True, include_session=False)
```
tswast, line 58: This parameter doesn't actually exist. Might be nice to have a `start_session=False` argument, though.

PR #1942, opened 2025-07-25
chore: use `bigframes.pandas._read_gbq_colab()` in benchmarks

tswast:

### DIFF
--- a/tests/benchmark/read_gbq_colab/dry_run.py
+++ b/tests/benchmark/read_gbq_colab/dry_run.py
@@ -15,20 +15,20 @@ import pathlib
 
 import benchmark.utils as utils
 
-import bigframes.session
+import bigframes.pandas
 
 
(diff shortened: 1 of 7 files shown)


tests/benchmark/read_gbq_colab/dry_run.py
```diff
@@ -38,5 +37,4 @@ def dry_run(*, project_id, dataset_id, table_id, session: bigframes.session.Sess
         project_id=config.project_id,
         dataset_id=config.dataset_id,
         table_id=config.table_id,
-        session=config.session,
```
tswast, line 41: probably need to pass the global session for these. But we don't want to fetch the global session too early. 🤔  Maybe need to update `get_execution_time` to use the global session if none was supplied.

TrevorBergeron (approved):

jialuoo (approved):

PR #1956, opened 2025-07-30
fix: add warnings for duplicated or conflicting type hints in bigfram…

jialuoo:

Fixes b/409565963.


jialuoo:
> Maybe we should just warn if the decorator actually conflicts with the signature?

I considered that. While warning only on a signature conflict is an option, I'm concerned about the risk of users defining type hints in both the decorator and the function signature. Since this is just a warning and won't fail the execution, I lean toward warning users about any duplicated type hints too. What do you think?

jialuoo:
> > Maybe we should just warn if the decorator actually conflicts with the signature?
> 
> I considered that. While warning only on a signature conflict is an option, I'm concerned about the risk of users defining type hints in both the decorator and the function signature. Since this is just a warning and won't fail the execution, I lean toward warning users about any duplicated type hints too. What do you think?

Per off-line discussion, we'll warn only the conflict.

TrevorBergeron (approved):

PR #1968, opened 2025-08-06
test: Mock Claude models to improve LLM test reliability

shuoweil:
Mock Claude models to improve LLM test reliability
b/436340035


shuoweil:
> us-east5 region,

It is true, I can use a session in us-east5 region, it also limits our testing. I have talked to Jasper before I work on this PR, we can use claude in us-east5. Let's chat offline tomorrow.

PR #2020, opened 2025-08-22
feat: Support callable for dataframe mask method

jialuoo:

Fixes b/440085224 🦕

### DIFF
--- a/bigframes/dataframe.py
+++ b/bigframes/dataframe.py
@@ -2787,6 +2787,19 @@ class DataFrame(vendored_pandas_frame.DataFrame):
             for item in df.itertuples(index=index, name=name):
                 yield item
 
+    def _apply_callable(self, condition):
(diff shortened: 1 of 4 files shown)


bigframes/dataframe.py
```diff
@@ -2899,7 +2904,7 @@ def where(self, cond, other=None):
         return result
 
     def mask(self, cond, other=None):
-        return self.where(~cond, other=other)
+        return self.where(~self._apply_callable(cond), other=other)
```
TrevorBergeron, line 2907: I don' think this change is needed, as `self.where` will do `_apply_callable`

jialuoo: Without this change, the callable `cond` will do `~` operation before doing `_apply_callable` from `self.where`, which will error out.

TrevorBergeron (approved):

PR #2054, opened 2025-09-05
feat: Add str.join method

TrevorBergeron:

### DIFF
--- a/bigframes/core/compile/ibis_compiler/scalar_op_registry.py
+++ b/bigframes/core/compile/ibis_compiler/scalar_op_registry.py
@@ -1220,6 +1220,10 @@ def array_reduce_op_impl(x: ibis_types.Value, op: ops.ArrayReduceOp):
         lambda arr_vals: agg_compilers.compile_unary_agg(
             op.aggregation, typing.cast(ibis_types.Column, arr_vals)
         )
+        if op.aggregation.order_independent
+        else agg_compilers.compile_ordered_unary_agg(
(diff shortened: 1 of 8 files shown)


bigframes/core/compile/ibis_compiler/scalar_op_registry.py
```diff
@@ -1220,6 +1220,10 @@ def array_reduce_op_impl(x: ibis_types.Value, op: ops.ArrayReduceOp):
         lambda arr_vals: agg_compilers.compile_unary_agg(
             op.aggregation, typing.cast(ibis_types.Column, arr_vals)
         )
+        if op.aggregation.order_independent
+        else agg_compilers.compile_ordered_unary_agg(
+            op.aggregation, typing.cast(ibis_types.Column, arr_vals)
+        )
```
sycai, line 1226: Let's just use a if-else block here:

```
casted_x = typing.cast(ibis_types.ArrayValue, x)
if op.aggregation.order_independent:
  return casted_x.reduce(...)
else:
  return casted_x.reduce(...)
```

sycai (commented):
Thanks! 

Just a heads-up for @chelsea-lin , we need to have this operator in the SQLGlot compiler too.

bigframes/core/compile/ibis_compiler/scalar_op_registry.py
TrevorBergeron: fixed

sycai (approved):

PR #2067, opened 2025-09-10
feat: support to cast struct to json

chelsea-lin:
Fixes internal issue 444196993 🦕


jialuoo (approved):

TrevorBergeron (changes requested):
please extend the astype engine test if we are extending cast operator.
I'd almost rather use a separate operator instead of the cast operator, as the casting operator is incredibly laden with logic

chelsea-lin:
#2070 to follow up the astype engine test

PR #2081, opened 2025-09-13
refactor: Define window column expression type

TrevorBergeron:

### DIFF
--- a/bigframes/core/compile/compiled.py
+++ b/bigframes/core/compile/compiled.py
@@ -166,18 +166,6 @@ class UnorderedIR:
             bigframes.core.compile.ibis_types.ibis_dtype_to_bigframes_dtype(ibis_type),
         )
 
-    def row_count(self, name: str) -> UnorderedIR:
-        original_table = self._to_ibis_expr()
-        ibis_table = original_table.agg(
-            [
-                original_table.count().name(name),
-            ]
-        )
-        return UnorderedIR(
-            ibis_table,
-            (ibis_table[name],),
-        )
-
     def _to_ibis_expr(
         self,
         *,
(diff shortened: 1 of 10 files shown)


bigframes/core/compile/ibis_compiler/scalar_op_compiler.py
```diff
@@ -20,16 +20,18 @@
...
+from bigframes.core import agg_expressions, ordering
 import bigframes.core.compile.ibis_types
 import bigframes.core.expression as ex
 
 if TYPE_CHECKING:
     import bigframes.operations as ops
 
 
-class ScalarOpCompiler:
+class ExpressionCompiler:
```
chelsea-lin, line 34: Do we want to apply the same renaming to the SQLGlot: https://github.com/googleapis/python-bigquery-dataframes/blob/main/bigframes/core/compile/sqlglot/scalar_compiler.py

TrevorBergeron: If it starts compiling analytic expressions then yeah. For now, ScalarOpCompiler is fine, as it only compiles scalar ops right now.

bigframes/core/compile/compiled.py
```diff
@@ -442,113 +431,64 @@ def project_window_op(
...
                 is_observation = functools.reduce(
-                    lambda x, y: x & y, per_col_does_count
-                ).cast(int)
-                observation_count = agg_compiler.compile_analytic(
-                    ex_types.UnaryAggregation(
-                        agg_ops.sum_op, ex.deref("_observation_count")
-                    ),
-                    window,
-                    bindings={"_observation_count": is_observation},
+                    lambda x, y: ops.and_op.as_expr(x, y), per_col_does_count
```
chelsea-lin, line 453: Instead compiling to Ibis nodes directly, it rewrites the BF expressions at middle of the compiler. The SQLGlot compiler follows the old pattern to write the SQLGlot expressions. What benefits to have the additional conversions to BF expressions here?

chelsea-lin (approved):

PR #2087, opened 2025-09-15
feat: add `**kwargs` support to `XGBClassifier` and `XGBRegressor`

tswast:
This is in alignment with https://xgboost.readthedocs.io/en/stable/python/python_api.html\#xgboost.XGBRegressor if considering BQML to be a booster type.

Fixes internal customer request 🦕

### DIFF
--- a/bigframes/ml/ensemble.py
+++ b/bigframes/ml/ensemble.py
@@ -17,7 +17,7 @@ https://scikit-learn.org/stable/modules/ensemble.html"""
 
 from __future__ import annotations
 
-from typing import Dict, List, Literal, Optional
+from typing import Dict, List, Literal, Optional, Union
 
 import bigframes_vendored.sklearn.ensemble._forest
 import bigframes_vendored.xgboost.sklearn
@@ -78,6 +78,7 @@ class XGBRegressor(
         tol: float = 0.01,
         enable_global_explain: bool = False,
         xgboost_version: Literal["0.9", "1.1"] = "0.9",
+        **kwargs: Union[str, str | int | bool | float | List[str]],
     ):
         self.n_estimators = n_estimators
         self.booster = booster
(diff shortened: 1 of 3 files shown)


bigframes/ml/ensemble.py
```diff
@@ -139,6 +141,8 @@ def _bqml_options(self) -> Dict[str, str | int | bool | float | List[str]]:
             "enable_global_explain": self.enable_global_explain,
             "xgboost_version": self.xgboost_version,
         }
+        options.update(self._extra_bqml_options)
```
GarrettWu, line 144: It may override existing options, if user pass in both.

We use sklearn terms instead of BQML terms for existing parameters, e.g. learning_rate instead of LEARN_RATE. User may get confused looking at two sets of options at the same time.

It becomes hard for us to both stick with sklearn-like experience and support BQML other offerings. 

Or maybe we can ask them to only use one set of parameters, and raise error of conflicts?

tests/unit/ml/test_golden_sql.py
```diff
@@ -286,3 +287,83 @@ def test_decomposition_mf_score_with_x(mock_session, bqml_model, mock_X):
...
+              max_tree_depth=6,
+              subsample=1.0,
+              l1_reg=0.0,
+              l2_reg=1.0,
+              learn_rate=0.3,
+              max_iterations=20,
+              min_rel_progress=0.01,
+              enable_global_explain=False,
+              xgboost_version='0.9',
+              category_encoding_method='LABEL_ENCODING',
```
GarrettWu, line 324: Do they want to use label encoding on features? This is actually wrong. https://stackoverflow.com/a/34346937

I talked with jiashangliu@, but BQML won't change the default.

tswast:
Closing. I think the alternative of exposing the SQL methods directly in the `bigframes.bigquery.ml` module, similar to `bigframes.bigquery.ai` will be more useful.

PR #2095, opened 2025-09-17
refactor: reorganize the sqlglot scalar compiler layout - part 3

chelsea-lin:
This change follows up on #2075 by splitting binary_compiler.py and its unit test file into smaller files.


tswast (approved):
THanks!

PR #2103, opened 2025-09-18
chore: implement ai.generate_bool in SQLGlot compiler

sycai:
Fixes #443853093 🦕


bigframes/core/compile/sqlglot/expressions/ai_ops.py
```diff
@@ -0,0 +1,63 @@
...
+        sge.Kwarg(
+            this="request_type", expression=sge.Literal.string(op.request_type.upper())
+        )
+    )
+
+    if op.model_params is not None:
+        args.append(
+            sge.Kwarg(
+                this="model_params",
+                expression=sge.JSON(this=sge.Literal.string(op.model_params)),
```
chelsea-lin, line 59: Can you please leave comment for `sge.JSON`, that requires a newer version of SQLGlot. And the `Parse_JSON` does not work well because a JSON literal is required here. https://screenshot.googleplex.com/9GTcjSYhrjYdZ5k

sycai: Done.

chelsea-lin (approved):

PR #2112, opened 2025-09-24
fix: Prevent invalid syntax for no-op .replace ops

TrevorBergeron:


bigframes/core/compile/ibis_compiler/scalar_op_registry.py
```diff
@@ -1173,6 +1173,10 @@ def udf(*inputs):
 
 @scalar_op_compiler.register_unary_op(ops.MapOp, pass_op=True)
 def map_op_impl(x: ibis_types.Value, op: ops.MapOp):
```
chelsea-lin, line 1175: Could you please check if the SQLGlot has the same issue: bigframes/core/compile/sqlglot/expressions/generic_ops.py

TrevorBergeron: yeah, probably?

chelsea-lin (approved):

PR #2119, opened 2025-09-25
fix: show progress even in job optional queries

tswast:

Fixes b/409390651
Towards b/409104302
🦕


bigframes/core/events.py
```diff
@@ -0,0 +1,210 @@
...
+import google.cloud.bigquery.job.query
+import google.cloud.bigquery.table
+
+import bigframes.formatting_helpers
+import bigframes.session.executor
+
+
+@dataclasses.dataclass(frozen=True)
+class Subscriber:
+    callback_ref: weakref.ref
```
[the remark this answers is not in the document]
tswast: > What is the scenario we are imagining?

The anywidget table widgets. When they re-run the cell, the TableWidget we create is no longer needed and should go out of scope, but as far as I know, we don't really have an opportunity to unsubscribe at that time.

bigframes/core/events.py
```diff
@@ -0,0 +1,210 @@
...
+    # TODO(tswast): Add block_id to allow filter in context managers.
+
+
+class Publisher:
+    def __init__(self):
+        self._subscribers: List[Subscriber] = []
+        self._subscribers_lock = threading.Lock()
+
+    def subscribe(self, callback):
```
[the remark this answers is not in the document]
TrevorBergeron: Yeah, I guess as long as this is internal-use only, we can tack on features like this later, and refactor callers as needed.

tswast:
e2e failure: `FAILED tests/system/large/blob/test_function.py::test_blob_transcribe[gemini-2.0-flash-lite-001-False]` appears to be unrelated

bigframes/core/events.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def subscribe(self, callback: Callable[[Event], None]) -> Subscriber:
+        # TODO(b/448176657): figure out how to handle subscribers/publishers in
+        # a background thread. Maybe subscribers should be thread-local?
+        subscriber = Subscriber(callback, publisher=self)
+        self._subscribers.add(subscriber)
+        return subscriber
+
+    def unsubscribe(self, subscriber: Subscriber):
+        self._subscribers.remove(subscriber)
```
TrevorBergeron, line 78: Might cause errors if done concurrently with iteration?

bigframes/core/events.py
```diff
@@ -0,0 +1,232 @@
...
+import google.cloud.bigquery.table
+
+import bigframes.session.executor
+
+
+class Subscriber:
+    def __init__(self, callback: Callable[[Event], None], *, publisher: Publisher):
+        self._publisher = publisher
+        self._callback = callback
+        self._subscriber_id = str(uuid.uuid4())
```
TrevorBergeron, line 33: do we even need to convert to string?

bigframes/core/events.py
```diff
@@ -0,0 +1,232 @@
...
+    def __hash__(self) -> int:
+        return hash(self._subscriber_id)
+
+    def __eq__(self, value: object):
+        if not isinstance(value, Subscriber):
+            return NotImplemented
+        return value._subscriber_id == self._subscriber_id
+
+    def close(self):
+        self._publisher.unsubscribe(self)
```
TrevorBergeron, line 47: Another alternative to explicitly removing from subscriber list in a blocking way is to just flag oneself as closed, and the publisher can then remove at its convenience. I think this approach works fine though

TrevorBergeron (approved):

PR #2123, opened 2025-09-26
fix!: Address the series input type issue in bigframes functions

jialuoo:
The primary issue is the type recognition issue when a UDF uses a Series type hint, since the underlying serialization process converts the input series into a string.

Solutions implemented:
1) Managed Functions: Solved by adding `from __future__ import annotations`, which enables the postponed evaluation of type annotations, preventing an eager runtime type error.

2) Remote Functions: Solved by enforcing pandas.Series as the required input type hint. This is because the cloudpickle used in remote function can recognize pandas series since the later serialization is between string and pandas series, not bigframes series.

In general, this PR improves the robustness of bigframes functions and prevents a potential type error during serialization.

Fixes b/445182819 🦕

BEGIN_COMMIT_OVERRIDE
BREAKING CHANGE: argument type change.
Release-As: 2.25.0
END_COMMIT_OVERRIDE


bigframes/functions/_function_session.py
```diff
@@ -983,7 +983,18 @@ def _convert_row_processor_sig(
...
+        # Type hints for Series inputs should use pandas.Series because the
+        # underlying serialization process converts the input to a string
+        # representation of a pandas Series (not bigframes Series). Using
+        # bigframes Series will lead to TypeError when creating the function
+        # remotely. See more from b/445182819.
+        if param_type == bf_series.Series:
+            raise bf_formatting.create_exception_with_feedback_link(
+                TypeError,
+                "Due to current UDF serialization requirements, argument type "
+                "hint must be Pandas Series, not BigFrames Series.",
```
TrevorBergeron, line 995: I don't think we should try to explain why? eg "udf serialization requirements"

jialuoo: Good point. Fixed.

TrevorBergeron (approved):

PR #2159, opened 2025-10-09
refactor: ExecuteResult is reusable, sampleable

TrevorBergeron:


bigframes/core/blocks.py
```diff
@@ -724,19 +720,21 @@ def to_pandas_batches(
...
-                execute_result.to_pandas_batches(page_size, max_results),
+                result_batches.to_pandas_batches(page_size, max_results),
                 [0],
                 fillvalue=empty_val,
             ),
         )
         dfs = iter(map(self._copy_index_to_pandas, dfs))
 
-        total_rows = execute_result.total_rows
+        total_rows = result_batches.approx_total_rows
```
tswast: As we discussed offline, I think it'd be helpful to have cases where we know (or at least have very high confidence) of this being exact, such as when reading the destination table of a query.

tswast (commented):
At a high level, I think this makes a lot of sense. My main worry would be the case of ExecuteResult with local data. If we are keeping these around more, is there a risk of memory leaks?

bigframes/core/blocks.py
TrevorBergeron: We do insert row count directly in the tree when known exactly (or at least some of the time we know exactly, might be missing some opportunities). At a certain point we just intermingle known and approx though, before surfacing. Could keep separate if really needed, but a bit more complexity.

bigframes/core/compile/ibis_compiler/ibis_compiler.py
```diff
@@ -186,7 +186,7 @@ def compile_readtable(node: nodes.ReadTableNode, *args):
     # TODO(b/395912450): Remove workaround solution once b/374784249 got resolved.
     for scan_item in node.scan_list.items:
         if (
-            scan_item.dtype == dtypes.JSON_DTYPE
+            node.source.schema.get_type(scan_item.source_id) == dtypes.JSON_DTYPE
```
tswast: I'm curious what the purpose of this extra layer of indirection is compared to the previous? I guess it's partially undoing some of the changes Shenyang and Chelsea made to plumb type info into the tree?

tswast (approved):

tswast:
e2e failures are "Blob" related. Probably flakes.

PR #2185, opened 2025-10-21
feat: include local data bytes in the dry run report when available

sycai:
All local PyArrow tables referenced in the expression tree contribute to the "totalBytesProcessed" dry run statistics. 

Data that are directly embeded in the SQL (e.g. `bpd.read_gbq("SELECT * FROM UNNEST([1,2,3])")`) are not counted.

Fixes #437707810 🦕


tswast (approved):
Woohoo, thanks!

PR #2222, opened 2025-11-03
fix: Correct connection normalization in blob system tests

shuoweil:
This PR addresses a bug in the blob storage system tests where the normalize_connection_id fixture was incorrectly implemented, causing several tests to fail.

  The key changes include:
   - Correcting the normalize_connection_id fixture to return a function that normalizes the connection ID.
   - Updating type hints in test_io.py and test_properties.py to use typing.Callable.
   - Applying the normalization function consistently across all relevant blob tests.
   - Skipping flaky tests related to blob properties and multimodal LLMs.

Fixes #<457416070> 🦕

### DIFF
--- a/tests/system/conftest.py
+++ b/tests/system/conftest.py
@@ -71,6 +71,23 @@ def _hash_digest_file(hasher, filepath):
 
 
 @pytest.fixture(scope="session")
+def normalize_connection_id():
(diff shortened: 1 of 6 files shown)


chelsea-lin (changes requested):
The commit should be named starting with the "fix: " prefix.

chelsea-lin (approved):

PR #2244, opened 2025-11-07
refactor: make sqlglot compile_sql a top-level function

chelsea-lin:
This change prepares to enable SQLGlot as the default compiler. Making compile_sql a top-level function makes it consistent with the existing Ibis API, which simplifies onboarding.


sycai (approved):
Nice! Thank you!

PR #2246, opened 2025-11-10
chore: make labels tests more robust to test order

tswast:

Fixes internal issue b/459432106 🦕

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,6 +2,9 @@ on:
   pull_request:
     branches:
       - main
+  push:
+    branches:
+      - main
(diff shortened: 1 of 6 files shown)


.github/workflows/docs.yml
```diff
@@ -2,6 +2,9 @@ on:
   pull_request:
     branches:
       - main
+  push:
```
tswast: This will make it easier to look through https://github.com/googleapis/python-bigquery-dataframes/actions?query=branch%3Amain+event%3Apush to see when failures started.

TrevorBergeron (approved):

PR #2250, opened 2025-11-10
feat: Refactor TableWidget and to_pandas_batches

shuoweil:
This PR refactors DataFrame.to_pandas_batches to return a blocks.PandasBatches object, improving the handling of streamed data and total row counts.

  Additionally, this PR updates TableWidget to:
   * Utilize the new PandasBatches return type for more robust data loading.
   * Improve thread safety during HTML updates using a re-entrant lock (RLock), ensuring concurrent
     interactions (like rapid sorting or pagination) are handled correctly without race conditions.
   * Fix compatibility with older Python versions by updating type annotations.
   * Refactor index handling logic within the widget for better clarity and stability.

  User Impact:
   * More Stable Interactive Tables: Users experiencing the interactive table widget in notebooks
     should see more reliable updates when paging or sorting rapidly.
   * Internal API Consistency: to_pandas_batches now provides a more structured return type for
     downstream consumers.

Fixes #<459515995> 🦕

### DIFF
--- a/bigframes/display/anywidget.py
+++ b/bigframes/display/anywidget.py
@@ -99,9 +99,8 @@ class TableWidget(_WIDGET_BASE):
         self._batch_iter: Optional[Iterator[pd.DataFrame]] = None
         self._cached_batches: list[pd.DataFrame] = []
         self._last_sort_state: Optional[_SortState] = None
-        # RLock is needed because _set_table_html can be re-entrant when
-        # self.page is updated within the method, triggering the observer.
-        self._setting_html_lock = threading.RLock()
+        # Lock to ensure only one thread at a time is updating the table HTML.
+        self._setting_html_lock = threading.Lock()
 
         # respect display options for initial page size
         initial_page_size = bigframes.options.display.max_rows
@@ -270,6 +269,7 @@ class TableWidget(_WIDGET_BASE):
 
     def _set_table_html(self) -> None:
         """Sets the current html data based on the current page and page size."""
+        new_page = None
         with self._setting_html_lock:
             if self._error_message:
                 self.table_html = (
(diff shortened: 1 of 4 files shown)


bigframes/display/html.py
```diff
@@ -29,40 +29,137 @@ def _is_dtype_numeric(dtype) -> bool:
...
+
+        # Finalize the last span
+        if current_span > 0:
+            current_level_spans[len(dataframe.index) - current_span] = current_span
+
+        rowspans.append(current_level_spans)
+
+    return rowspans
+
+
```
tswast: While this makes for a somewhat cleaner view, I think we'd be better off not implementing rowpans. For example sorting by the various index columns is less intuitive if the index isn't rendered the same as other columns.

bigframes/display/html.py
```diff
@@ -29,40 +29,137 @@ def _is_dtype_numeric(dtype) -> bool:
...
     classes = "dataframe table table-striped table-hover"
     table_html = [f'<table border="1" class="{classes}" id="{table_id}">']
     precision = options.display.precision
 
     # Render table head
     table_html.append("  <thead>")
     table_html.append('    <tr style="text-align: left;">')
+
+    # Add index headers
+    for name in dataframe.index.names:
```
tswast: Let's try to actually refer to the original BigFrames DataFrame when deciding which index columns to render or not. In particular, we should not render the index if the BigFrame DataFrame has a NULL index.

bigframes/display/html.py
shuoweil: A new implementation is used without rowspans.

bigframes/display/html.py
shuoweil: My new version correctly refers to the original BigFrames DataFrame's _block.has_index property. If has_index is False (indicating a "NULL index"), the _prepare_dataframe_for_display method ensures that a default "Row" column is inserted for display instead of attempting to render a non-existent index. My recent refactoring preserved this established logic.

bigframes/display/anywidget.py
```diff
@@ -65,17 +65,13 @@ class TableWidget(WIDGET_BASE):
...
-        allow_none=True,
-    ).tag(sync=True)
+    row_count = traitlets.Int(allow_none=True, default_value=None).tag(sync=True)
     table_html = traitlets.Unicode().tag(sync=True)
     sort_column = traitlets.Unicode("").tag(sync=True)
     sort_ascending = traitlets.Bool(True).tag(sync=True)
     orderable_columns = traitlets.List(traitlets.Unicode(), []).tag(sync=True)
     _initial_load_complete = traitlets.Bool(False).tag(sync=True)
-    _batches: Optional[blocks.PandasBatches] = None
+    _batches: blocks.PandasBatches | None = None
```
tswast: This syntax was added in Python 3.10. https://peps.python.org/pep-0604/

Currently, we are supporting Python 3.9 https://github.com/googleapis/python-bigquery-dataframes/blob/3c54c68fd86bd85784f09e7ff819d39ca6e8409b/setup.py#L155 but since this version is end-of-life, we should probably drop it sooner rather than later.

Please wait to use this syntax until we drop Python 3.9. I have filed b/470438395 to track this.

bigframes/display/anywidget.py
```diff
@@ -255,95 +260,123 @@ def _reset_batch_cache(self) -> None:
...
         self._reset_batch_cache()
 
     def _set_table_html(self) -> None:
         """Sets the current html data based on the current page and page size."""
-        if self._error_message:
-            self.table_html = (
-                f"<div class='bigframes-error-message'>{self._error_message}</div>"
-            )
+        if hasattr(self, "_setting_html"):
             return
```
tswast: I'd prefer we initialize this to false in __init__ instead of relying on hasattr.

tswast: Also, consider multi threading. We probably need a lock to fully prevent two table htmls being set at the same time.

bigframes/display/anywidget.py
```diff
@@ -255,95 +260,123 @@ def _reset_batch_cache(self) -> None:
...
+        self._setting_html = True
 
-        # Apply sorting if a column is selected
-        df_to_display = self._dataframe
-        if self.sort_column:
-            # TODO(b/463715504): Support sorting by index columns.
-            df_to_display = df_to_display.sort_values(
-                by=self.sort_column, ascending=self.sort_ascending
+        try:
```
tswast: I would prefer a context manager instead of a try/except/finally block. That way we can be more explicit about what the intention is. Are you sure you want to competely skip setting the html if another thread tries to set it? If not, just do a with block with a lock here. If so, please explain why it's safe to skip.

Aside: I would actually expect the opposite here, where later set calls would interrupt any set call that is currently running.

tswast (changes requested):

bigframes/display/anywidget.py
shuoweil: Here is what I change to:  (1) self._setting_html = False and self._setting_html_lock = threading.Lock() were added to the TableWidget.__init__ method 
(2) The _set_table_html method was refactored to use with self._setting_html_lock: and the self._setting_html flag to prevent race conditions during HTML rendering. The flag is reset in the finally block to ensure it's always released.

bigframes/display/anywidget.py
```diff
@@ -255,95 +260,123 @@ def _reset_batch_cache(self) -> None:
...
+                dataframe=page_data,
+                table_id=f"table-{self._table_id}",
+            )
+        finally:
+            delattr(self, "_setting_html")
+
+    def _get_page_data(self) -> pd.DataFrame:
+        """Get the data for the current page, handling unknown row count."""
+        # This loop is to handle auto-correction of page number when row count is unknown
+        while True:
```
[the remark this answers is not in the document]
shuoweil: No, it is not related to multiindex display. I mix so many features into the same pr. I have refactor them out. And move this one to a separate new branch.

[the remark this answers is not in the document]
shuoweil: Apologies for the confusion. I have updated the scope of this PR to focus strictly on refactoring DataFrame.to_pandas_batches and updating TableWidget to use it.

  I have removed the unsafe while True retry loop and the associated helper methods to simplify the implementation and address your safety concerns. The logic now uses the standard pattern of updating self.page to handle unknown row counts (triggering the observer), rather than attempting to resolve it inside a loop.

bigframes/display/anywidget.py
```diff
@@ -99,15 +97,17 @@ def __init__(self, dataframe: bigframes.dataframe.DataFrame):
         self._table_id = str(uuid.uuid4())
         self._all_data_loaded = False
         self._batch_iter: Optional[Iterator[pd.DataFrame]] = None
-        self._cached_batches: List[pd.DataFrame] = []
+        self._cached_batches: list[pd.DataFrame] = []
         self._last_sort_state: Optional[_SortState] = None
+        self._setting_html_lock = threading.RLock()
```
tswast: Why use an RLock instead of a lock? Seems to me you only want one thread at a time running setting_html, not multiple.


```suggestion
        self._setting_html_lock = threading.Lock()
```

shuoweil: RLock is needed because _set_table_html can be re-entrant when self.page is updated within the method, triggering the observer. I also added this information as comment

tswast (dismissed):

tswast (dismissed):

tswast (approved):

PR #2255, opened 2025-11-11
feat: Implement single-column sorting for interactive table widget

shuoweil:
This PR introduces single-column sorting functionality to the interactive table widget.

1) **Three-State Sorting UI**

1.1) The sort indicator dot (●) is now hidden by default and only appears when the user hovers the mouse over a column header
1.2) Implemented a sorting cycle: unsorted (●) → ascending (▲) → descending (▼) → unsorted (●).
1.3) Visual indicators (●, ▲, ▼) are displayed in column headers to reflect the current sort state.
1.4) Sorting controls are now only enabled for columns with orderable data types.

2) **Tests**
2.1) Updated `paginated_pandas_df` fixture for better sorting test coverage
2.2) Added new system tests to verify ascending, descending, and multi-column sorting.

**3. Frontend Unit Tests**
JavaScript-level unit tests have been added to validate the widget's frontend logic, specifically the new sorting functionality and UI interactions.

**How to Run Frontend Unit Tests**:
To execute these tests from the project root directory:
```bash
cd tests/js
npm install  # Only needed if dependencies haven't been installed or have changed
npm test
```

Docs has been updated to document the new features. The main description now mentions column sorting and adjustable widths, and a new section has been added to explain how to use the column resizing feature. The sorting section was also updated to mention that the indicators are only visible on hover.

Fixes #<459835971> 🦕


bigframes/display/table_widget.js
```diff
@@ -124,6 +126,36 @@ function render({ model, el }) {
 		// Note: Using innerHTML is safe here because the content is generated
 		// by a trusted backend (DataFrame.to_html).
 		tableContainer.innerHTML = model.get(ModelProperty.TABLE_HTML);
+
+		// Add click handlers to column headers for sorting
+		const headers = tableContainer.querySelectorAll("th");
+		headers.forEach((header) => {
+			const columnName = header.textContent.trim();
+			if (columnName) {
```
tswast: Not all data types are sortable. See the orderable property in our dtypes. https://github.com/googleapis/python-bigquery-dataframes/blob/9f497a6ddd52c8aa41141c8641b6d189bfb15210/bigframes/dtypes.py#L137

We should not add the handler or arrow to columns that we can't sort.

shuoweil: Here is the plan:
1. Display a dot (●) indicator for all sortable columns by default
2. Allow users to cycle through three states: unsorted (●) → ascending (▲) → descending (▼) → unsorted (●)
3. Only show sorting UI for columns with orderable data types

bigframes/display/anywidget.py
```diff
@@ -61,6 +62,9 @@ class TableWidget(WIDGET_BASE):
         allow_none=True,
     ).tag(sync=True)
     table_html = traitlets.Unicode().tag(sync=True)
+    sort_column = traitlets.Unicode("").tag(sync=True)
```
tswast: Have we considered having multiple columns as a possibility? I think a single column is a good starting point, but I think it's an alternative worth considering, especially when a particular column contains lots of duplicate values, like a "date" column.

bigframes/display/anywidget.py
```diff
@@ -87,15 +91,20 @@ def __init__(self, dataframe: bigframes.dataframe.DataFrame):
         self._all_data_loaded = False
         self._batch_iter: Optional[Iterator[pd.DataFrame]] = None
         self._cached_batches: List[pd.DataFrame] = []
+        self._last_sort_state: Optional[Tuple[str, bool]] = None
```
tswast: I can guess based on context what these mean, but I think a frozen dataclass would be much easier to understand at a glance.

bigframes/display/anywidget.py
```diff
@@ -87,15 +91,20 @@ def __init__(self, dataframe: bigframes.dataframe.DataFrame):
...
 
         # respect display options for initial page size
         initial_page_size = bigframes.options.display.max_rows
 
         # set traitlets properties that trigger observers
         self.page_size = initial_page_size
+        self.orderable_columns = [
+            col
+            for col in dataframe.columns
```
tswast: We should iterate through dtypes directly to get both the type and the column name in one call.

Calling `.dtypes` for each iteration is making this into an O(n^2) algorithm. The n could be as large as 10,000, where that would definitely make a difference, though I imagine our widget would break well before then.

Aside: could you file an issue to check our limitations on the number of columns from a ux perspective? I imagine we'll need to solve that at some point soon-ish.

bigframes/display/anywidget.py
```diff
@@ -235,6 +245,27 @@ def _set_table_html(self) -> None:
...
             return
 
+        # Apply sorting if a column is selected
+        df_to_display = self._dataframe
+        if self.sort_column:
+            try:
+                df_to_display = df_to_display.sort_values(
+                    by=self.sort_column, ascending=self.sort_ascending
+                )
+            except KeyError:
```
tswast: I'm afraid of this. KeyError is a relatively common error in Python. Just looking at this code, it's not clear to me that this would always be the case of a missing column. Please check for the column presence explicitly, instead.

shuoweil: It should trigger this error to be honest. The user only allows the click on a sortable column name to trigger this part of code. Theoretically, we do not even need to catch this bug, because this exception should never be hit in our design. I will remove this try catch block.

bigframes/display/table_widget.js
```diff
@@ -126,14 +140,93 @@ function render({ model, el }) {
...
 		tableContainer.innerHTML = model.get(ModelProperty.TABLE_HTML);
+
+		// Get sortable columns from backend
+		const sortableColumns = model.get(ModelProperty.ORDERABLE_COLUMNS);
+		const currentSortColumn = model.get(ModelProperty.SORT_COLUMN);
+		const currentSortAscending = model.get(ModelProperty.SORT_ASCENDING);
+
+		// Add click handlers to column headers for sorting
+		const headers = tableContainer.querySelectorAll("th");
+		headers.forEach((header) => {
```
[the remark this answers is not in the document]
shuoweil: I've set up a comprehensive JavaScript unit testing environment and implemented tests for the `TableWidget`'s interactive sorting functionality.

**Testing Framework & Environment**:
    *   [**Jest**:](https://g3doc.corp.google.com/company/teams/appsheet/engDocs/g3doc/testing/frontend-unit-tests.md?cl=head) Chosen as the primary testing framework due to its all-in-one nature (runner, assertion library, mocking) and strong community support.
    *   **JSDOM**: Utilized within Jest to simulate a browser environment, enabling DOM manipulation and event handling within Node.js.
    *   **Babel**: Configured to transpile modern JavaScript (ES modules) used in the widget and tests, ensuring compatibility.

**Test Coverage**: The `table_widget.test.js` file now includes unit tests that verify:
    *   The basic DOM structure generated by the widget.
    *   The interaction logic for column header clicks, including:
        *   Initiating ascending sort on the first click.
        *   Reversing to descending sort on the second click.
        *   Clearing the sort (returning to unsorted state) on the third click.
    *   The correct display of sorting indicators (▲ for ascending, ▼ for descending, ● for unsorted).
    *   Proper interaction with the `model.set` and `model.save_changes` methods.
    
    **How to Run**:
    To execute these tests from the project root directory:
    ```bash
    cd tests/js
    npm install  # Only needed if dependencies haven't been installed or have changed
    npm test
    ```

bigframes/display/table_widget.js
```diff
@@ -1,5 +1,5 @@
-/**
- * Copyright 2025 Google LLC
+/*
+ * Copyright 2024 Google LLC
```
tswast: We shouldn't update the copyright year.

```suggestion
 * Copyright 2025 Google LLC
```

bigframes/display/table_widget.js
```diff
@@ -19,155 +19,251 @@ const ModelProperty = {
...
 
-	// Footer: Total rows label
-	const rowCountLabel = document.createElement("div");
+	// Footer styles
+	footer.style.display = "flex";
+	footer.style.justifyContent = "space-between";
+	footer.style.alignItems = "center";
+	footer.style.padding = "8px";
+	footer.style.fontFamily =
+		'-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, sans-serif';
```
tswast: Same with these.

bigframes/display/table_widget.js
```diff
@@ -19,155 +19,251 @@ const ModelProperty = {
...
-	pageSizeLabel.textContent = "Page Size";
-	for (const size of [10, 25, 50, 100]) {
+	const pageSizeInput = document.createElement("select");
+
+	prevPage.textContent = "<";
+	nextPage.textContent = ">";
+	pageSizeLabel.textContent = "Page size:";
+	pageSizeLabel.style.marginRight = "8px";
+	pageIndicator.style.margin = "0 8px";
+	rowCountLabel.style.margin = "0 8px";
```
tswast: Same here. I think we can move these to the CSS file.

bigframes/display/anywidget.py
```diff
@@ -237,6 +259,20 @@ def _set_table_html(self) -> None:
             )
             return
 
+        # Apply sorting if a column is selected
+        df_to_display = self._dataframe
+        if self.sort_column:
+            df_to_display = df_to_display.sort_values(
```
tswast: Could we file a bug and a TODO to support sorting by the index column(s) as well? For those, we'd use sort_index and I thing the "levels" argument.

tswast (commented):
Thanks! Looking good, but I want to make sure we have tests for some different cases for column labels.

tswast (approved):

PR #2273, opened 2025-11-15
refactor: Remove never_skip_nulls param from window node def

TrevorBergeron:


tests/unit/core/compile/sqlglot/aggregations/snapshots/test_unary_compiler/test_all/window_partition_out.sql
```diff
@@ -6,11 +6,7 @@ WITH `bfcte_0` AS (
 ), `bfcte_1` AS (
   SELECT
     *,
-    CASE
-      WHEN `bool_col` IS NULL
-      THEN NULL
-      ELSE COALESCE(LOGICAL_AND(`bool_col`) OVER (PARTITION BY `string_col`), TRUE)
-    END AS `bfcol_2`
+    COALESCE(LOGICAL_AND(`bool_col`) OVER (PARTITION BY `string_col`), TRUE) AS `bfcol_2`
```
tswast: Is this a breaking change? The coalesce means that it's never return NULL, only True, right?

tswast (approved):

PR #2280, opened 2025-11-19
fix: Pass credentials properly for read api instantiation

TrevorBergeron:

### DIFF
--- a/bigframes/session/clients.py
+++ b/bigframes/session/clients.py
@@ -191,10 +191,11 @@ class ClientsProvider:
             client_options=bq_options,
             project=self._project,
             location=self._location,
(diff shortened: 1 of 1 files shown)


tswast (approved):
I double checked on the Client base class for google-cloud-bigquery, and it does appear that both `credentials` and `_http` can be set.

https://github.com/googleapis/python-cloud-core/blob/dc9aa13ed59384914a58f586eafafeb4afac49fa/google/cloud/client/__init__.py#L199-L208

Might be good to add some tests to https://github.com/googleapis/python-cloud-core/tree/main so they don't break that assumption, but I doubt they would as that package is very much in maintenance mode and shouldn't be getting major rewrites.

PR #2290, opened 2025-11-24
fix: Fix issue with stream upload batch size upload limit

TrevorBergeron:

### DIFF
--- a/bigframes/session/loader.py
+++ b/bigframes/session/loader.py
@@ -19,6 +19,7 @@ import dataclasses
 import datetime
 import io
 import itertools
+import math
 import os
 import typing
 from typing import (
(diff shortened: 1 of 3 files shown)


bigframes/session/loader.py
```diff
@@ -434,6 +452,15 @@ def write_data(
         offsets_col: str,
     ) -> bq_data.BigqueryDataSource:
         """Load managed data into bigquery"""
+        MAX_BYTES = 10000000  # streaming api has 10MB limit
```
GarrettWu: Seems the consts can be moved to the top of the file, and if different safety margin there, you can give them different names, SAFTY_MARGIN_STREAM, SAFTY_MARGIN_WRITE, would be easier to read and compare.

GarrettWu (approved):

PR #2298, opened 2025-11-26
feat: Auto-plan complex reduction expressions

TrevorBergeron:


bigframes/core/array_value.py
```diff
@@ -271,15 +270,32 @@ def compute_general_expression(self, assignments: Sequence[ex.Expression]):
...
+        return (ArrayValue(new_root), target_ids)
+
+    def compute_general_reduction(
+        self,
+        assignments: Sequence[ex.Expression],
+        by_column_ids: typing.Sequence[str] = (),
+        *,
+        dropna: bool = False,
+    ):
+        # Warning: this function does not check if the expression is a valid reduction, and may fail spectacularly on invalid inputs
```
tswast: I wonder if there is anything we can do to get the type system to help out here?

bigframes/core/blocks.py
```diff
@@ -1167,6 +1167,45 @@ def project_block_exprs(
             index_labels=self._index_labels,
         )
 
+    # This is a new experimental version of the aggregate that supports mixing analytic and scalar expressions\
+    def reduce_general(
+        self,
+        aggregations: typing.Sequence[ex.Expression] = (),
```
tswast: Are there any restrictions on what kinds of aggregations? As with my comment above, perhaps we can get the type system to help enforce that these are actual aggregations?

tswast (commented):
Neat! My main concern is around making sure we're all on the same page regarding terminology and intention for some of these internal classes and methods.

bigframes/core/expression_factoring.py
```diff
@@ -1,13 +1,100 @@
...
+    graphs,
+    identifiers,
+    nodes,
+    window_spec,
+)
 
 _MAX_INLINE_COMPLEXITY = 10
 
 
+def plan_general_col_exprs(
```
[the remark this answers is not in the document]
TrevorBergeron: Changing name to `apply_agg_exprs_to_plan`, where the plan is the BFET and the exprs are the column expressions describing the full aggregation.

bigframes/core/blocks.py
TrevorBergeron: There is no restriction, as long as input is a column, output is a scalar, added full docstring

bigframes/core/array_value.py
TrevorBergeron: I don't think the python system is sufficiently powerful to do this, instead, will just have to settle for runtime checks. In this case, will get a ValueError if expression is not an aggregating one.

bigframes/core/array_value.py
```diff
@@ -267,19 +266,75 @@ def compute_values(self, assignments: Sequence[ex.Expression]):
...
+        self,
+        assignments: Sequence[ex.Expression],
+        by_column_ids: typing.Sequence[str] = (),
+        *,
+        dropna: bool = False,
+    ):
+        """
+        Applies arbitrary aggregation expressions to the block, optionally grouped by keys.
+
+        This method handles reduction operations (e.g., sum, mean, count) that collapse
```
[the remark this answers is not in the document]
tswast: Presumably, these ops are handled at a diiferent layer to avoid such cases? It'd be good to be explicit in the docs here that ops provided need to work as both analytical and aggregate functions. Or if we can use such functions here, what to do to avoid windows / force group by.

bigframes/core/block_transforms.py
```diff
@@ -622,40 +622,28 @@ def skew(
...
             agg_ops.PopVarOp(),
             ex.deref(col),
         )
-        aggregations.extend([count_agg, moment3_agg, variance_agg])
+        skew_expr = _skew_from_moments_and_count(count_agg, moment3_agg, variance_agg)
+        aggregations.append(skew_expr)
 
-    block, agg_ids = block.aggregate(
-        by_column_ids=grouping_column_ids, aggregations=aggregations
+    block, _ = block.reduce_general(
```
tswast: It's be worth a comment explaining why it's safe to ignore the second element of the returned tuple.

bigframes/core/array_value.py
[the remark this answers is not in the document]
TrevorBergeron: yeah, best to catch these early. adding a windowizable prop to ops, and will throw if try to windowize non-windowizable op

tswast (approved):
Thanks!

PR #2311, opened 2025-12-04
feat: Display custom single index column in anywidget mode

shuoweil:
This PR introduces the single column display for anywidget mode.

A screenshot of single index column display is here: screen/6VLA4Nk68TsYczV

Fixes #<459515995> 🦕


bigframes/display/anywidget.py
```diff
@@ -294,7 +294,15 @@ def _set_table_html(self) -> None:
                 break
 
         # Get the data for the current page
-        page_data = cached_data.iloc[start:end]
+        page_data = cached_data.iloc[start:end].copy()
+
+        # Handle index display
+        if page_data.index.name is not None:
```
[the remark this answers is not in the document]
tswast: Actually, it looks like `df.index.names` won't work in partial ordering mode.

https://github.com/googleapis/python-bigquery-dataframes/blob/481d938fb0b840e17047bc4b57e61af15b976e54/bigframes/dataframe.py#L287

I'll try to figure out the correct alternative and report back.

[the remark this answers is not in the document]
shuoweil: Thanks for the guidance on this, @tswast. Your suggestion to use df._block.has_index was spot on for avoiding the   expensive df.index.names check.

  I've implemented that change. Initially, I used if self._dataframe._block.has_index: directly, but this uncovered  a subtle edge case that caused the test_widget_with_default_index_should_display_row_column test to fail.

  The issue is that a DataFrame can have a materialized index (_block.has_index is True), but that index can still  be the default, unnamed RangeIndex. In that scenario, the previous logic would try to add an index column with page_data.index.name (which is None), resulting in an empty column header where "Row" should be.

  To fix this, I've landed on the following logic:
` if self._dataframe._block.has_index and page_data.index.name is not None:`

  This approach uses your recommended performant check on the original BigFrames DataFrame as the primary condition. The secondary check on page_data.index.name is a cheap, final formatting step on the local in-memory data to correctly handle the display of unnamed indices.

  This fixes the test failure and ensures we correctly display:
   1. A "Row" column for default indices.
   2. The named index column for custom indices.

  Let me know if this approach looks good to you.

bigframes/display/anywidget.py
```diff
@@ -294,7 +294,15 @@ def _set_table_html(self) -> None:
                 break
 
         # Get the data for the current page
-        page_data = cached_data.iloc[start:end]
+        page_data = cached_data.iloc[start:end].copy()
+
+        # Handle index display
```
tswast: I think we need a TODO and bug for MultiIndex support.

bigframes/display/anywidget.py
```diff
@@ -294,8 +294,18 @@ def _set_table_html(self) -> None:
                 break
 
         # Get the data for the current page
-        page_data = cached_data.iloc[start:end]
-
+        page_data = cached_data.iloc[start:end].copy()
+
+        # Handle index display
+        # TODO(b/332316283): Add tests for custom multiindex
```
tswast: Looks like this bug ID is incorrect. Maybe this is the correct one?


```suggestion
        # TODO(b/459515995): Add tests for custom multiindex
```

tests/system/small/test_anywidget.py
```diff
@@ -998,6 +998,140 @@ def test_dataframe_repr_mimebundle_anywidget_with_metadata(
...
+        widget = TableWidget(custom_index_bf_df)
+        html = widget.table_html
+
+    assert "row_1" in html
+    assert "row_2" in html
+    assert "row_3" in html
+    assert "row_4" not in html  # Verify it respects max_rows
+
+
+# TODO(b/332316283): Add tests for custom multiindex
```
tswast: Bug ID looks incorrect here, too.


```suggestion
# TODO(b/459515995): Add tests for custom multiindex
```

tswast (approved):

PR #2271, opened 2025-11-14
refactor: Migrate DataFrame display to use IPython's _repr_mimebundle_() protocol for anywidget mode  

shuoweil:
This PR refactors the DataFrame display system to properly implement IPython's _repr_mimebundle_() protocol for anywidget mode, moving widget handling logic out of _repr_html_() and into the appropriate display method. 

See design doc here: screen/8XWbJDa2d9mb6r5

Fixes #<458796812> 🦕

PR #2316, opened 2025-12-08
revert: DataFrame display uses IPython's `_repr_mimebundle_`

shuoweil:
revert "refactor: Migrate DataFrame display to use IPython's _repr_mimebundle_() protocol for anywidget mode (#2271)"

This reverts commit 41630b5f6b3cd91d700690aee40f342677903805 for bug 466155761.
Verified at Colab: screen/AjTEQC8SrSfMqhN

Fixes #< 466155761 > 🦕


tswast (approved):

bigframes/dataframe.py
```diff
@@ -827,138 +829,68 @@ def __repr__(self) -> str:
...
-        widget_repr_result = widget._repr_mimebundle_(include=include, exclude=exclude)
-
-        # Handle both tuple (data, metadata) and dict returns
-        if isinstance(widget_repr_result, tuple):
-            widget_repr = dict(widget_repr_result[0])  # Extract data dict from tuple
+            for col in blob_cols:
+                # TODO(garrettwu): Not necessary to get access urls for all the rows. Update when having a to get URLs from local data.
+                df[col] = df[col].blob._get_runtime(mode="R", with_metadata=True)
         else:
-            widget_repr = dict(widget_repr_result)
```
tswast: I believe this is where the bug was. We should not discard the metadata. We should save it and return it.

shuoweil: Thanks for the suggestion, I will take a closer look at it after lunch.

PR #2319, opened 2025-12-09
fix: Correct DataFrame widget rendering in Colab

shuoweil:
This PR fix the _get_anywidget_bundle method. Previously, when the underlying widget's _repr_mimebundle_ method returned a (data, metadata) tuple, the code was only  extracting the data portion and discarding the metadata. This resulted in the widget not   rendering correctly in environments like Colab, which rely on this metadata.

The change corrects this by properly unpacking the tuple into widget_repr and widget_metadata. The method now preserves the metadata and returns it along with the data, ensuring that the necessary information for widget rendering is passed on.

We also revert commit 4df3428dde83a57ed183b5441b7d9af7ddab1a17 to reapply  "refactor: Migrate DataFrame display to use IPython's repr_mimebundle() protocol for anywidget mode (https://github.com/googleapis/python-bigquery-dataframes/pull/2271)"

A testcase is added to verify this new change. We also verified at colab: screen/AzGa5RMTJnMH5NH

Fixes #<466155761> 🦕


bigframes/dataframe.py
```diff
@@ -829,68 +827,143 @@ def __repr__(self) -> str:
...
-                # TODO(garrettwu): Not necessary to get access urls for all the rows. Update when having a to get URLs from local data.
-                df[col] = df[col].blob._get_runtime(mode="R", with_metadata=True)
+            if blob_cols:
+                df = self.copy()
+                for col in blob_cols:
+                    # TODO(garrettwu): Not necessary to get access urls for all the rows. Update when having a to get URLs from local data.
+                    df[col] = df[col].blob._get_runtime(mode="R", with_metadata=True)
+        return df, blob_cols
+
+    def _get_anywidget_bundle(self, include=None, exclude=None):
```
tswast: Let's add a return type here, which is a tuple with two dictionaries.

bigframes/dataframe.py
```diff
@@ -829,68 +827,143 @@ def __repr__(self) -> str:
...
+        widget = display.TableWidget(df)
+        widget_repr_result = widget._repr_mimebundle_(include=include, exclude=exclude)
+
+        # Handle both tuple (data, metadata) and dict returns
+        if isinstance(widget_repr_result, tuple):
+            widget_repr, widget_metadata = widget_repr_result
         else:
-            blob_cols = []
+            widget_repr = widget_repr_result
+            widget_metadata = None
```
tswast: we can simplify our return type by constructing an empty dictionary here.

tswast (commented):
Thanks! Some small changes requested.

I've tested this in Jupyter, VS Code, and BQ Studio as well, and it works great.

Image

Even though, we're using Anywidget, I couldn't get it to work on Marimo. https://github.com/marimo-team/marimo/issues/5099#issuecomment-3637697554 I've filed b/467647693 to track this (very low priority).

tswast (approved):

PR #2333, opened 2025-12-16
perf: Fragmentize multiple col expressions as one

TrevorBergeron:

Fixes #470433515 🦕

### DIFF
--- a/bigframes/core/expression_factoring.py
+++ b/bigframes/core/expression_factoring.py
@@ -18,7 +18,10 @@ import dataclasses
 import functools
 import itertools
 from typing import (
+    Callable,
     cast,
+    Dict,
+    Generator,
     Hashable,
     Iterable,
     Iterator,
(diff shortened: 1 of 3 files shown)


bigframes/core/expression_factoring.py
```diff
@@ -101,14 +158,26 @@ class FactoredExpression:
...
+    roots: Sequence[nodes.ColumnDef],
+) -> Sequence[nodes.ColumnDef]:
     """
     The goal of this functions is to factor out an expression into multiple sub-expressions.
     """
-
-    factored_expr = root.expression.reduce_up(gather_fragments)
-    root_expr = nodes.ColumnDef(factored_expr.root_expr, root.id)
-    return (root_expr, *factored_expr.sub_exprs)
+    # TODO: Fragmentize a bit less aggressively
```
sycai: nit: Let's add that bug number after TODO so that we will know in the future what we were doing: go/todo-style#todo-style

sycai (approved):

PR #2338, opened 2025-12-17
feat: Use write api for automatic data uploads

TrevorBergeron:


bigframes/session/__init__.py
```diff
@@ -1026,7 +1025,7 @@ def _read_pandas(
         mem_usage = pandas_dataframe.memory_usage(deep=True).sum()
         if write_engine == "default":
             write_engine = (
-                "bigquery_load"
```
tswast: I'm a bit worried about https://docs.cloud.google.com/bigquery/docs/sandbox users. Maybe we could provide a global option for the default so that such users can switch back to load jobs (accepting the limitations for data types).

TrevorBergeron: added config option in new version

PR #2359, opened 2025-12-29
feat: support nested STRUCT and ARRAY data display in anywidget mode

shuoweil:
Implements flattening and expansion for complex data types in the interactive display for anywidget mode.

  Key Features:
   * Automatic Flattening: STRUCT columns are flattened into intuitive dot-notation columns (e.g., user.name).
   * Array Expansion: ARRAY columns are expanded into multiple rows with visual grouping.
   * Visual Continuity: Continuation rows for arrays are styled for better parent-row context.

verified at:
- vs code notebook: screen/3ST4m9xN9w3iqD9
- colab notebook: screen/7NG4LiTEPuAC27F

Fixes #<438181139> 🦕


bigframes/display/_flatten.py
```diff
@@ -0,0 +1,313 @@
...
+                new_row[f"{col_name}"] = pd.NA
+            exploded_rows.append(new_row)
+            orig_key = str(orig_idx)
+            if orig_key not in array_row_groups:
+                array_row_groups[orig_key] = []
+            array_row_groups[orig_key].append(len(exploded_rows) - 1)
+            continue
+
+        # Create one row per array element, up to max_len_in_row
+        for array_idx in range(max_len_in_row):
```
tswast: This is looping through each element of each array in Python, which is going to be even slower.

tswast (changes requested):

bigframes/display/_flatten.py
shuoweil: I have completely refactored _explode_array_columns to use a vectorized approach with pandas.explode and merge. This eliminated all Python loops, including the slow inner loop you pointed out, significantly improving performance.

bigframes/display/_flatten.py
```diff
@@ -0,0 +1,313 @@
...
+    return (
+        struct_columns,
+        array_columns,
+        array_of_struct_columns,
+        clear_on_continuation_cols,
+        nested_originated_columns,
+    )
+
+
+def _flatten_array_of_struct_columns(
```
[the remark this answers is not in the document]
shuoweil: Thanks for the suggestion. I've refactored the array explosion logic to use a much faster vectorized approach with pandas.explode and merge, which removes the Python loops entirely.

bigframes/display/_flatten.py
```diff
@@ -0,0 +1,528 @@
...
+        if pa_type:
+            if pa.types.is_struct(pa_type):
+                return "struct"
+            if pa.types.is_list(pa_type):
+                return (
+                    "array_of_struct"
+                    if pa.types.is_struct(pa_type.value_type)
+                    else "array"
+                )
+        return "clear"
```
tswast: These magic strings worry me. Could you create an enum for category, instead?

https://docs.python.org/3/library/enum.html

bigframes/display/_flatten.py
```diff
@@ -0,0 +1,528 @@
...
+
+    dataframe: pd.DataFrame
+    row_labels: list[str] | None
+    continuation_rows: set[int] | None
+    cleared_on_continuation: list[str]
+    nested_columns: set[str]
+
+
+@dataclasses.dataclass(frozen=True)
+class ColumnClassification:
```
tswast: Please put a leading `_` in front of class names that aren't intended to be used outside of this module.

bigframes/display/html.py
```diff
@@ -117,6 +130,10 @@ def render_col_header(col):
 
 def _render_table_body(
     dataframe: pd.DataFrame,
+    row_labels: list[str] | None,
+    continuation_rows: set[int] | None,
+    clear_on_continuation: list[str],
```
tswast: Same here, add some more explanation to the docstrings. To keep it shorter, you could reference bigframes/display/_flatten.py so that folks can look there for the complete explanation.

PR #2248, opened 2025-11-10
feat: enable SQLGlot compiler as default

chelsea-lin:
Fixes internal issue 417774347🦕

PR #2362, opened 2025-12-31
refactor: fix str.extract, blob ops and geo ops compiler

chelsea-lin:
This change can resolve multiple bug fixes in #2248: 
- `str.extract` doctest
- multiple blob ops failures
- `test_geo_x_non_point` and `test_geo_drop_duplicates`

Fixes internal issue 417774347 🦕


tswast (approved):
Thanks, Chelsea!

shuoweil (approved):

PR #2380, opened 2026-01-13
feat: Add BigQuery ObjectRef functions to `bigframes.bigquery.obj`

tswast:
This change introduces support for BigQuery ObjectRef functions:
- `OBJ.FETCH_METADATA`
- `OBJ.GET_ACCESS_URL`
- `OBJ.MAKE_REF`

These are exposed via a new `bigframes.bigquery.obj` module.

Replaces https://github.com/googleapis/python-bigquery-dataframes/pull/2340
🦕


bigframes/bigquery/_operations/obj.py
```diff
@@ -0,0 +1,107 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""
+ObjectRef functions defined from
+https://cloud.google.com/bigquery/docs/reference/standard-sql/object-ref-functions
```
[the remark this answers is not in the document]
GarrettWu: https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/objectref_functions

tests/unit/core/compile/sqlglot/expressions/snapshots/test_blob_ops/test_obj_get_access_url/out.sql
```diff
@@ -11,7 +11,7 @@ WITH `bfcte_0` AS (
 ), `bfcte_2` AS (
   SELECT
     *,
-    OBJ.GET_ACCESS_URL(`bfcol_4`) AS `bfcol_7`
+    OBJ.GET_ACCESS_URL(`bfcol_4`, 'R') AS `bfcol_7`
```
GarrettWu: We may still need minimum system tests. Otherwise we may fail to catch errors like this. WDYT?

jialuoo (commented):
Are these test failures related to this PR? I’ve also added Trevor as a reviewer since he probably have more context on the background of these changes than I do.

bigframes/bigquery/_operations/obj.py
```diff
@@ -0,0 +1,107 @@
...
+        duration_micros = utils.timedelta_to_micros(duration)
+
+    return objectref._apply_unary_op(
+        ops.ObjGetAccessUrl(mode=mode, duration=duration_micros)
+    )
+
+
+@log_adapter.method_logger(custom_base_name="bigquery_obj")
+def make_ref(
+    uri_or_json: series.Series,
```
GarrettWu: It can also be a single value in SQL. E.g. OBJ.MAKE_REF("gs://something", "us.my_connection"). It can be useful in local experiments. Do we want to support the case?

bigframes/bigquery/_operations/obj.py
```diff
@@ -0,0 +1,107 @@
...
+import bigframes.core.utils as utils
+import bigframes.operations as ops
+import bigframes.series as series
+
+
+@log_adapter.method_logger(custom_base_name="bigquery_obj")
+def fetch_metadata(
+    objectref: series.Series,
+) -> series.Series:
+    """The OBJ.FETCH_METADATA function returns Cloud Storage metadata for a partially populated ObjectRef value.
```
GarrettWu: Should add the preview disclaimer.

bigframes/bigquery/_operations/obj.py
GarrettWu: Great it can accept list like input now. I meant it can also be a single string value. E.g. OBJ.MAKE_REF("gs://something", "us.my_connection").

tswast: Since I still want to return a Series, this would be a bit confusing. Alternatively, if we get a scalar, we could immediately execute and return a scalar, but I'd like to keep these APIs with as little client-side logic as we can do.

GarrettWu (approved):

PR #2384, opened 2026-01-15
chore: define type logging bit masks

sycai:
This PR only defines the data types and their bit positions. There's a slight update from our design: I reserved the least significant bit for unknown types. 

Tree traversal code will be implemented in the next PR.

Related bug: b/406578908


sycai:
Both E2E and notebook test failures seem be to unrelated to this change.

chelsea-lin (approved):

PR #2390, opened 2026-01-17
chore: Add a function to traverse BFET and encode type usage

sycai:
Next step is to add the encoded usage to the job_config.label before SQL dispatch. 

Related bug: b/406578908

### DIFF
--- a/bigframes/core/logging/data_types.py
+++ b/bigframes/core/logging/data_types.py
@@ -1,0 +1,66 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from bigframes import dtypes
+
+
+def _add_data_type(existing_types: int, curr_type: dtypes.Dtype) -> int:
+    return existing_types | _get_dtype_mask(curr_type)
+
+
+def _get_dtype_mask(dtype: dtypes.Dtype) -> int:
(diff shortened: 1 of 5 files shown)


bigframes/core/logging/data_types.py
```diff
@@ -12,15 +12,94 @@
...
+            )
+        for ordering_expr in node.window_spec.ordering:
+            curr_result = curr_result | _encode_type_refs_from_expr(
+                ordering_expr.scalar_expression, node.child
+            )
+        for col_def in node.agg_exprs:
+            curr_result = curr_result | _encode_type_refs_from_expr(
+                col_def.expression, node.child
+            )
+
```
chelsea-lin: If we can explicitly handle all of existing nodes, we should throw an error for unknown new nodes. If we want to add new BF nodes, this can help us remember handling logs for new BF nodes.

bigframes/core/logging/data_types.py
```diff
@@ -12,15 +12,94 @@
...
+    return child_result | curr_result
+
+
+def _encode_type_refs_from_expr(
+    expr: expression.Expression, child_node: bigframe_node.BigFrameNode
+) -> int:
+    # TODO(b/409387790): Remove this branch once SQLGlot compiler fully replaces Ibis compiler
```
chelsea-lin: Can this throw any errors if we forget to remove this branch?

chelsea-lin (changes requested):

bigframes/core/logging/data_types.py
```diff
@@ -12,15 +12,94 @@
...
+    elif isinstance(node, nodes.ProjectionNode):
+        for assignment in node.assignments:
+            expr = assignment[0]
+            if isinstance(expr, (expression.DerefOp)):
+                # Ignore direct assignments in projection nodes.
+                continue
+            curr_result = curr_result | _encode_type_refs_from_expr(
+                assignment[0], node.child
+            )
+    elif isinstance(node, nodes.SelectionNode):
```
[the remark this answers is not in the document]
sycai: Thanks! I think only `ExplodeNode` applies here because it has `deref` field. There are no column-specific operations for all the other nodes, so I will leave them out.

Added the tracking of columns for `ExplodeNode` in the this change.

chelsea-lin (approved):

PR #2441, opened 2026-02-06
fix: suppress JSONDtypeWarning in Anywidget mode and clean up progress output

shuoweil:
This PR improves the user experience when using the interactive anywidget display mode (bpd.options.display.repr_mode = "anywidget") by reducing console noise.

Verified at:
vs code notebook: screen/ACCJRLwyThciMk8
colab notebook: screen/BhNxzpvckYg9Wp8

Fixes #<482120359> 🦕

### DIFF
--- a/bigframes/display/anywidget.py
+++ b/bigframes/display/anywidget.py
@@ -117,6 +117,7 @@ class TableWidget(_WIDGET_BASE):
         if all(isinstance(col, str) for col in dataframe.columns):
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", bigframes.exceptions.JSONDtypeWarning)
+                warnings.simplefilter("ignore", category=FutureWarning)
                 self.orderable_columns = [
                     str(col_name)
                     for col_name, dtype in dataframe.dtypes.items()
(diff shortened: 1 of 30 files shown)


bigframes/display/anywidget.py
```diff
@@ -114,11 +115,14 @@ def __init__(self, dataframe: bigframes.dataframe.DataFrame):
...
-                if dtypes.is_orderable(dtype)
-            ]
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore", bigframes.exceptions.JSONDtypeWarning)
+                warnings.simplefilter("ignore", category=FutureWarning)
+                self.orderable_columns = [
+                    str(col_name)
+                    for col_name, dtype in dataframe.dtypes.items()
+                    if dtypes.is_orderable(dtype)
+                ]
```
sycai: (No action needed) It seems this constructor is getting way more complicated than necessary. We may need a refactoring some time in the future. Given the tight timeline we have at the moment, this change can be merged as-is in this PR

sycai (commented):
Would it be possible to add a test, assuming that it would not be forbiddingly hard?

sycai (approved):

sycai (approved):

PR #2545, opened 2026-03-31
fix: avoid views when querying BigLake tables from SQL cells

tswast:

Workaround for internal issue b/493608478 🦕


tswast:
`test_read_gbq_w_ambigous_name` failure looks like a real one. I'll revert the change to the ibis compiler. I don't think that part is necessary for the workaround. Just aesthetic for how we escape query parts.

Edit: done in latest commit.

tswast:
Closing. Let's move this over to https://github.com/googleapis/google-cloud-python once https://github.com/googleapis/google-cloud-python/pull/16493 merges.

